PR #7, opened 2021-08-25
add com.unity.modules.jsonserialize to package dependencies

c3-hoge-fuga-piyo:
https://github.com/PlayEveryWare/eos_plugin_for_unity_upm/pull/2

https://github.com/PlayEveryWare/eos_plugin_for_unity_upm/pull/1#issuecomment-902951223


adrianpew:
Thanks for the bug fix.  "com.unity.modules.jsonserialize" is a build-in package, but can be disabled in Package Manager.  Adding this ensures that json package is enabled and available.

PR #21, opened 2021-09-07
fix: Fixed EOSFriendsManager.GetDisplayName cached friends dictionary lookup

acolean-pew:
**EOSFriendsManager.GetDisplayName** currently compares a given user's **EpicAccountId** against each cached friend's **ProductUserId**, meaning the function won't ever find the friend in question. This PR fixes that behavior by using the given user's **EpicAccountId** as the key into the cached friends dictionary (mapped by **EpicAccountId**) to find the friend in question.


adrianpew (approved):
Looks good.  Thanks for the fix.

PR #68, opened 2022-06-16
fix: set LibraryName when using EOS_DISABLE directive

scovertb:
LibraryName wasn't being set when EOS_DISABLE was active, resulting in a syntax error.


zak-pew:
Task linked: [CU-2gpy93e Build Fails if using EOS_DISABLE directive](https://app.clickup.com/t/2gpy93e)

pew-dahs (approved):

PR #104, opened 2022-08-05
Receive Config file on Android

arthur740212:
File access for streaming assets on Android is different, this work around uses UnityWebRequest to receive the file needed.

CU-2p8d19f


Assets/Plugins/Source/EOSManager.cs
```diff
@@ -453,13 +453,36 @@ public void Init(IEOSCoroutineOwner coroutineOwner, string configFileName)
...
+                {
+                    request.timeout = 2; //seconds till timeout
+                    request.SendWebRequest();
 
+                    //Wait till webRequest completed
+                    while (!request.isDone) { }
+
+                    if (request.result != UnityEngine.Networking.UnityWebRequest.Result.Success)
+                    {
+                        UnityEngine.Debug.LogWarning("Requesting " + eosFinalConfigPath + ", please make sure it exists and is a valid config");
```
andrew-hirata-playeveryware: These calls to Unity's logging need to be changed to calls to 'print' so that we can allow users of the plugin to turn off all the logging with `ENABLE_DEBUG_EOSMANAGER`

andrew-hirata-playeveryware (changes requested):

Assets/Plugins/Source/EOSManager.cs
arthur740212: Completed

PR #109, opened 2022-08-08
feat: unsupported platform error

scovertb:
Produce compilation warning and log error when building/running on an unsupported platform


Assets/Plugins/Source/EOSManager.cs
```diff
@@ -419,6 +419,11 @@ private void InitializeOverlay(IEOSCoroutineOwner coroutineOwner)
             // those cases, this code will early out.
             public void Init(IEOSCoroutineOwner coroutineOwner)
             {
+#if !UNITY_EDITOR && !(UNITY_STANDALONE_WIN)
```
pew-dahs, line 422: nit: Is the extra parenthesis needed here?

pew-dahs (approved):
Looks good to me.

Assets/Plugins/Source/EOSManager.cs
scovertb: That's there because the parenthesized section will expand with more supported platforms.

andrew-hirata-playeveryware:
I did a cherry-pick of this on to development instead, so we don't get unneeded commits from stable into development

PR #158, opened 2022-10-24
feat(config): Config Verification Prebuild Step

Raxnam:
A prebuild script that checks for a valid platform config file before starting the build. This throws the error and stops the build before the build actually starts instead of after the build is done

### DIFF
--- a/Assets/Plugins/Source/Editor/PreProcessConfigConfirmation.cs
+++ b/Assets/Plugins/Source/Editor/PreProcessConfigConfirmation.cs
@@ -43,11 +43,11 @@ class PreProcessConfigConfirmation : IPreprocessBuildWithReport
         fileDictionary.Add(BuildTarget.StandaloneLinux64,   "Assets/StreamingAssets/EOS/eos_linux_config.json");
         fileDictionary.Add(BuildTarget.Android,             "Assets/StreamingAssets/EOS/eos_android_config.json");
         /* Unupported cases(Playstation and XBox)
-        file
         fileDictionary.Add(BuildTarget.PS4,             "Assets/StreamingAssets/EOS/.json");
         fileDictionary.Add(BuildTarget.PS5,             "Assets/StreamingAssets/EOS/.json");
         fileDictionary.Add(BuildTarget.Xbox,             "Assets/StreamingAssets/EOS/.json");
         */
+
         if (fileDictionary.ContainsKey(target))
         {
             Debug.Log("PreProcessConfigConfirmation.OnPreprocessBuild for target: " + target + "Target is supported");
--- a/Assets/Plugins/Source/Editor/PreProcessConfigConfirmation.cs
+++ b/Assets/Plugins/Source/Editor/PreProcessConfigConfirmation.cs
@@ -1,0 +1,155 @@
+﻿/*
+* Copyright (c) 2021 PlayEveryWare
+* 
+* Permission is hereby granted, free of charge, to any person obtaining a copy
+* of this software and associated documentation files (the "Software"), to deal
+* in the Software without restriction, including without limitation the rights
+* to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+* copies of the Software, and to permit persons to whom the Software is
+* furnished to do so, subject to the following conditions:
+* 
+* The above copyright notice and this permission notice shall be included in all
+* copies or substantial portions of the Software.
+* 
+* THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+* IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+* FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+* AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+* LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+* OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+* SOFTWARE.
+*/
+
+using UnityEditor.Build.Reporting;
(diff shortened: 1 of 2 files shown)


Assets/Plugins/Source/Editor/PreProcessConfigConfirmation.cs
```diff
@@ -0,0 +1,155 @@
...
+using UnityEditor;
+using UnityEngine;
+using System.IO;
+
+class PreProcessConfigConfirmation : IPreprocessBuildWithReport
+{
+    public int callbackOrder { get { return 0; } }
+    public void OnPreprocessBuild(BuildReport report)
+    {
+        Debug.Log("PreProcessConfigConfirmation_Windows.OnPreprocessBuild for target " + report.summary.platform + " at path " + System.IO.Directory.GetCurrentDirectory());
```
pew-dahs: I'm assuming this should be PreProcessingConfigConfirmation by itself instead of _Windows attached to it? Unless there's going to be platform specific pre process confirmations.

Assets/Plugins/Source/Editor/PreProcessConfigConfirmation.cs
```diff
@@ -0,0 +1,155 @@
...
+        BuildTarget target = report.summary.platform;
+
+        switch (target)
+        {
+            case BuildTarget.StandaloneWindows:
```
pew-dahs: You can simplify this copy/pasted switch to instead use a readonly dictionary (or you could create a local one) of the target and the config file, something like
```
Dictionary<BuildTarget, string> buildFileMap = new Dictionary<BuildTarget, string> {
  { BuildTarget.StandaloneWindows, "Assets/StreamingAssets/EOS/EpicOnlineServicesConfig.json" },
  { BuildTarget.StandaloneWindows64, "Assets/StreamingAssets/EOS/EpicOnlineServicesConfig.json" },
  ...
}
```
Then the whole function can be simplified to
```
if (System.IO.File.Exists(buildFileMap[target]))
{
    Debug.Log("PreProcessConfigConfirmation.OnPreprocessBuild for target " + target + " config file exists");
}
else
{
    Debug.LogError("PreProcessConfigConfirmation.OnPreprocessBuild for target " + target + " config file is missing");
    throw new BuildFailedException("Config file is missing, refer to above message");
}
```
and you can remove the extra functions at the end. In this way, when we need to add new platforms, we don't have to copy/paste a section, but add a new line to the dictionary (or uncomment a line).

pew-dahs (changes requested):

Assets/Plugins/Source/Editor/PreProcessConfigConfirmation.cs
pew-dahs: Also, you should have a default check for the case where the target platform isn't covered with some error message for that too.

Assets/Plugins/Source/Editor/PreProcessConfigConfirmation.cs
Raxnam: Ah yea, I had initially planned to have one for each platform, but the preprocess function provides the target so everything can be done in one function.

Assets/Plugins/Source/Editor/PreProcessConfigConfirmation.cs
Raxnam: Thanks, I'm not too familiar with c# data structures, I just know it doesn't have a map, but that works really well.

pew-dahs: A dictionary in C# is the same as a map in C++, just a different name.

Assets/Plugins/Source/Editor/PreProcessConfigConfirmation.cs
```diff
@@ -0,0 +1,70 @@
...
+
+        Dictionary<BuildTarget, string> fileDictionary = new Dictionary<BuildTarget, string>();
+        fileDictionary.Add(BuildTarget.StandaloneWindows,   "Assets/StreamingAssets/EOS/EpicOnlineServicesConfig.json");
+        fileDictionary.Add(BuildTarget.StandaloneWindows64, "Assets/StreamingAssets/EOS/EpicOnlineServicesConfig.json");
+        fileDictionary.Add(BuildTarget.iOS,                 "Assets/StreamingAssets/EOS/eos_ios_config.json");
+        fileDictionary.Add(BuildTarget.StandaloneOSX,       "Assets/StreamingAssets/EOS/eos_macos_config.json");
+        fileDictionary.Add(BuildTarget.StandaloneLinux64,   "Assets/StreamingAssets/EOS/eos_linux_config.json");
+        fileDictionary.Add(BuildTarget.Android,             "Assets/StreamingAssets/EOS/eos_android_config.json");
+        /* Unupported cases(Playstation and XBox)
+        file
```
pew-dahs: You should remove this line.

Assets/Plugins/Source/Editor/PreProcessConfigConfirmation.cs
```diff
@@ -0,0 +1,70 @@
...
+class PreProcessConfigConfirmation : IPreprocessBuildWithReport
+{
+    public int callbackOrder { get { return 0; } }
+    public void OnPreprocessBuild(BuildReport report)
+    {
+        Debug.Log("PreProcessConfigConfirmation.OnPreprocessBuild for target: " + report.summary.platform);
+        BuildTarget target = report.summary.platform;
+
+        Dictionary<BuildTarget, string> fileDictionary = new Dictionary<BuildTarget, string>();
+        fileDictionary.Add(BuildTarget.StandaloneWindows,   "Assets/StreamingAssets/EOS/EpicOnlineServicesConfig.json");
```
pew-dahs, line 39: nit: You can actually initialize a dictionary directly without adding key/value pairs after. No need to change it, but nice to know about C# initialization.
```
Dictionary<BuildTarget, string> fileDictionary = new Dictionary<BuildTarget, string> {
  { BuildTarget.StandaloneWindows, "Assets/StreamingAssets/EOS/EpicOnlineServicesConfig.json" },
  ...
};
```

pew-dahs (approved):
Looks good. Just need to remove an extra line that's there.

andrew-hirata-playeveryware (approved):

PR #160, opened 2022-11-01
fix: invalid IntPtr Null check

joetaylor1987:
fix: invalid IntPtr Null check

`m_OverrideLibraryPath != null` was casuing the following error using warning level 5 and warnaserrors enabled:
```
error CS8073: The result of the expression is always 'true' since a value of type 'IntPtr' is never equal to 'null' of type 'IntPtr?'
```

### DIFF
--- a/Assets/Plugins/Linux/EOSManager_Linux.cs
+++ b/Assets/Plugins/Linux/EOSManager_Linux.cs
@@ -91,7 +91,7 @@ namespace PlayEveryWare.EpicOnlineServices
         {
             set
             {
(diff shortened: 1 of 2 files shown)


andrew-hirata-playeveryware:
Good find!

andrew-hirata-playeveryware:
I'm going to wait until tomorrow my time (Currently 16:25 PDT) till I squash & merge this in just in case you find any more IntPtr comparison warnings you'd like to fix, @joetaylor1987

joetaylor1987:
Cool, thanks!

PR #161, opened 2022-11-02
fix: EOSbootstrapper and EAC binary install when building from plugin

andrew-hirata-playeveryware:
This work does two things:
1) Fix some path issues that can occur when the code is run from the plugin directory
2) Adds some of the work for allowing a user specified path to find the EAC cert and public key.

### DIFF
--- a/Assets/Plugins/Windows/Editor/EOSOnPostprocessBuild_Windows.cs
+++ b/Assets/Plugins/Windows/Editor/EOSOnPostprocessBuild_Windows.cs
@@ -107,7 +107,7 @@ public class EOSOnPostprocessBuild_Windows:  IPostprocessBuildWithReport
         string appFilenameExe = Path.GetFileName(report.summary.outputPath);
         string installDirectory = Path.GetDirectoryName(report.summary.outputPath);
         string installPathForEOSBootStrapper = Path.Combine(installDirectory, "EOSBootStrapper.exe");
-        string workingDirectory = Path.Combine(Application.dataPath, "../bin/");
+        string workingDirectory = GetPathToEOSBin();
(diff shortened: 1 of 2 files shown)


Assets/Plugins/Windows/Editor/EOSOnPostprocessBuild_Windows.cs
```diff
@@ -304,11 +304,24 @@ public void OnPostprocessBuild(BuildReport report)
 
             InstallFiles(report);
             
-            string pathToEOSBootStrapperTool = GetPathToEOSBin() + "/EOSBootstrapperTool.exe";
-            string pathToEOSBootStrapper = GetPathToEOSBin() + "/EOSBootStrapper.exe";
-            string pathToEACIntegrityTool = GetPathToEOSBin() + "/EAC/anticheat_integritytool.exe";
+            string pathToEOSBootStrapperTool = Path.Combine(GetPathToEOSBin(), "EOSBootstrapperTool.exe");
+            string pathToEOSBootStrapper = Path.Combine(GetPathToEOSBin(), "EOSBootStrapper.exe");
+            string pathToEACIntegrityTool = Path.Combine(GetPathToEOSBin(), "EAC/anticheat_integritytool.exe");
```
scovertb, line 309: This should maybe be `Path.Combine(GetPathToEOSBin(), "EAC", "anticheat_integritytool.exe")`

scovertb (approved):

pew-dahs (approved):

mattiascibien:
I think this should be in the project as with 2.1.6 the paths to EAC are not good as the package path does not contain eac files

PR #166, opened 2022-11-16
feat: p2p netcode sample

scovertb:
Implemented sample using EOSTransport as an interface for Netcode for GameObjects. Netcode plugin is only required to run sample.
Refactored UIFriendsMenu to generalize interaction with friend list entries.
Fixed editor crashes when exiting play mode while using the EOS P2P interface.
Made EOSSessionsManager.SetJoininfo a static function which should probably be moved somewhere else eventually.


Assets/Scripts/UI/UIP2PTransportMenu.cs
```diff
@@ -0,0 +1,299 @@
...
+        }
+
+        public override void ResetDirtyFlag()
+        {
+            uiDirty = false;
+        }
+
+        public void StartHostOnClick()
+        {
+            if (isHost)
```
pew-dahs, line 121: I'm assuming that it's not possible to have `isHost` and `isClient` both true due to how the UI is connected since there's no checks for it.

pew-dahs (approved):
The only thing I saw was since the invite, add, chat, and report buttons have been compressed to one button, does that mean you can only do one of those actions per friend entry? Is there a case where you might need to have more than one?

Assets/Scripts/UI/UIP2PTransportMenu.cs
scovertb: Yes, being a host and client is mutually exclusive and becoming either disables all UI for becoming or joining a host.

scovertb: I couldn't find any cases where multiple friend actions were available simultaneously. Plus, all the old buttons were stacked on top of each other so it wouldn't have working anyways. If that is wanted later we can add support then.

PR #178, opened 2022-12-07
Android lib link statically/dynamically

arthur740212:
Toggle for linking eos library statically in config editor
The target eos-sdk would be added from PlatformSpecificStreamingAssets to Plugin/Android folder
according to the selected method


andrew-hirata-playeveryware:
The assets that were added to this review need to be moved from the ```Assets``` directory and into ```PlatformSpecificAssets/EOS/Android```, which is one directory up, otherwise I believe both will be included in the build.

Assets/Plugins/AndroidEditor/Editor/EOSOnPreprocessBuild_android.cs
```diff
@@ -221,4 +223,32 @@ public void ConfigureEOSDependentLibrary()
         }
     }
 
+    private void DetermineLibraryLinkingMethod()
+    {
+        var configFilenamePath = EpicOnlineServicesConfigEditor.GetConfigPath(PlatformSpecificConfigEditorAndroid.ConfigFilename);
+        EOSConfigFile<EOSAndroidConfig> configFile = new EOSConfigFile<EOSAndroidConfig>(configFilenamePath);
+        configFile.LoadConfigFromDisk();
+
+        var sdkAndroidPath = System.IO.Path.Combine("Assets", "PlatformSpecificStreamingAssets", "EOS", "Android");
```
andrew-hirata-playeveryware: The path here will need to be corrected when the added assets are moved to the correct location.

andrew-hirata-playeveryware: Additionally, this needs to be modified such that it works when the code is run from the plugin directory, which has a different path layout. The general way to get files relative the package install location is by doing something like this: 

```
string packagePathname = Path.GetFullPath("Packages/" + GetPackageName() + "/PlatformSpecificStreamingAssets~/" + subpath);
```

Where `GetPackageName` is defined like so:
```
    private static string GetPackageName()
    {
        return "com.playeveryware.eos";
    }
```

While not strictly required, you can also use the following to find the assets, so that it's not 'hard coded' to the asset's directory:
```
string streamingAssetsSamplesPathname = Path.Combine(Application.dataPath, "../PlatformSpecificStreamingAssets/" + subpath);
```
As you'll see here, the second parameter into the `Path.Combine` method has a '..' in front of it, which allows one to find sibling directories to the `Assets` directory.

Assets/Plugins/AndroidEditor/Editor/PlatformSpecificConfigEditorAndroid.cs
```diff
@@ -69,6 +69,8 @@ public void OnGUI()
                 overrideValues = configFile.currentEOSConfig.overrideValues;
             }
 
+            EpicOnlineServicesConfigEditor.AssigningBoolField("Link EOS-SDK Statically", ref configFile.currentEOSConfig.linkLibStatically);
```
andrew-hirata-playeveryware: This config option should be put in a plugin editor config section, similar to how `EOSPluginEditorToolsConfig` is done, as this config option is ideally for runtime only specific configuration.

Assets/Plugins/Android/EOSAndroidConfig.cs
```diff
@@ -34,6 +34,7 @@ public class EOSAndroidConfig : ICloneableGeneric<EOSAndroidConfig>, IEmpty
 
         public EOSConfig overrideValues;
 
+        public bool linkLibStatically = false;
```
andrew-hirata-playeveryware: As mentioned in the PlatformSpecificConfigEditorAndroid comment, this should preferably be in it's own IEOSPluginEditorConfigurationSection for android options

andrew-hirata-playeveryware (changes requested):

Assets/Plugins/AndroidEditor/Editor/EOSOnPreprocessBuild_android.cs
```diff
@@ -221,4 +223,34 @@ public void ConfigureEOSDependentLibrary()
...
+            {
+                androidBuildConfigSection.LoadConfigFromDisk();
+            }
+        }
+
+        var androidAssetFilepath = "../PlatformSpecificAssets/EOS/Android/";
+        var libSubpath = "/aar/eos-sdk.aar";
+
+        //string packagePathname = Path.GetFullPath("Packages/" + GetPackageName() + "/PlatformSpecificAssets~/" );
+        string dynamicPathName = Path.Combine(Application.dataPath, androidAssetFilepath, "dynamic-stdc++" + libSubpath);
```
andrew-hirata-playeveryware: This still needs to have changes done so that it handles the case where it's running in the plugin, as the path names aren't shared. One can get the plugin path in the way described here https://github.com/PlayEveryWare/eos_plugin_for_unity/pull/178#discussion_r1046513278

andrew-hirata-playeveryware (changes requested):

Assets/Plugins/Android/Editor/EOSOnPreprocessBuild_android.cs
```diff
@@ -221,4 +223,33 @@ public void ConfigureEOSDependentLibrary()
...
+            {
+                androidBuildConfigSection.LoadConfigFromDisk();
+            }
+        }
+
+        string packagePathname = Path.GetFullPath("Packages/" + GetPackageName() + "/PlatformSpecificAssets~/EOS/Android/");
+        string libSubpath = "/aar/eos-sdk.aar";
+
+        string dynamicPathName = Path.Combine(Application.dataPath, packagePathname, "dynamic-stdc++" + libSubpath);
+        string staticPathName = Path.Combine(Application.dataPath, packagePathname, "static-stdc++" + libSubpath);
```
andrew-hirata-playeveryware: This is checking for the paths incorrectly.
It needs check if the libraries exist in the `packagePathname` directory, and if they don't, assume they exist in the `../PlatformSpecificAssets/EOS/Android` path and copy them into `destPathname`.

andrew-hirata-playeveryware (changes requested):

andrew-hirata-playeveryware (approved):
I think this needs to have the conflict resolved and then be squashed, but this code otherwise looks good to merge.

PR #183, opened 2022-12-16
Make EOSManager extendable

kamyker:
I wanted to add event right before eos platform/dll is released (to release other callbacks like LobbyInterface.RemoveNotifyLobbyMemberStatusReceived).

This change allows me to extend EOSManager and add it:
```
protected override void OnApplicationQuit()
{
    OnShutdown?.Invoke();
    base.OnApplicationQuit();
}
```


RikuTheFuffs (approved):
+1 on the feature, makes sense and the changes look good to me

scovertb:
This seems like a good change, but would need to be merged to development instead of stable.

kamyker:
> This seems like a good change, but would need to be merged to development instead of stable.

Ok, didn't know that branch exists https://github.com/PlayEveryWare/eos_plugin_for_unity/pull/209

PR #221, opened 2023-02-09
Readme updates

Raxnam:
unify the readmes
     because the structures of the two repos are different, they cannot be exactly the same and need slightly different paths and information. Either by restructuring the documentation folders of each or through a script they could probably get to the point where all but the first few paragraphs are identical
fix attempting to rejoin a lobby you are already in
updated the friends menu to be color coded

### DIFF
--- a/docs/upm_readme.md
+++ b/docs/upm_readme.md
@@ -16,7 +16,7 @@ For [support issues](https://github.com/PlayEveryWare/eos_plugin_for_unity/issue
 
 
 ## Supported Platforms
-The follow target platforms are supported in Unity for the current release of the EOS Unity Plugin.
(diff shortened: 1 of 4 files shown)


andrew-hirata-playeveryware (approved):

docs/upm_readme.md
```diff
@@ -16,7 +16,7 @@ For [support issues](https://github.com/PlayEveryWare/eos_plugin_for_unity/issue
 
 
 ## Supported Platforms
-The follow target platforms are supported in Unity for the current release of the EOS Unity Plugin.
+The follow target platforms are supported in Unity for the current release of the plugin.
```
scovertb, line 19: This should say "following"

scovertb (approved):

PR #226, opened 2023-03-05
Allow Android development on MacOS

leXorNet:
Using MacOS code paths when running in editor on MacOS as expected, even when Unity platform is set to Android


andrew-hirata-playeveryware:
May I have you do this PR such that it merges into `development` instead of `stable`? While stable is the default branch, because it's "guaranteed" to work, PR and new changes should all go into development, to ensure it works with the latest things we might be working on.

leXorNet:
I'll create a new pull request in the dev branch

PR #227, opened 2023-03-06
Use MacOS code paths when running in editor on mac

leXorNet:
Use MacOS code paths when running in editor on mac, even when Unity platform is set to Android.

### DIFF
--- a/Assets/Plugins/Source/EOSManager.cs
+++ b/Assets/Plugins/Source/EOSManager.cs
@@ -527,7 +527,7 @@ namespace PlayEveryWare.EpicOnlineServices
 
                 string configDataAsString = "";
 
-#if UNITY_ANDROID
+#if UNITY_ANDROID && !UNITY_EDITOR
 
                 configDataAsString = AndroidFileIOHelper.ReadAllText(eosFinalConfigPath);
 #else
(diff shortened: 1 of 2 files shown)


leXorNet:
@andrew-hirata-playeveryware Is this better? :)

Assets/Plugins/Source/EOSManager.cs
```diff
@@ -527,7 +527,7 @@ public void Init(IEOSCoroutineOwner coroutineOwner, string configFileName)
 
                 string configDataAsString = "";
 
-#if UNITY_ANDROID
+#if UNITY_ANDROID && !UNITY_EDITOR
```
arthur740212, line 530: Thanks for the discovery.
Though this might break for android on editor_windows.
!UNITY_EDITOR_OSX might fix the problem

leXorNet: I don't think you want to run android code in the editor on windows either?
That is only ever useful in standalone builds (on device) as far as I know...

arthur740212: With platform set to Android, it was running on editor, partially using android utilities, that is why we wouldn't want to change UNITY_EDITOR_WIN part.
Though it seems that with !UNITY_EDITOR android on windows editor still works, and after reading what the code it makes sense to use !UNITY_EDITOR.
Thanks for the PR.

PR #228, opened 2023-03-09
TMP removal

Raxnam:
Removed TMP assets and package from the performance stress test scene


pew-dahs (approved):

andrew-hirata-playeveryware (approved):
Looks good!

PR #240, opened 2023-03-27
fix: Don't check for valid EOS config files when the build target has

andrew-hirata-playeveryware:
Small fix related to #237 . This change checks the defines for the for the build, and if it includes EOS_DISABLED, it won't check if the files exist on disk.

One edge case with this, is that if another prebuild script is run that adds EOS_DISABLED later, this script might not pick it up.

### DIFF
--- a/Assets/Plugins/Source/Editor/PreProcessConfigConfirmation.cs
+++ b/Assets/Plugins/Source/Editor/PreProcessConfigConfirmation.cs
@@ -32,6 +32,13 @@ class PreProcessConfigConfirmation : IPreprocessBuildWithReport
     public int callbackOrder { get { return 0; } }
     public void OnPreprocessBuild(BuildReport report)
     {
+        // Ensure that we don't do checks for config files if the current platform group has EOS disabled
+        var defineSymbolsForGroup = PlayerSettings.GetScriptingDefineSymbolsForGroup(report.summary.platformGroup);
+        if (defineSymbolsForGroup.Contains("EOS_DISABLE"))
+        {
+            return;
+        }
+
         Debug.Log("PreProcessConfigConfirmation.OnPreprocessBuild for target: " + report.summary.platform);
         BuildTarget target = report.summary.platform;
 
--- a/Assets/Plugins/Source/Editor/PreProcessConfigConfirmation.cs
+++ b/Assets/Plugins/Source/Editor/PreProcessConfigConfirmation.cs
@@ -29,7 +29,7 @@ using System.Collections.Generic;
 
 class PreProcessConfigConfirmation : IPreprocessBuildWithReport
 {
-    public int callbackOrder { get { return 0; } }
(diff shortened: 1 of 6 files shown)


Assets/Plugins/Source/Editor/PreProcessConfigConfirmation.cs
```diff
@@ -32,6 +32,13 @@ class PreProcessConfigConfirmation : IPreprocessBuildWithReport
     public int callbackOrder { get { return 0; } }
     public void OnPreprocessBuild(BuildReport report)
     {
+        // Ensure that we don't do checks for config files if the current platform group has EOS disabled
+        var defineSymbolsForGroup = PlayerSettings.GetScriptingDefineSymbolsForGroup(report.summary.platformGroup);
+        if (defineSymbolsForGroup.Contains("EOS_DISABLE"))
```
pew-dahs: What's the pros/cons of doing it this way vs adding #if !EOS_DISABLE/endif around the function body? I never thought to check for defines in the player settings itself.

scovertb:
This should also be applied to EOSOnPreprocessBuild and the post process scripts for mobile and standalone.

Assets/Plugins/Source/Editor/PreProcessConfigConfirmation.cs
scovertb: One con is that GetScriptingDefineSymbolsForGroup is marked for deprecation.

scovertb: It might be worth making a separate editor assembly that is disabled when EOS_DISABLE is set.

andrew-hirata-playeveryware: I used GetScriptingDefineSymbolsForGroup, instead of GetScriptingDefineSymbols because it's available in all versions in 2020+ so far released, but GetScriptingDefineSymbols, is not. 

The pro is that if something adds a define, using the method to access the defines that will be used in the build will still work, i.e. it works in more cases than just using the EOS_DISABLE conditional compile.

I'm not aware of any particular cons with using the method, except that unity sometimes deprecates methods and adds replacements for them, as in the case that Steven pointed out.

scovertb: For slight future proofing it might be good to use the `PlayerSettings.GetScriptingDefineSymbols` function using a `NamedBuildTarget` received from `UnityEditor.Build.NamedBuildTarget.FromBuildTargetGroup`

scovertb: That said, `UnityEditor.Build.NamedBuildTarget.FromBuildTargetGroup` appears to be undocumented so who know who that'll change.

scovertb (approved):

pew-dahs (approved):

PR #284, opened 2023-05-16
[Fix] EOSConfig.sandboxDeploymentOverrides nullref

jsoysouvanh:
Nullref happened when opening EpicOnlineServicesConfigEditor with no config file to load from disk.
Newing the list on the default object construction fixes the problem and doesn't impact subsequent deserializations of the object.

### DIFF
--- a/Assets/Plugins/Source/Core/EOSConfig.cs
+++ b/Assets/Plugins/Source/Core/EOSConfig.cs
@@ -63,7 +63,7 @@ namespace PlayEveryWare.EpicOnlineServices
         public string deploymentID;
 
         /// <value><c>SandboxDeploymentOverride</c> pairs used to override Deployment ID when a given Sandbox ID is used</value>
-        public List<SandboxDeploymentOverride> sandboxDeploymentOverrides;
+        public List<SandboxDeploymentOverride> sandboxDeploymentOverrides = new();
 
         /// <value><c>Client Secret</c> defined in the [Development Portal](https://dev.epicgames.com/portal/)</value>
         public string clientSecret;


andrew-hirata-playeveryware:
Hello @jsoysouvanh, thank you for the merge request. We currently we already have a fix for this in `development` in the commit c7784fc2. Additionally, your change is using the new C# 9.0  target typed syntax, which isn't supported by Unity 2020, which we don't "technically" support, but I do believe there are some users that are still using that version. Was this issue still reproducing for you in the `development` branch without your change here?

jsoysouvanh:
Hi @andrew-hirata-playeveryware, thank you for your feedback.
Sorry for this, I installed the package from the last stable release so I didn't try the development branch.
I just checked the new statement was not in this file in the last development commit, but haven't taken the time to check the rest of the codebase for a potential fix for this.
If it has already been fixed somewhere else, my apologies!
Regarding the syntax, indeed my mistake, it could easily be replaced by the more standard full syntax with the class name instead. 
Thanks for your time!

andrew-hirata-playeveryware:
The issue was fixed it in a different way, in the [EpicOnlineServicesConfigEditor.cs](https://github.com/PlayEveryWare/eos_plugin_for_unity/blob/e0e1bf35b921a47a2b6990fad8f985f49c247732/Assets/Plugins/Source/Editor/EpicOnlineServicesConfigEditor.cs#LL512C16-L512C16), where it gets created before we try to read it in the editor. Concerning the syntax, I should add a section to specifically call out which kind of syntax we allow in the core code to ensure that it works with the widest number of Unity versions and C# standards. 

Also, thank you very much both replying and for initial pull request, even if we're not actually going to integrate it.

I'll leave this PR open for a little bit just in case you find a situation where it still is broken, despite our fix.

PR #289, opened 2023-05-17
fix(iOS) : IOS Login Options Refactor

arthur740212:
Moves MakeIOSLoginOptions from EOSManager_iOS to a new helper script in iOS/Core to fit the core assembly structure.

### DIFF
--- a/Assets/Plugins/iOS/Core/EOS_iOSLoginOptionsHelper.cs
+++ b/Assets/Plugins/iOS/Core/EOS_iOSLoginOptionsHelper.cs
@@ -1,0 +1,43 @@
+using System;
+using Epic.OnlineServices.Auth;
+using System.Runtime.InteropServices;
+
+#if UNITY_IOS && !UNITY_EDITOR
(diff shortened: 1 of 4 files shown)


Raxnam (approved):

Assets/Plugins/iOS/Core/EOS_iOSLoginOptionsHelper.cs
```diff
@@ -0,0 +1,41 @@
+using System;
```
andrew-hirata-playeveryware, line 25: This needs to have a license added to the top of the file.

andrew-hirata-playeveryware (changes requested):

Assets/Plugins/iOS/Core/EOS_iOSLoginOptionsHelper.cs
arthur740212: Ready, I added the license with year 2023, is that correct

andrew-hirata-playeveryware (approved):

PR #307, opened 2023-06-07
Feat(Transport Layer) Added functionality to Start Client

Raxnam:
added fields to copy and input PUID for hosting/joining servers via the network manager

### DIFF
--- a/Assets/Scripts/Networking/EOSTransport.cs
+++ b/Assets/Scripts/Networking/EOSTransport.cs
@@ -76,6 +76,12 @@ namespace PlayEveryWare.EpicOnlineServices.Samples.Network
         // Override local user id for testing multiple clients at once
         public ProductUserId LocalUserIdOverride = null;
 
+        //editor field to input a PUID to connect to(ease of access when using editor buttons in the network manager
+        public String ServerUserIdToConnectToInput = null;
+
+        //String of the current server PUID for copying purposes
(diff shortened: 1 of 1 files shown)


Assets/Scripts/Networking/EOSTransport.cs
```diff
@@ -76,6 +76,12 @@ public class EOSTransport : NetworkTransport
         // Override local user id for testing multiple clients at once
         public ProductUserId LocalUserIdOverride = null;
 
+        //editor field to input a PUID to connect to(ease of access when using editor buttons in the network manager
+        public String ServerUserIdToConnectToInput = null;
```
andrew-hirata-playeveryware, line 81: If this is editor only, shouldn't we wrap this in a UNITY_EDITOR conditional compile?

arthur740212 (approved):

andrew-hirata-playeveryware (changes requested):

Assets/Scripts/Networking/EOSTransport.cs
andrew-hirata-playeveryware: Oh my apologizes: I meant the fields and their uses should be wrapped, not just the usage.

andrew-hirata-playeveryware (changes requested):

andrew-hirata-playeveryware (approved):

PR #319, opened 2023-06-14
Fix Added Preserve tags to platform specific code

Raxnam:
Added the [Preserve} tag to each platform's EOSManager to ensure it does not get stripped during IL2CPP compilation


andrew-hirata-playeveryware:
I added more people to the review because I would like people to be aware of this change.

andrew-hirata-playeveryware (approved):

arthur740212 (approved):

PR #323, opened 2023-06-17
doc(android) : getiing started and version

arthur740212:


docs/android/readme_android.md
```diff
@@ -1,28 +1,67 @@
...
+
+
+# <div align="center">$\textcolor{deeppink}{\textsf{Android Getting Started}}$</div> <a name="getting-started" />
+---
+
+## Prerequisites
+
+
+* The standard <a href="/readme.md#prerequisites">Prerequisites</a> for all platforms.
+* Windows Build Support (IL2CPP)
```
andrew-hirata-playeveryware: Is windows build support actually required? That's somewhat shocking if it is true.

arthur740212: I thought it was. It was in the docs I written back when we started android
But I also question it now.

arthur740212: Should I remove it for now?

RedMazer: ill run things through a test process, so ill check this when i get there and make any needed changes

PR #343, opened 2023-07-05
fix(android,login) : fix for login not returning to the application

arthur740212:
The reason was because the eos_login_protocol_scheme was overwritten by an empty entry. This fix assigns the correct string to the empty entry to avoid this problem

### DIFF
--- a/Assets/Plugins/Android/Editor/EOSOnPreprocessBuild_android.cs
+++ b/Assets/Plugins/Android/Editor/EOSOnPreprocessBuild_android.cs
@@ -394,10 +394,20 @@ public class EOSOnPreprocessBuild_android : IPreprocessBuildWithReport
         currentEOSValuesConfigAsXML.Load(pathToEOSValuesConfig);
 
         var node = currentEOSValuesConfigAsXML.DocumentElement.SelectSingleNode("/resources");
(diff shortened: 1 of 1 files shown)


RedMazer (approved):
im still not a fan of 'var' particularly when doing a review, but i think it looks good?

though questions on the ifs?
```        if (storedClientID != clientIDAsLower)
        {
            node2.InnerText = $"eos.{clientIDAsLower}";
            currentEOSValuesConfigAsXML.Save(pathToEOSValuesConfig);
        }
```


        like `storedClientID ` cant be uppercase?
        
        and the two null checks, are they not supposed to zip or delete the newly created things first?

arthur740212:
The part to modify `value.xml` was mostly the same as the one for `eos_value.xml`
the vars are from there, which isn't ideal.
Though this fix would be obsolete when 1.16 drops,
I figure getting a working build asap would be better, and we could revamp the vars for configuring `eos_value.xml` after this.

As for storedClientID, in this particular case it is always empty, since the file is copied from a clean arr.

And you're right about cleaning stuff before the returns, though it is actually expected that it never hits those two.
I'll make some changes there.

andrew-hirata-playeveryware (approved):

PR #357, opened 2023-07-25
feat: add bool to control if the Plugin shuts down the SDK on App quit.

andrew-hirata-playeveryware:


RedMazer:
is on quit the definitive place for that? like how unity has multiple 'onstart' options, are there any other places this needs to be added?

Raxnam:
> 

It looks like there may be some platform specific shutdown sequences for ios and specific android/windows phones

https://docs.unity3d.com/ScriptReference/MonoBehaviour.OnApplicationQuit.html

andrew-hirata-playeveryware:
It depends on one's goals, @RedMazer : For our purposes, it is to do a best case shutdown of the EOS SDK, plus any of our EOS handles. So for us, this is fine. And yes, as @Raxnam points out, on many mobile platforms, it's possible that the application will be terminated before any of our shutdown code will be called, which for what we do in `OnShutdown` is fine. 

Of note: the original way I was planning to implement this was by adding a callback system of some sort so Plugin users could do any cleanup they want in a async way. However, I realized this kind of API might imply it would _always_ be called, even if we mentioned that it can't be guaranteed.

Raxnam (approved):

PR #361, opened 2023-07-28
Fix IL2CPP stripping platform-specific initialization code

WI-Lee:
This fixes IL2CPP stripping the platform-specific `EOSManager_*` classes by adding the [`[assembly: AlwaysLinkAssembly]`](https://docs.unity3d.com/ScriptReference/Scripting.AlwaysLinkAssemblyAttribute.html) attribute.

Without this attribute, the entire `com.playeveryware.eos` assembly is stripped, because none of the classes are referenced by any other assembly. This means all of the `[RuntimeInitializeOnLoadMethod]` Register methods are stripped, so the plugin is not initialized properly.

I've also removed the `[Preserve]` attributes added in #319 as they are no longer necessary with this fix.

### DIFF
--- a/Assets/Plugins/Android/EOSManager_Android.cs
+++ b/Assets/Plugins/Android/EOSManager_Android.cs
@@ -41,6 +41,7 @@ using System.Diagnostics;
 
 
 #if UNITY_ANDROID && !UNITY_EDITOR
+[assembly: AlwaysLinkAssembly]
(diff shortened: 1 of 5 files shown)


andrew-hirata-playeveryware (approved):
Looks good! Thank you for the PR and the fix!

arthur740212 (approved):
Splendid find.  Huge thanks for the fix. 💯

PR #363, opened 2023-08-01
feature : Manual remove persistent auth token

arthur740212:


Raxnam (approved):

Assets/Plugins/Source/Core/EOSManager.cs
```diff
@@ -1372,6 +1372,28 @@ public void ClearConnectId(ProductUserId userId)
...
+            //-------------------------------------------------------------------------
+            /// <summary>
+            /// Clears the stored token for persistent login
+            /// </summary>
+            public void RemovePersistentToken()
+            {
+                var authInterface = EOSManager.Instance.GetEOSPlatformInterface().GetAuthInterface();
+                var options = new Epic.OnlineServices.Auth.DeletePersistentAuthOptions();
+
+                authInterface.DeletePersistentAuth(ref options, null, (ref DeletePersistentAuthCallbackInfo deletePersistentAuthCallbackInfo) =>
```
paulhazen, line 1384: I'm so used to C++. Can you remind me what the `ref` keyword does in this context?

paulhazen (approved):
Looks good to me :)

Assets/Plugins/Source/Core/EOSManager.cs
andrew-hirata-playeveryware: `ref` is a lot like passing a value by address in C++: it passed it by reference/address instead of by value.

andrew-hirata-playeveryware (approved):

PR #366, opened 2023-08-03
Build: Added support for command-line package creation

paulhazen:
Without trying to change the implementation of *how* the export tool works, I added support for exporting the EOS Plug-In from the command line. This should help in the future with build automation and testing endeavors. 

### UnityPackageCreationUtility
To facilitate the change, I took the functionality in `UnityPackageCreationTool.cs` that could be made static and move it into a new static class `UnityPackageCreationUtility`, so the functionality is the same, but it's now possible to utilize those methods without dealing with the `EditorWindow` base class of the `UnityPackageCreationTool`.

### BuildPackage
Moving the functionality to a static place made it easier to implement the static class `BuildPackage`, the central function of which `ExplortPlugin` can be called from the command line.

The following command generates a new `com.playeveryware.eos-2.3.3.tgz` file at the indicated output directory, the same exact way it would if you created a package via Unity Editor Tools -> Create Package (and subsequently pressing "Create UPM Package"):

```
Unity.exe -batchMode \
    -nographics \
    -quit \
    -projectPath [Path to eos_plugin_for_unity root directory] \
    -executeMethod BuildPackage.ExportPlugin \
    -EOSPluginOutput [Absolute Path to Output]
```

The following command-line arguments are used by unity, and you can read the docs on them for fuller detail [here](https://docs.unity.com/ugs/en-us/manual/ccd/manual/UnityCCDCLI):

`-batchMode`
Indicates whether Unity should launch a window or run 'headless'.

`-nographics`
Allows the build to be done on hardware that does not have a GPU. 

`-quit` 
On success or failure, quit. 

`-projectPath` 
The path to the root of this repository.

`-executeMethod` 
Indicates which static class and which static function on that class should be executed. 

`-EOSPluginOutput` 
The directory in which the newly created tarball should be placed.

### DIFF
--- a/Assets/Plugins/Source/Editor/Internal/BuildPackage.cs
+++ b/Assets/Plugins/Source/Editor/Internal/BuildPackage.cs
@@ -140,7 +140,7 @@ public static class BuildPackage
 
             foreach (string dirpath in Directory.EnumerateDirectories(output_path))
             {
-                Directory.Delete(dirpath);
+                Directory.Delete(dirpath, true);
             }
         }
     }
--- a/Assets/Plugins/Source/Editor/Internal/BuildPackage.cs
+++ b/Assets/Plugins/Source/Editor/Internal/BuildPackage.cs
@@ -41,19 +41,21 @@ public static class BuildPackage
         /// <summary>
         /// Compressed directory of UPM contents
         /// </summary>
-        TarBall,
+        UPMTarBall,
     }
 
     /// <summary>
     /// Unless another path is explicitly provided, default to writing the output to a
     /// directory in the root of the project folder named "Build"
     /// </summary>
-    private const string OUTPUT_DIRECTORY = "Build";
+    private const string DEFAULT_OUTPUT_DIRECTORY = "Build";
 
     /// <summary>
     /// Note that this is a Non-Unity defined command line argument.
     /// </summary>
     private const string ARG_FLAG_OUTPUT = "-EOSPluginOutput";
+
+    private const string ARG_FLAG_PACKAGE_TYPE = "-PackageType";
 
     /// <summary>
     /// Returns a path to output the plugin to.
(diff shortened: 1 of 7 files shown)


Assets/Plugins/Source/Editor/Internal/UnityPackageCreationUtility.cs
```diff
@@ -0,0 +1,285 @@
...
+
+    /// <summary>
+    /// Contains section of package.json file pertaining to configuration
+    /// </summary>
+    public static PackagingConfigSection packageConfig;
+
+    /// <summary>
+    /// This is used in order to use StartCoroutine from a static context.
+    /// </summary>
+    public static CoroutineExecutor ExecutorInstance;
```
andrew-hirata-playeveryware: Could you kindly rename this variable to match the casing of the other fields?

Assets/Plugins/Source/Editor/Internal/UnityPackageCreationUtility.cs
```diff
@@ -0,0 +1,285 @@
...
+            File.WriteAllText(destPath, newContents);
+        }
+    }
+
+    
+
+    //-------------------------------------------------------------------------
+    public static void CreateUPMPackage(
+        string outputPath, 
+        string pathToJSONPackageDescription)
```
andrew-hirata-playeveryware: Can you move the params for this method to be on the same line?

Assets/Plugins/Source/Editor/Internal/UnityPackageCreationUtility.cs
```diff
@@ -0,0 +1,285 @@
...
+    {
+        UnityEngine.Debug.Log("DEBUG " + pathToJSONPackageDescription);
+        var JSONPackageDescription = File.ReadAllText(
+            pathToJSONPackageDescription
```
andrew-hirata-playeveryware: Passed params should not be broken up on multiple lines.

andrew-hirata-playeveryware (changes requested):

arthur740212 (approved):
Never knew you could do this outside of Unity.
This is really cool!

andrew-hirata-playeveryware (approved):

Assets/Plugins/Source/Editor/Internal/BuildPackage.cs
```diff
@@ -0,0 +1,226 @@
...
+        }
+
+        return path;
+    }
+
+    private static PackageType GetPackageType()
+    {
+        return PackageType.UPMDirectory;
+    }
+    private static void Clean()
```
pew-dahs, line 132: really minor nit: needs a space

pew-dahs (approved):

Assets/Plugins/Source/Editor/Internal/BuildPackage.cs
paulhazen: I can't believe my OCD brain didn't hit on this. Good catch.

PR #370, opened 2023-08-04
fix/ensure eos platform interface is not null

### LINKED ISSUE #367: The finalizer of the NotifyEventHandle class raises a Null exception during Editor DomainReload.
This problem does not occur in the actual build and can only be reproduced on UnityEditor.

NullException

The reason for this exception is that the finalizer of GenericSafeHandle, the base class of the NotifyEventHandle class, is called after the EOSPLatformInterface reference is lost.

NullException2

On UnityEditor, the finalizer is called at the next EntryPlayMode, so it has already lost reference to the EOSPlatformInterface. 

This error is not displayed as an error in the UnityEditor console, because Unity discards errors during DomainReload. 

However, the IDE's debugger detects this error, so debugging always stops at the exception when running debug.

TemporaryFix

To solve this problem, we need to add another null check to the callback in the initialization code within the EOS singleton, as shown in the screenshot above.

andrew-hirata-playeveryware:


arthur740212 (approved):

paulhazen (approved):
I forgot about the C# null-condition operator. I need to remember to make use of it, cause it's pretty slick.

PR #378, opened 2023-08-23
Exempt issues labeled "pending" from becoming marked stale and subsequently removed

paulhazen:
# Nuanced Automated Issue Closure Workflow:
Currently, when an issue has no activity on it after 7 days, it gets marked as "stale." If, after an additional 7 days there is still no activity, the issue is closed.

## Changes Made to Workflow:
- If the issue is marked `pending` or `bug` it will neither be labeled `stale`, nor will it be automatically closed.
- The number of days after which to mark an inactive issue has been increased from 7 to 14.
- _When_ an issue is closed automatically by this workflow, a new label `auto-closed` will be added to the issue

## Things to Consider:
- Because issues labeled `bug` are now exempt from being auto-marked stale and subsequently closed, it will be important for maintainers to carefully consider whether an issue truly is a bug, and removing the label if the issue doesn't truly meet the threshold of being a "bug" vs being a feature request.

### Flowchart of Auto-Close Logic:
image

### DIFF
--- a/.github/workflows/close_inactive_issues.yml
+++ b/.github/workflows/close_inactive_issues.yml
@@ -1,5 +1,6 @@
 name: Close inactive issues
 on:
+  workflow_dispatch:
   schedule:
     # Everyday at 1:30PM GMT+7
     - cron: "30 1 * * *"
--- a/.github/workflows/close_inactive_issues.yml
+++ b/.github/workflows/close_inactive_issues.yml
@@ -1,9 +1,8 @@
 name: Close inactive issues
 on:
-  workflow_dispatch:
-  #schedule:
-    # Everyday at 1:30PM GMT+7
-  #  - cron: "30 1 * * *"
+  schedule:
+    # Everyday at 1:30PM
+    - cron: "30 1 * * *"
 
 jobs:
   close-issues:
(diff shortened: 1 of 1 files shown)


andrew-hirata-playeveryware (approved):

.github/workflows/close_inactive_issues.yml
```diff
@@ -1,26 +1,49 @@
...
   schedule:
+    # Everyday at 1:30PM
     - cron: "30 1 * * *"
 
 jobs:
   close-issues:
     runs-on: ubuntu-latest
     permissions:
       issues: write
-      pull-requests: write
```
arthur740212, line 11: Is this where pull-requests don't get send notifications anymore?

arthur740212 (approved):

pew-dahs (approved):

.github/workflows/close_inactive_issues.yml
paulhazen: It's unrelated AFAIK. The gitbot would have to be configured in a really really really weird way for this to be the cause - particularly because this hasn't been merged into `development`, or `stable` yet. Also, the workflows only happen on the `stable` branch I believe (they only run on the default branch unless otherwise specified).

arthur740212: That is so right. Hmmm...

PR #382, opened 2023-08-28
Revised Doc on Doc

paulhazen:
<div align="center"></div>

## Revision Notes:
* The document that provides guidance on writing a README.md file _specifically_ still does not exist, I wanted to put this one on more solid footing first.
* Alerts do not appear to be capable of being indented, but I still have yet to see someone confirm that, and they _can_ be indented via list formatting, but I don't think that's the use case we were primarily concerned about with respect to re-formatting existing documentation that might be indented. I think that's something that can be addressed during the reformatting phase.
* I had a note indicating that the original document showed another example of a list type - but we were incorrect about that.


paulhazen:
Because this branch uses a different file for _my_ changes to the doc, I'm going to go ahead and merge this - I understand that there are still discussions to be had before it actually becomes a part of the documentation proper.

PR #384, opened 2023-08-28
Spelling & punctuation fixes for documentation

paulhazen:
While I did not change the content of the documentation, I went around fixing issues of spelling, grammar, punctuation, etc. In some places I very modestly rephrased things for clarity. 

This change does not introduce any changes to the documentation philosophy - just a bit of clean-up.

I was careful to go one document at a time and commit each file individually, so you can scan the commit history to get the 40,000ft view of the changes really quickly.


docs/macOS/macOS_supported_versions.md
```diff
@@ -1,11 +1,7 @@
 Compatible macOS versions for the EOS apk
 -------------------------------------------------------------------
 
+|Version|Standalone Build|Editor|
```
arthur740212, line 4: The table formatting didn't go through.
Adding a line of |----------|---------|----------| 
should do the trick.

docs/standards.md
```diff
@@ -1,42 +1,38 @@
...
 
-### Language Style Guides
+## Language Style Guides
 
-### For All source files
+### For all source files
 Each source code file needs to contain the license which it is licensed under.
 
-#### For C
+### For C / C++ / C#:
```
arthur740212, line 8: I remember @andrew-hirata-playeveryware mentioned this section was intended for C (and probably C++)?

arthur740212 (approved):

docs/macOS/macOS_supported_versions.md
paulhazen: Ah, good catch. You actually can just do single dashes - but that doesn't quite look as good in plane text format :\

docs/standards.md
paulhazen: Yeah - this is true - but of the standards that are actually listed on this page the same standards tend to apply, and previously there were empty sections for C and C++, with the only standards written beneath the "C#" section.

docs/egs/egs_readme.md
Raxnam, line 1: I think this should stay specify, although I may be reading it my head wrong

Raxnam (approved):
Looks pretty much good, it did make me realize the number of places we specify a unity version, but ill post those elsewhere as they are outside the scope of this PR

PR #385, opened 2023-08-31
Facepunch steamworks integration

arthur740212:
Facepunch is an alternative .net framework to use Steamworks functionality.
The two will clash if both exists however.

Current evil implementation:
The two frameworks share the same `SteamManager` name and some methods,
Only one `SteamManager`  would be enabled at any given time.


arthur740212:
Might be too evil for users.
Any thoughts would be highly appreciated

Assets/Scripts/FacePunch/FacePunchSteamManager.cs
```diff
@@ -0,0 +1,181 @@
...
+        // In case of disabled Domain Reload, reset static members before entering Play Mode.
+        [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.SubsystemRegistration)]
+        private static void InitOnPlayMode()
+        {
+            s_instance = null;
+        }
+#endif
+        protected virtual void Awake()
+        {
+            // Only one instance of DiscordManager at a time!
```
andrew-hirata-playeveryware, line 71: This comment needs to be updated

Assets/Scripts/FacePunch/FacePunchSteamManager.cs
```diff
@@ -0,0 +1,181 @@
...
+            if (s_instance != null)
+            {
+                Destroy(gameObject);
+                return;
+            }
+            s_instance = this;
+
+            // We want our DiscordManager Instance to persist across scenes.
```
andrew-hirata-playeveryware, line 79: Comment references discord manager

Assets/Scripts/FacePunch/FacePunchSteamManager.cs
```diff
@@ -0,0 +1,181 @@
...
+                Steamworks.SteamClient.Init(1679310);
+                Debug.LogWarning("[Facepunch Steamworks] Initialized");
+            }
+            catch (System.Exception e)
+            {
+                Debug.LogError(e.Message);
+                Debug.LogError("[Facepunch Steamworks] SteamAPI_Init() failed. Refer to Valve's documentation or the comment above this line for more information.", this);
+               
+                return;
+                // Something went wrong - it's one of these:
```
andrew-hirata-playeveryware, line 93: Is this comment supposed to be before the LogError? The LogError line on 90 seems to make reference to it.

Assets/Scripts/FacePunch/com.facepunch.steamworks.net.asmdef
andrew-hirata-playeveryware, line 1: Technically, this is asmdef is for our code, isn't it? If that's the case, the nameing convention should still follow ours, i.e. `com.playeveryware.eos.facepunch-steamworks` or something similar.

Assets/Scripts/Steam/SteamManager.cs
```diff
@@ -27,6 +27,8 @@
 //
 // Version: 1.0.12
 
+#if !ENABLE_FACEPUNCH_STEAM
```
andrew-hirata-playeveryware, line 30: I think it would be better to integrate this conditional compile into the one on line 32. Otherwise I think the 'fake' steamwork.net code won't work in this file, correct?

andrew-hirata-playeveryware (changes requested):

PR #386, opened 2023-09-01
UI Menu Consolidation

paulhazen:
<div align="center"></div>

## Simplified Menu

People use our plugin alongside many others in Unity. Grouping our menu options helps users find features faster and reduces clutter.

## Why Not `Edit`?

It might seem natural to put some options under `Edit`, but a separate menu adds clarity. Don't worry about the change—existing users won't be much affected, and it'll help new users.

> "Ah, changes!" - Expected reaction, but it sets us up for the long run.

## Doc Update

When I changed a menu item, I used `grep` on our `*.md` files to update any docs that referenced the menu item in question. This also highlighted that many menu items aren't even in our docs—something to address later.

## Misc Changes

- Removed `UIUGFitAnchors.cs` as it's unrelated to EOS functionality. Holler if I missed something.

## Sorting Issue

- The submenu item order is off—like "Version" should probably be at the bottom. Thoughts?

<div align="center">image</div>


Raxnam (approved):
I'm not too familiar with these side of the project, but it looks like it shoudln't cause any issues

pew-dahs (approved):

PR #387, opened 2023-09-01
Dev/flag fix

paulhazen:
Refactoring the way `EOS_PREVIEW_PLATFORM` is managed to improve the developer experience on unsupported platforms. Previously, adding this script symbol in the player settings was cumbersome and affected version-controlled files.

After analyzing its usage, it became clear that the flag was intended to make sure developers are aware they're on a preview platform. However, this approach led to silent build failures, offering no hints to set the flag.

To maintain user awareness without causing disruptive issues, I've added warning log statements for both macOS and Linux platforms to indicate the platform's preview status.

### DIFF
--- a/Assets/Plugins/Source/Core/EOSManager_DynamicLoading.cs
+++ b/Assets/Plugins/Source/Core/EOSManager_DynamicLoading.cs
@@ -67,34 +67,30 @@ namespace PlayEveryWare.EpicOnlineServices
 
             public const string EOSBinaryName = Epic.OnlineServices.Config.LibraryName;
 
+
 #if USE_EOS_GFX_PLUGIN_NATIVE_RENDER
-//#if UNITY_WSA
-                        //delegate void UnloadEOS_delegate();
-                        //delegate IntPtr EOS_GetPlatformInterface_delegate();
-                        //static UnloadEOS_delegate UnloadEOS;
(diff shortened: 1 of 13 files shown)


Assets/Plugins/Source/Core/EOSManager_DynamicLoading.cs
```diff
@@ -67,34 +67,30 @@ public partial class EOSSingleton
 
             public const string EOSBinaryName = Epic.OnlineServices.Config.LibraryName;
 
+
 #if USE_EOS_GFX_PLUGIN_NATIVE_RENDER
-//#if UNITY_WSA
```
arthur740212, line 71: Not sure if this is disabled temporarily or just not used anymore.

paulhazen: Pretty sure it's not used anymore - looking at the rest of the file (for instance `static extern void UnloadEOS()` on line `85`).

paulhazen: In fact - I just looked through the history - these lines have been commented out for over a year now - so I'm confident they can be removed.

arthur740212 (approved):

Raxnam (approved):

PR #409, opened 2023-09-22
Fix .unitypackage creation bug

paulhazen:
Prior to these changes, creating the package via `Tools -> EOS Plugin -> Create Package` and selecting `Create .unitypackage` button would result in a `NullReferenceException`.

These changes fix that and simplify the `Create Package` window by removing the custom build output directory field, and using instead just the output directory, and "hid" the package JSON file selection under an "Advanced" carrot. 

This is what the new window looks like:

image

This is what it used to look like:

image


Raxnam (approved):

Jason-J-PEW (approved):
Looks good to me. The new UI is an improvement.

PR #417, opened 2023-09-26
Setup Scripts

paulhazen:
This PR introduces a series of scripts that help set up development environments suitable for making contributions to the EOS Plugin. 

The intent is that for each platform (Windows, MacOS, and Linux) the corresponding `setup-*` script will automatically install as many of the pre-requisites as are possible for the given platform.

Currently, only `setup-windows.ps1` has been fully tested. 

`setup-windows.ps1` script installs the following:
- Package manager `winget`
- UnityHub
- Whichever version of Unity Editor is specified for the project
- Visual Studio 2019 Community Edition

On a blank (or relatively clean) machine, after running this script, it should be possible to open the project and build the sample scenes without any additional software.


scripts/etc/Brewfile
andrew-hirata-playeveryware, line 1: This is a non-canonical location for this file, so we might need to modify documentation to ensure that it it's obvious that we have this file to assist contributors with dependency installing.

scripts/setup-linux.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# Add source for unity hub
+wget -qO - https://hub.unity3d.com/linux/keys/public | gpg --dearmor | sudo tee /usr/share/keyrings/Unity_Technologies_ApS.gpg > /dev/null
+sudo sh -c 'echo "deb [signed-by=/usr/share/keyrings/Unity_Technologies_ApS.gpg] https://hub.unity3d.com/linux/repos/deb stable main" > /etc/apt/sources.list.d/unityhub.list'
+
+# Update apt-get 
+sudo apt-get update
+
+# Install prereqs
+sudo dnf install openssl1.1 openssl-libs
```
andrew-hirata-playeveryware: Isn't dnf a redhat thing, while the rest of these are debian based package installer things?

scripts/setup-macos.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
andrew-hirata-playeveryware: If we're going to be pedantic, I'd suggest 
`!/usr/bin/env bash` instead of `/bin/bash`. I can't remember if the location of 'bash' is "guaranteed" to be in `/bin/`, unlike 'sh`, which I believe is required to be there by some POSIX-like standard.

andrew-hirata-playeveryware (changes requested):

scripts/setup-macos.sh
paulhazen: Nice catch! I'm surprised that it's not guaranteed on mac, but oh well :)

scripts/setup-linux.sh
paulhazen: Interesting - this is an artifact of me putting together several script examples I found. I think this line can be removed though. I'll test it and make the change.

scripts/etc/Brewfile
paulhazen: Makes sense - however right now we don't provide any documentation anywhere that references the file. Might make more sense to just use a bash script anyways, since we can't be positive that brew is installed in the first place, which is why I open the Brewfile _via_ a bash script.

scripts/setup-linux.sh
paulhazen: Confirmed that that line in the script is superfluous.

arthur740212 (approved):
I am not that well versed in shell scripts, but overall the scripts look great :D

andrew-hirata-playeveryware (approved):

PR #419, opened 2023-09-29
Tidy Pt. 1

paulhazen:
Second attempt at doing some tidy of the directory structure. Limited the changes so as to not create a massive PR.

# Cleaning up directory structure

This PR does not alter any functionality of the project, but changes where some things exist in the repository. There were a lot of items that we had in the root of the repository that more appropriately should have been placed somewhere else. 

> [!NOTE] 
> Code was only updated to reflect the new location of the moved files, and all docs were updated with new file locations.

These are the itemized changes that were made - each one happens in a single commit:

## Deleted files

- `EOSZipInstallDescriptionConfigs`
  _Neither this directory, nor any of the files contained within are referenced anywhere in the project_

- `UnityHelpers_Android`
 _This tool is no longer used by the project._

## Moved files

`bin`⟶ `tools/bin` 
`accessible-urls.txt` ⟶ `etc/accessible-urls.txt`
`EOSUnityPlugin_package_template` ⟶ `etc/PackageTemplate`

### etc
The `etc` directory should contain any files related to the configuration of the project. These are typically JSON files or other small pieces of data that help define properties of the project and how it is to be built.

### tools
The `tools` directory contains utilities that are used outside of the plugin but are part of the plugin's development pipeline. Some of these tools exist in our repository as binary executable files, and some of these tools exist in the form of source code that is meant to be compiled. The former category lives in `tools/bin`, and the latter resides in `tools/src`.


EOSZipInstallDescriptionConfigs/eos_public_platforms_description.json
andrew-hirata-playeveryware, line 1: This file is being used: it's use is to allow for 'easy' installing of just the files we need from the zip file that Epic has for their EOS SDK downloads.

Jason-J-PEW (commented):
I agree with the premise of moving things out of the root of the project for better organization. 

However, I don't know enough about these things to determine if the changes here are done correctly. I'm also nervous about deleting things that I know nothing about.

So I will defer to the other members of the team for this approval.

andrew-hirata-playeveryware (changes requested):

Raxnam (approved):

EOSZipInstallDescriptionConfigs/eos_public_platforms_description.json
paulhazen: I re-introduced the file to `etc/eos_plublic_platforms_description.json`

andrew-hirata-playeveryware (approved):

PR #424, opened 2023-10-04
Ingest getting started doc feedback

arthur740212:


Raxnam (approved):

paulhazen (approved):
Looks good!

Is there any particular reason you transitioned to adding strings:

`Application.dataPath + "/../etc/PlatformSpecificAssets/EOS/Android/`

instead of:

`Path.Combine(Application.dataPath, "../etc/PlatformSpecificAssets/EOS/Android/")`

?

I don't think it's a problematic change at all, just curious.

arthur740212:
I read on some discussions that `FileUtil.CopyFileOrDirectory()` only allows `/` in file paths,
while `Path.Combine` merges the strings with `\`.
I am not sure if it really breaks with `\`. But with `/` it does work.

PR #453, opened 2023-11-02
Fix: Do not receive a packet of size 0

Jason-J-PEW:
This fixes a rare crash that happened when `GetNextReceivedPacketSize` returned a size of 0 but `ReceivePacket` would write some number of bytes to a 0 byte buffer.

### DIFF
--- a/Assets/Scripts/EOSPeer2PeerManager.cs
+++ b/Assets/Scripts/EOSPeer2PeerManager.cs
@@ -306,6 +306,11 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
                 RequestedChannel = null
             };
             P2PHandle.GetNextReceivedPacketSize(ref getNextReceivedPacketSizeOptions, out uint nextPacketSizeBytes);
+            
+            if (nextPacketSizeBytes == 0)
+            {
+                return null;
+            }
 
             byte[] data = new byte[nextPacketSizeBytes];
             var dataSegment = new ArraySegment<byte>(data);
--- a/Assets/Scripts/Networking/EOSTransportManager.cs
+++ b/Assets/Scripts/Networking/EOSTransportManager.cs
@@ -1112,6 +1112,15 @@ namespace PlayEveryWare.EpicOnlineServices.Samples.Network
             };
             P2PHandle.GetNextReceivedPacketSize(ref getNextReceivedPacketSizeOptions, out uint nextPacketSizeBytes);
 
+            if (nextPacketSizeBytes == 0)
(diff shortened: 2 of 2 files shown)


Assets/Scripts/Networking/EOSTransportManager.cs
```diff
@@ -1112,6 +1112,15 @@ public bool TryReceivePacket(out ProductUserId remoteUserId, out string socketNa
...
             P2PHandle.GetNextReceivedPacketSize(ref getNextReceivedPacketSizeOptions, out uint nextPacketSizeBytes);
 
+            if (nextPacketSizeBytes == 0)
+            {
+                remoteUserId = null;
+                socketName = null;
+                channel = 0;
+                packet = null;
+                return false;
+            }
```
paulhazen, line 1122: Can't comment on other sections of code, but this is adjacent. Would it not make sense to move lines 1101-1106 to be below this section of code? Since `receivePackatOptions` isn't used until later in the function?

Assets/Scripts/EOSPeer2PeerManager.cs
```diff
@@ -306,6 +306,11 @@ public ProductUserId HandleReceivedMessages()
                 RequestedChannel = null
             };
             P2PHandle.GetNextReceivedPacketSize(ref getNextReceivedPacketSizeOptions, out uint nextPacketSizeBytes);
+            
+            if (nextPacketSizeBytes == 0)
+            {
+                return null;
+            }
```
paulhazen, line 313: Same as my other comment, shouldn't lines 296-301 be moved to be _after_ line 313? Sort of a nit-picky detail, which is why I'm not requesting changes. But curious if I've missed some detail.

paulhazen (approved):
Had some questions about the order of some statements, but none that rise to the level of needing changes. 

Really, really good work on finding this.

Raxnam (approved):

Assets/Scripts/Networking/EOSTransportManager.cs
Jason-J-PEW: Yes that would make sense to do. I somehow didn't see this before I merged this PR in. But that was how the code was already laid out. I agree with you though. I always put the options struct directly above the EOS SDK call that uses it.

Assets/Scripts/EOSPeer2PeerManager.cs
Jason-J-PEW: Yeah, you're right, same as above.

PR #458, opened 2023-11-06
Feat: Shutdown callbacks added to EOS Manager

Raxnam:
Adds a list of callbacks that individual managers can add to to make sure they have specific code run before the main manager shuts down. This Fixes the P2P crash on close bug

### DIFF
--- a/Assets/Plugins/Source/Core/EOSManager.cs
+++ b/Assets/Plugins/Source/Core/EOSManager.cs
@@ -320,6 +320,11 @@ namespace PlayEveryWare.EpicOnlineServices
                 s_onAuthLogoutCallbacks.Add(authLogout.OnAuthLogout);
             }
 
+            public void AddApplicationCloseListener(Action listener)
+            {
+                s_onApplicationShutdownCallbacks.Add(listener);
+            }
+
             public void RemoveConnectLoginListener(IEOSOnConnectLogin connectLogin)
             {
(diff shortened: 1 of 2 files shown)


Assets/Plugins/Source/Core/EOSManager.cs
```diff
@@ -332,6 +335,11 @@ public void RemoveAuthLogoutListener(IEOSOnAuthLogout authLogout)
                 s_onAuthLogoutCallbacks.Remove(authLogout.OnAuthLogout);
             }
 
+            public void AddApplicationCloseListener(Action listener)
```
Jason-J-PEW: This should be up on line 322 after the other add functions. And a remove function should be added. Just to maintain consistency with what is already there.

Jason-J-PEW (changes requested):
Only one minor code cleanliness thing to take care of.

Assets/Plugins/Source/Core/EOSManager.cs
andrew-hirata-playeveryware: I agree.

andrew-hirata-playeveryware: I agree.

andrew-hirata-playeveryware (changes requested):

andrew-hirata-playeveryware (approved):

Jason-J-PEW (approved):

PR #462, opened 2023-11-10
Fix platform native lib output path

arthur740212:
Following the folder rearrangements, native libraries that used to be built directly in place went rouge, this PR fixes that.


paulhazen (approved):
Nice catch @arthur740212, I'm certain this was a regression introduced by me. In the future, I'll be more careful to catch issues like this, thanks for fixing it for us!

arthur740212:
You did a great job at the folder rearrangements. If I was making those changes, I wouldn't notice these sneaky ones either. 
I didn't even know you could do post build commands in an sln. 
Glad it worked out and we learned something new as a bonus.

PR #478, opened 2023-11-23
 Import and Export descriptions rename

arthur740212:
This is to easier tell what is for what.
Import : Imports cs scripts and assets from EOSSDK, and native libs from public platforms
Export : Exports an UPM from our project.


Assets/Plugins/Source/Editor/Internal/InstallEpicOnlineServiceZip.cs
```diff
@@ -95,23 +98,90 @@ public string ToCapitalize(string str)
...
+                }
+            }
+
+            if (GUILayout.Button("Select All", GUILayout.MaxWidth(100)))
+            {
+                foreach (var platformImportInfo in importInfoList.platformImportInfoList)
+                {
+                    {
+                        platformImportInfo.isGettingImported = true;
+                    }
```
paulhazen: Why did you enclose this in extra curly brackets?

paulhazen (changes requested):
I have a few thoughts, some of which pertain to this PR, some of which do not apply to _this_ PR, but do apply to the feature in question.

### Related to this PR

I think as a rule, whenever we introduce UI changes, we should include a before / after screenshot to demonstrate the change. That makes it a lot easier to review as well.

I think I only had one question regarding code changes - my main request is to include before / after screenshots in the PR.

### _Not_ related to this PR, but worth discussing:

I think we should remove this tool from the menu system for the following reasons:

1. I think it's unclear that this code (mainly existing in `InstallEpicOnlineServiceZip.cs`) is used to _import_ a new drop of the EOS SDK. While it is a tool we ourselves no doubt make use of, because we ship the plugin with the latest and greatest SDK, 99.99% of our users should **not** need it. Therefore, if we can remove it from the menu system of the plugin, we can make using the plugin much more straightforward.

1. Another reason to get rid of this tool is that when we make a release, we cannot guarantee that the release will work with any version of the EOS SDK other than the one we ship it with, so providing a method for the SDK to be changed by the user has the risk of providing an unstable product.

A possible solution could be to create another sub-menu like "Advanced," wherein we can place this tool, renamed to be called "Tools/EOS Plugin/Advanced/Change EOS SDK Version...".

Assets/Plugins/Source/Editor/Internal/InstallEpicOnlineServiceZip.cs
arthur740212: Oops, that is a mistake, I copied pasted got rid of the ifs and elses but missed the brackets. Thanks for pointing this out.

arthur740212:
> ### Related to this PR
> I think as a rule, whenever we introduce UI changes, we should include a before / after screenshot to demonstrate the change. That makes it a lot easier to review as well.

Totally agree, I missed this part.

image


> ### _Not_ related to this PR, but worth discussing:
> I think we should remove this tool from the menu system for the following reasons:
> 
> 1. I think it's unclear that this code (mainly existing in `InstallEpicOnlineServiceZip.cs`) is used to _import_ a new drop of the EOS SDK. While it is a tool we ourselves no doubt make use of, because we ship the plugin with the latest and greatest SDK, 99.99% of our users should **not** need it. Therefore, if we can remove it from the menu system of the plugin, we can make using the plugin much more straightforward.
> 2. Another reason to get rid of this tool is that when we make a release, we cannot guarantee that the release will work with any version of the EOS SDK other than the one we ship it with, so providing a method for the SDK to be changed by the user has the risk of providing an unstable product.
> 
> A possible solution could be to create another sub-menu like "Advanced," wherein we can place this tool, renamed to be called "Tools/EOS Plugin/Advanced/Change EOS SDK Version...".

These are great considerations for users, though this tool wasn't included in the packages we release, it will only be used by us and contributors who clone our repo.
Moving to another sub-menu sounds cool, and a rename to better describe the tool is better too.

arthur740212:
> This is to easier tell what is for what. Import : Imports cs scripts and assets from EOSSDK, and native libs from public platforms Export : Exports an UPM from our project.

There are more changes after the initial PR,
here is a better description for the new feature.

Our file configs for import tools used to be single jsons file that copies scripts for every platform.
While our repo actually has multiple description files with a lot of overlaps between them.
such that every change to the SDK we will have to modify same content across every file.
The changes are to modularize the descriptions, and also to batch install multiple platforms of content.

paulhazen (approved):

Raxnam (approved):

Issue #480, opened 2023-11-27
Error "Player data storage: could not receive data: too much of it."


GalBorSan:
**Describe the bug**
Error "Player data storage: could not receive data: too much of it." is triggered by the ReceiveData method of the EOSPlayerDataStorage.cs file

**To Reproduce**
1- Save a file to cloud (named gabo_visionpictures_20231123131149). The callback log Success ok is triggered.
2- Try to get the file from cloud using EOSPlayerDataStorage::StartFileDataDownload method.
3- The log Error "[EOS SDK] Player data storage: could not receive data: too much of it." is triggered from EOSPlayerDataStorage::ReceiveData method.

We can't read a file downloaded from cloud. The transfers progress seem to be completed because this log "[EOS SDK] Player data storage: transfer progress 27444 / 27444" is triggered before the error. It's only 27,4 KB so is not too much.

**Expected behavior**
To be able to download a file previously saved in the cloud without problems.

**Desktop (please complete the following information):**
 - OS: [Windows x64]
 - Standalone build
 - Unity Version: 2021.3.0f1
 - Plugin Version: 3.0.1
 - Epic Online Services Version: 1.16.0-27024038

paulhazen:
Can confirm this is a bug. Working on a fix now, will update with PR shortly.

paulhazen:
PR #481 resolves this issue. Once it's merged it will be reflected in the `development` branch, and after a few weeks or so it will be in a new release. Let me know if you need assistance exporting the plugin from the development branch.

paulhazen:
It's been merged. Marking this issue as "pending" until the change gets into `stable`

paulhazen:
I'm closing this issue as resolved. It's unclear to me whether the solution is yet in the `stable` branch, but the solution has been tested, and will certainly be in `stable` shortly.

PR #481, opened 2023-11-27
Fixed bug where files over 4096 bytes would fail to download.

paulhazen:
This PR fixes a bug in the player storage implementation where any file over the `MAX_CHUNK_SIZE` (4096) bytes would fail to download. The logic of the code was faulty. It has since been updated. I also changed the maximum file size to reflect the usage restrictions dictated by the EOS SDK documentation (files are allowed to be up to 200MB in size). 

This resolves issue #480

### DIFF
--- a/Assets/Scripts/EOSPlayerDataStorageManager.cs
+++ b/Assets/Scripts/EOSPlayerDataStorageManager.cs
@@ -29,6 +29,8 @@ using UnityEngine;
 
 using Epic.OnlineServices;
 using Epic.OnlineServices.PlayerDataStorage;
+using static Codice.Client.Common.WebApi.WebApiEndpoints;
+using static UnityEditor.PlayerSettings;
 
 namespace PlayEveryWare.EpicOnlineServices.Samples
 {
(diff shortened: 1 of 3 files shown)


Raxnam (approved):

Assets/Scripts/EOSPlayerDataStorageManager.cs
```diff
@@ -29,6 +29,8 @@
 
 using Epic.OnlineServices;
 using Epic.OnlineServices.PlayerDataStorage;
+using static Codice.Client.Common.WebApi.WebApiEndpoints;
```
arthur740212: Is this necessary for the fixes?

Assets/Scripts/EOSPlayerDataStorageManager.cs
```diff
@@ -29,6 +29,8 @@
...
+using static UnityEditor.PlayerSettings;
```
arthur740212: Same as above, probably isn't ideal for editor code to exist in manager script.

Assets/Scripts/EOSPlayerDataStorageManager.cs
paulhazen: Nope, just removed them. Must have been introduced between my experimentation and the finished fix.

arthur740212 (approved):
Looking great.
Side note, we should do a comparison with the data code in Title Storage.

PR #491, opened 2023-12-12
refactor(rename): Rename invalidAuthType to connect to be more clear

Jason-J-PEW:
While implementing the plugin, I found that this name was confusing. I was looking for examples of how to do Connect and there wasn't a Connect type. It took a while to figure out that invalid meant connect, not an actually invalid login type. I believe this name is more clear.


andrew-hirata-playeveryware (approved):

arthur740212 (approved):

paulhazen (approved):
Good change.

Issue #456, opened 2023-11-05
EOSPluginEditorConfigEditor and EpicOnlineServicesConfigEditor windows start piling up over time


julcreutz:
**Describe the bug**
Over time, multiple instances of the aforementioned editor window are being created and they are causing editor exceptions.

**To Reproduce**
1. Work inside Unity
2. Maximize the console window and un-maximize it
3. After some time, the errors start popping up

**Expected behavior**
Only one instance of either window is possible, without exceptions being caused.

**Screenshots**
<img width="684" alt="image" src="https://github.com/PlayEveryWare/eos_plugin_for_unity/assets/24528976/4387cef3-95ed-4c19-a424-3cb26e831c38">

<img width="1168" alt="image" src="https://github.com/PlayEveryWare/eos_plugin_for_unity/assets/24528976/4a3f9426-fa02-4e47-9503-d8fd14c5855a">

**Desktop (please complete the following information):**
 - OS: macOS ARM
 - Editor
 - Unity 2022.3.10f1
 - 2.3.3

arthur740212:
We weren't able to reproduce the problem on either v2.3.3 or the latest v3.0.2.
Was there extra steps needed to hit the issue? (Opening the config editors for instance)

julcreutz:
I myself haven't found an easy way to reproduce it. When I reset my Unity layout, it stops happening for a while, but then happens again. I don't even make use of any of the epic APIs as of right now, the plugin is just sitting in the project.
I don't open the config editors at all, nor are they opened in some tab. It's super weird.
I will try upgrading to 3.0.2 and see if the issue persists.

paulhazen:
> I myself haven't found an easy way to reproduce it. When I reset my Unity layout, it stops happening for a while, but then happens again. I don't even make use of any of the epic APIs as of right now, the plugin is just sitting in the project. I don't open the config editors at all, nor are they opened in some tab. It's super weird. I will try upgrading to 3.0.2 and see if the issue persists.

Did upgrading resolve the issue for you?

julcreutz:
> > I myself haven't found an easy way to reproduce it. When I reset my Unity layout, it stops happening for a while, but then happens again. I don't even make use of any of the epic APIs as of right now, the plugin is just sitting in the project. I don't open the config editors at all, nor are they opened in some tab. It's super weird. I will try upgrading to 3.0.2 and see if the issue persists.
> 
> Did upgrading resolve the issue for you?

[...]

PR #503, opened 2024-01-03
Fix: Editor Windows pile up over time

paulhazen:
This PR addresses issue #456.

## Changes & Reasoning

In Unity, when an `EditorWindow` is not explicitly marked as `[Serializable]`, or if the implementing class makes use of `static` resources in a manner that is not thread-safe, the Unity Editor can get into a state where it becomes flooded with exceptions indicating that one or more editor windows are invalid.

To fix this, some of the editor windows needed to be restructured to avoid use of static methods and rely instead on the much more linear and straightforward instance classes.

In the process of doing this, I created an abstract class `EOSEditorWindow`, which contains functionality common to all the `EditorWindow` classes we implement in the project. I took the opportunity to implement in that base class a method for auto-resizing the editor windows based on their contents, as well as enforcing a consistent padding around the window border, which helps other issues that result from controls within the window not being reachable.

When making these changes, I noticed that the exception "Should not be capturing when there is a hotcontrol" would sometimes happen as a warning - which is something that I was not able to replicate, but I also don't know why it was happening. So, when you test this, please keep an eye out for that.

> [!IMPORTANT] 
> I went through this PR and commented on areas where the reasoning for the change may not be immediately clear, hopefully this helps ease the code review process.


Assets/Plugins/Android/Editor/PlatformSpecificConfigEditorAndroid.cs
```diff
@@ -28,34 +28,11 @@
...
-        public void Awake()
-        {
-            var configFilenamePath = EpicOnlineServicesConfigEditor.GetConfigPath(ConfigFilename);
-            configFile = new EOSConfigFile<EOSAndroidConfig>(configFilenamePath);
-        }
-
-        public void LoadConfigFromDisk()
-        {
-            configFile.LoadConfigFromDisk();
-        }
```
paulhazen, line 56: These are all examples of functionality that was moved to a base class (`ConfigEditor`).

Assets/Plugins/Android/Editor/PlatformSpecificConfigEditorAndroid.cs
```diff
@@ -28,34 +28,11 @@
...
     {
-        public static string ConfigFilename = "eos_android_config.json";
-        EOSConfigFile<EOSAndroidConfig> configFile;
+        public ConfigSectionAndroid() : base("Android", "eos_android_config.json") { }
 
-        [InitializeOnLoadMethod]
-        static void Register()
-        {
-            EpicOnlineServicesConfigEditor.AddPlatformSpecificConfigEditor(new PlatformSpecificConfigEditorAndroid());
-        }
```
paulhazen, line 40: Instead of allowing Unity to control when different config editors are added to our editor window, we allow the `EditorWindow` instance to instantiate the ones it needs when it opens.

Assets/Plugins/Android/Editor/PlatformSpecificConfigEditorAndroid.cs
```diff
@@ -69,29 +46,24 @@ public void OnGUI()
...
             }
 
-            EpicOnlineServicesConfigEditor.AssigningFlagTextField("Override Platform Flags (Seperated by '|')", ref overrideValues.platformOptionsFlags, 250);
+            GUIEditorHelper.AssigningFlagTextField("Override Platform Flags (Seperated by '|')", ref overrideValues.platformOptionsFlags, 250);
 
-            EpicOnlineServicesConfigEditor.AssigningFloatToStringField("Override initial button delay for overlay", ref overrideValues.initialButtonDelayForOverlay, 250);
+            GUIEditorHelper.AssigningFloatToStringField("Override initial button delay for overlay", ref overrideValues.initialButtonDelayForOverlay, 250);
 
-            EpicOnlineServicesConfigEditor.AssigningFloatToStringField("Override repeat button delay for overlay", ref overrideValues.repeatButtonDelayForOverlay, 250);
+            GUIEditorHelper.AssigningFloatToStringField("Override repeat button delay for overlay", ref overrideValues.repeatButtonDelayForOverlay, 250);
```
paulhazen, line 53: Many static calls to the `EditorWindow`-derived `EpicOnlineServicesConfigEditor` class were being used across all the `EditorWindow` classes to draw GUI controls. Instead of having _one_ `EditorWindow`-derived class contain all these methods, I moved them to a new static class `GUIEditorHelper`.

Assets/Plugins/Linux/Editor/PlatformSpecificConfigEditorLinux.cs
```diff
@@ -27,34 +27,11 @@
 
 namespace PlayEveryWare.EpicOnlineServices
 {
-    public class PlatformSpecificConfigEditorLinux : IPlatformSpecificConfigEditor
+    public class ConfigSectionLinux : ConfigSection<EOSLinuxConfig>
```
paulhazen, line 30: This is an artifact of renaming the interface `IPlatformSpecificConfigEditor` to ultimately be an abstract class `ConfigSection` because ReSharper wanted to be helpful and rename more than I asked it to. Consequently, the filename no longer matches the class name - so this should be fixed and is reason enough to request changes.

Assets/Plugins/Android/Editor/PlatformSpecificConfigEditorAndroid.cs
```diff
@@ -28,34 +28,11 @@
 namespace PlayEveryWare.EpicOnlineServices
 {
 
-    public class PlatformSpecificConfigEditorAndroid : IPlatformSpecificConfigEditor
+    public class ConfigSectionAndroid : ConfigSection<EOSAndroidConfig>
```
paulhazen: This is an artifact of renaming the interface `IPlatformSpecificConfigEditor` to ultimately be an abstract class `ConfigEditor` because ReSharper wanted to be helpful and rename more than I asked it to. Consequently, the filename no longer matches the class name - so this should be fixed and is reason enough to request changes.

Assets/Plugins/Source/Editor/IConfigEditor.cs
```diff
@@ -22,18 +22,56 @@
...
-        void LoadConfigFromDisk();
+        void Read();
 
-        void SaveToJSONConfig(bool prettyPrint);
+        void Save(bool prettyPrint);
 
         void OnGUI();
     }
+
+    public abstract class ConfigEditor<T> : IConfigEditor where T : IEmpty, ICloneableGeneric<T>, new()
```
paulhazen, line 59: It's a little unconventional to have the interface and an implementing abstract class in the same file - but this practice exists across the project, and I figured that resolving that pattern should be done in one PR that doesn't also include functional changes.

Assets/Plugins/Source/Editor/Internal/DeploymentChecker.cs
paulhazen, line 1: Just a warning, having looked through the PR: I found the diff (as rendered here by GitHub) to be very confusing - the way it's rendered makes it look like lots of new things were added, which is not the case. I suggest looking at the file as a whole without the diff showing.

*Update:* I discovered this is because I had whitespace changes turned on in GitHub. In case you missed my other comment on the topic, you'll want to check this box:

image

and reload for a clearer rendering of the difference.

Assets/Plugins/Source/Editor/Internal/MakefileUtil.cs
```diff
@@ -46,55 +46,25 @@ public bool IsEmpty()
...
-
-        public string GetNameForMenu()
-        {
-            return "Platform Library Build Settings";
-        }
-
-        public void LoadConfigFromDisk()
-        {
-            configFile.LoadConfigFromDisk();
-        }
```
paulhazen, line 79: This is another concise example of the functionality that was moved from the most derived class to the base abstract class `ConfigEditor<T>`.

Assets/Plugins/Source/Editor/Internal/PackageFileUtils.cs
```diff
@@ -199,23 +199,6 @@ public static List<FileInfoMatchingResult> GetFileInfoMatchingPackageDescription
...
-                                toolsSection.LoadConfigFromDisk();
-                            }
-                            var pathToSignTool = toolsSection.GetCurrentConfig().pathToEACIntegrityTool;
-                            var pathToDefaultCertificate = toolsSection.GetCurrentConfig().pathToDefaultCertificate;
-                        }
-                        else
-                        {
-                            Debug.LogError("No tools configuration set!");
-                        }
-                    }
```
paulhazen, line 218: This code section was removed because it does nothing. If you evaluate it carefully, you'll see that it creates a variable `pathToSignTool` and `pathToDefaultCertificate` but doesn't do anything with either of them. Lack of documentation around this makes it unclear what the intended functionality was when it was written, so it was removed to avoid future confusion - but remains in the `git` record for historical context.

Assets/Plugins/Source/Editor/EpicOnlineServicesConfigEditor.cs
```diff
@@ -31,29 +31,12 @@
...
-            if (platformSpecificConfigEditors == null)
-            {
-                platformSpecificConfigEditors = new List<IPlatformSpecificConfigEditor>();
-            }
-            platformSpecificConfigEditors.Add(platformSpecificConfigEditor);
-        }
-
-        private static string IntegratedPlatformConfigFilenameForSteam = "eos_steam_config.json";
-
-        static List<IPlatformSpecificConfigEditor> platformSpecificConfigEditors;
```
paulhazen: It was necessary to remove these static members / methods in order to make our editor window play nice with Unity.

Their equivalent functionality was implemented in instance methods - except for `EncryptionKeyRegex` which was unused.

Assets/Plugins/Source/Editor/EpicOnlineServicesConfigEditor.cs
```diff
@@ -70,20 +53,15 @@ public static void AddPlatformSpecificConfigEditor(IPlatformSpecificConfigEditor
         [MenuItem("Tools/EOS Plugin/Dev Portal Configuration")]
         public static void ShowWindow()
         {
-            GetWindow(typeof(EpicOnlineServicesConfigEditor), false, "EOS Config Editor", true);
+            GetWindow<EpicOnlineServicesConfigEditor>("EOS Config Editor");
```
paulhazen: In all cases where we utilize the `GetWindow` function, calling the templated method as done here provides identical behavior, and in a manner that is much clearer.

Assets/Plugins/Source/Editor/EpicOnlineServicesConfigEditor.cs
```diff
@@ -193,27 +185,22 @@ private void Awake()
...
                 i++;
             }
 
             LoadConfigFromDisk();
         }
 
-        private bool DoesHaveUnsavedChanges()
-        {
-            return false;
-        }
```
paulhazen, line 207: In every instance where `DoesHaveUnsavedChanges` was defined, all implementations merely returned false, making it functionally meaningless, so they were all removed for the sake of clarity or until we have a clear way of implementing such a feature.

Assets/Plugins/Source/Editor/EpicOnlineServicesConfigEditor.cs
```diff
@@ -225,269 +212,6 @@ private void SaveToJSONConfig(bool prettyPrint)
...
 
-        public static string CollectFlags(List<string> flags)
-        {
-            return String.Join("|", EmptyPredicates.NewIfNull(flags));
-        }
-
-        public static List<string> SplitFlags(string collectedFlags)
-        {
-            return new List<string>(collectedFlags.Split('|'));
-        }
```
paulhazen, line 236: These one-line functions were only ever referenced in one place apiece, so I removed them - replacing them with the function body.

Assets/Plugins/Source/Editor/EpicOnlineServicesConfigEditor.cs
```diff
@@ -225,269 +212,6 @@ private void SaveToJSONConfig(bool prettyPrint)
...
-            HorizontalLine(color, defaultHorizontalLineStyle);
-        }
-
-        public static void HorizontalLine(Color color, GUIStyle guiStyle)
-        {
-            var currentColor = GUI.color;
-            GUI.color = color;
-            GUILayout.Box(GUIContent.none, guiStyle);
-            GUI.color = currentColor;
-        }
```
paulhazen, line 490: All of these static methods are used to create GUI elements via UnityEditor's ImGui-based UI system. These static methods were defined in _this_ specific `EditorWindow`-derived class but utilized by all classes that derived from `EditorWindow`, so they were moved to the new `GUIEditorHelper` static class.

Assets/Plugins/Source/Core/EOSManager.cs
paulhazen, line 1: Very annoyingly, this is a LF / CF diff. I'm not certain how it was created, and I'm not sure how to undo it. Not a lot changed in this file, so the rendered diff is misleading. You can skip those changes by checking this box:

image

Most of the non-whitespace changes are simplifications of namespaces accomplished via a tool. I was getting lost in this file specifically with the number of fully-qualified namespaces that were being used. I found that simplifying them greatly improved the readability.

Assets/Plugins/Source/Core/EOSManager.cs
```diff
@@ -151,18 +165,18 @@ enum EOSState
...
+            static private ProductUserId s_localProductUserId;
 
             static private NotifyEventHandle s_notifyLoginStatusChangedCallbackHandle;
             static private NotifyEventHandle s_notifyConnectLoginStatusChangedCallbackHandle;
             static private NotifyEventHandle s_notifyConnectAuthExpirationCallbackHandle;
             static private EOSConfig loadedEOSConfig;
 
             // Setting it twice will cause an exception
-            static bool hasSetLoggingCallback = false;
+            static bool hasSetLoggingCallback;
```
paulhazen, line 176: These kind of changes are an artifact of the tool I used to shorten namespaces - it also removed redundant code. In C#, initializing a variable to something other than it's default value is not necessary - even when the variable is `static`.

paulhazen-pew (approved):

Raxnam (approved):
looks like a lot but i think i get how you cleaned it all up

Raxnam (approved):

Assets/Plugins/Android/Editor/PlatformSpecificConfigEditorAndroid.cs
arthur740212: Isn’t this Register for the editor windows under Edit->Preference->EOS Unity?
Do they still exist after the changes?

Assets/Plugins/Android/Editor/PlatformSpecificConfigEditorAndroid.cs
arthur740212: Great call, makes total sense to separate these.

Assets/Plugins/Source/Editor/Internal/PackageFileUtils.cs
arthur740212: This is interesting. Does those get used in the restricted instead?

Assets/Plugins/Source/Editor/EpicOnlineServicesConfigEditor.cs
arthur740212: I remember these are for saving configurations, surprises me that they weren’t used at all.

Assets/Plugins/Android/Editor/PlatformSpecificConfigEditorAndroid.cs
paulhazen: That's actually a different thing - that gets done with a static method that gets the attribute `[SettingsProvider]` added to it. See [here](https://docs.unity3d.com/ScriptReference/SettingsProvider.html) for documentation on that. 

We actually do not consistently implement this across the different configuration windows that we have, and since we make no mention of accessing the options via `Edit` -> `Preference` -> `EOS Unity` in the documentation, I removed them until we can implement all of them.

But back to your original question, the register method is something that we engineered. We utilized the `[InitializeOnLoadMethod]` attribute like this. When Unity launches, it scans all scripts for the `[InitializeOnLoadMethod]` attribute and executes the static method that it's applied to. We leveraged this behavior combined with our static approach to make every "Platform Specific" configuration class register itself with the central configuration window `EpicOnlineServicesConfigEditor`. 

To fix issues that this PR addresses, removing the static-ness was a central goal. So now, instead of the platform-specific configuration editors adding themselves to the main config window statically and only when Unity launches, the main config window instantiates each section when the window is actually opened.

Assets/Plugins/Source/Editor/Internal/PackageFileUtils.cs
paulhazen: Nope. Eventually we'll need to circle back to this to resolve. It's not mentioned in the docs though, so until we fully implement it and reference it in the documentation, it's safe for it to quietly die (particularly because we have the history tracked via git).

Assets/Plugins/Source/Editor/EpicOnlineServicesConfigEditor.cs
paulhazen: Yes - there was some scaffolding in place that queried the return value of this function to determine whether or not to prompt the user to save settings - but again since all implementations only ever return `false`, they weren't doing anything. 

In a separate branch I actually have some work done that will re-enable this functionality in a manner that's consistent. In that branch I also implemented it so that the "Save" button on the config window will be disabled unless there are changes to save - in addition to prompting the user to save if they haven't. But it felt appropriate to restrict this PR to the scope of the problem it is intended to address.

arthur740212:
image
There is an error in the Preference settings after the fix.
Which is caused by keywords not being able to register in the Project Settings window
Though I think we are good to merge this PR and fix that later, since these exact same configs could be accessed through the `Tool->EOS Plugin` editor windows as well.
Additionally, the resizing seems to work weird when opening the windows from `Edit->Preference` or `Edit->Project Settings`

paulhazen:
> image There is an error in the Preference settings after the fix. Which is caused by keywords not being able to register in the Project Settings window Though I think we are good to merge this PR and fix that later, since these exact same configs could be accessed through the `Tool->EOS Plugin` editor windows as well. Additionally, the resizing seems to work weird when opening the windows from `Edit->Preference` or `Edit->Project Settings`

Good catch. I've now fixed the issue.

Jason-J-PEW (approved):

Issue #512, opened 2024-01-09
Using EOS Voice as an Audio Source (LocalRTCOptions) on Create/Join Lobbies


dvd900:
**Is your feature request related to a problem? Please describe.**
I need to be able to use the rtc audio buffer in an audio source to utilize unity's audio processing (think spatial audio). This works fine using the [Notify Before Render Callback](https://dev.epicgames.com/docs/en-US/api-ref/functions/eos-rtc-audio-add-notify-audio-before-render). However, without passing in LocalRTCOptions to EOS SDK, The RTC audio gets played directly to my output device, as well as giving me a buffer in the callback mentioned above.

**Describe the solution you'd like**
I want be able to pass in EOS_Lobby_LocalRTCOptions into the create/join lobby calls so that I can set the flag [UseManualAudioOutput](https://dev.epicgames.com/docs/api-ref/structs/eos-lobby-local-rtc-options) to true. This prevents the audio from rendering twice, once from my audio source, once directly to my output device.  

**Describe alternatives you've considered**
The create/join functions have a purposely trim interface but I'd like to use this feature without modifying the library. So I can stay up to date without having to reimplement the feature each release. I've considered writing an extension class, but that doesn't give me access to the private variables I need. 

EDIT: Previously I had assumed the feature was not working as intended because of platform issues, sorry about that.

arthur740212:
https://github.com/PlayEveryWare/eos_plugin_for_unity/pull/513

Hello, here is a PR attempting to allow more control over local RTC options.
We would like to invite you to review this feature, and please let us know if it solves your problem. :D

virtualboys:
It does not fully, as you also need to be able to set LocalRTC options when joining a lobby as well. Created a PR #514 solving this by overloading CreateLobby and JoinLobby. Considered extending the Lobby class but thought this was not the right way as these options are not properties of the lobby itself but rather local options.

arthur740212:
Hi, I changed the PR a bit, and added notes on how to use custom rtc options.
Let us know if it tackles your problem.
This change doesn't change the plugin code to much, and you would not need to manually change code every update.
https://github.com/PlayEveryWare/eos_plugin_for_unity/pull/513

virtualboys:
Yup this looks great!

PR #513, opened 2024-01-10
feat(rtc) : allow custom rtc options

arthur740212:
This PR creates a `localRTCOptions` variable in the `Lobby` class, that should allow the user to set the options to meet their needs.
The thing I am not so sure about is whether this variable should really exist in this class, or even if it needs to be held beyond the scope of a function/method

### DIFF
--- a/Assets/Scripts/EOSLobbyManager.cs
+++ b/Assets/Scripts/EOSLobbyManager.cs
@@ -47,6 +47,7 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
         public uint AvailableSlots = 0;
         public bool AllowInvites = true;
         public bool? DisableHostMigration;
+        public LocalRTCOptions? localRTCOptions;
 
         // Cached copy of the RoomName of the RTC room that our lobby has, if any
         public string RTCRoomName = string.Empty;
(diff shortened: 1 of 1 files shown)


dvd900:
Thanks for the help Arthur!
As mentioned [here](https://github.com/PlayEveryWare/eos_plugin_for_unity/issues/512#issuecomment-1885646582), we also need to be able to set these options on the join, Otherwise these settings will only be accessible to the host and not the other players.

This is my colleagues attempt at the same feature: https://github.com/PlayEveryWare/eos_plugin_for_unity/pull/514, with the necessary extensions to join.

paulhazen (commented):
How do these changes correspond to / overlap with the changes suggested by PR #512?

arthur740212:
> How do these changes correspond to / overlap with the changes suggested by PR #512?

Because a `Lobby` class can be created through the UI scripts, which are modifiable.
By giving `localRTCOptions` a value while setting lobby properties, the `CreateLobby` method will have the custom local RTC options set.
Otherwise it will use the default that we used to set.

Still, I am missing the Join Lobby part.

andrew-hirata-playeveryware (approved):

paulhazen:
> Still, I am missing the Join Lobby part.

Has this PR been updated to include this part?

arthur740212:
> > Still, I am missing the Join Lobby part.
> 
> Has this PR been updated to include this part?

Not yet. I am still trying to figure out how to pass Lobby.localRTCOptions across connection.

arthur740212:
Updated.
I misunderstood the descriptions of the localRTCOptions.
Different users in the same lobby should be able to set RTC options locally on their device.
At least this is my guess what the SDK intended to be.

virtualboys:
Okay yes this solution looks good to me!

paulhazen (approved):
Looks good to me.

andrew-hirata-playeveryware (approved):

PR #514, opened 2024-01-10
Manual rtc audio options

virtualboys:
Add support for echo, manual audio input, and manual audio output. Overloaded CreateLobby and JoinLobby, if there is a better way open to feedback.

### DIFF
--- a/Assets/Scripts/EOSLobbyManager.cs
+++ b/Assets/Scripts/EOSLobbyManager.cs
@@ -945,6 +945,19 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
         /// <param name="CreateLobbyCompleted">Callback when create lobby is completed</param>
         public void CreateLobby(Lobby lobbyProperties, OnLobbyCallback CreateLobbyCompleted)
         {
+            CreateLobby(lobbyProperties, false, false, false, CreateLobbyCompleted);
+        }
+
+        /// <summary>
+        /// Wrapper for calling [EOS_Lobby_CreateLobby](https://dev.epicgames.com/docs/services/en-US/API/Members/Functions/Lobby/EOS_Lobby_CreateLobby/index.html)
+        /// </summary>
+        /// <param name="lobbyProperties"><b>Lobby</b> properties used to create new lobby</param>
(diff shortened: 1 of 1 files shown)


Assets/Scripts/EOSLobbyManager.cs
```diff
@@ -977,13 +990,13 @@ public void CreateLobby(Lobby lobbyProperties, OnLobbyCallback CreateLobbyComple
                 createLobbyOptions.DisableHostMigration = (bool)lobbyProperties.DisableHostMigration;
             }
             // Voice Chat
-            if(lobbyProperties.RTCRoomEnabled)
+            if (lobbyProperties.RTCRoomEnabled)
```
paulhazen, line 993: How _dare_ you add a space here! /s

paulhazen (approved):
In light of the fact that your modification introduces an override, I think this is a good change - particularly since it resolves a specific scenario, and clear documentation is provided.

Special thanks to you for adding verbose documentation - it's our aspiration to have all our code comments look like this.

arthur740212:
I was hesitant to have all local RTC options member variables passed as parameters in CreateLobby or JoinLobby functions because I think they are better separated, and would set a troubling example for some other option structure has a dozen of members.
Also according to the JoinLobby script comments, the localRTCOptions seems to tied to the lobby.

virtualboys:
Yeah I agree that it would get out of hand supporting additional options. If you test with VC though you will see that localRTCoptions needs to be set when joining the lobby in addition to creating the lobby.

arthur740212:
In that case we'll have to figure out if the rtc options could be retrieved from the lobby when joining a lobby

arthur740212:
The JoinLobby function passes lobby.id as a parameter.
We could probably pass the Lobby class instead to access the member localRTCOptions.
However it might not be easy to pass this across remotely.

arthur740212:
https://github.com/PlayEveryWare/eos_plugin_for_unity/pull/513
I made an update in my previous PR.
I misunderstood the how the options should be used (I hope)
With this change, a user can set the customLocalRTCOptions through game code.
And CreateLobby and JoinLobby will use those accordingly.

paulhazen:
I'm closing this because the issue is resolved in PR #513

PR #535, opened 2024-02-17
Modular export tool - Part 1

arthur740212:
The package descriptions are now split into chunks for each platform.
And a package is now defined with which platforms are included, named as a preset.
The tool now allows building multiple presets at once.

*The presets need different package_templates or else they will overwrite each other.
*The json data are morphed into dictionaries, because unity json does not directly support dictionaries. And traversing through all platforms to check which platform should be included is triggering

### DIFF
--- a/Assets/Plugins/Source/Editor/Internal/EOSPackageExportConfiguration.cs
+++ b/Assets/Plugins/Source/Editor/Internal/EOSPackageExportConfiguration.cs
@@ -53,12 +53,13 @@ namespace Playeveryware.Editor
         public string name;
 
         [SerializeField]
+        public string packageIdentifierPath;
+
+        [SerializeField]
         public List<string> platformList;
 
         [SerializeField]
         public List<string> sampleList;
-
-        public bool isGettingExported;
     }
 
     [Serializable]
--- a/Assets/Plugins/Source/Editor/Internal/EOSPackageExportConfiguration.cs
+++ b/Assets/Plugins/Source/Editor/Internal/EOSPackageExportConfiguration.cs
@@ -1,0 +1,70 @@
+using System;
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace Playeveryware.Editor
+{
+    [Serializable]
+    public class PlatformExportInfo
+    {
+        [SerializeField]
+        public string platform;
+
+        [SerializeField]
+        public string descPath;
+
+        [SerializeField]
+        public bool isGettingExported;
+    }
+
+    [Serializable]
+    public class PlatformExportInfoList
+    {
+        [SerializeField]
+        public List<PlatformExportInfo> platformExportInfoList;
+
+    }
+
+    [Serializable]
+    public class SampleExportInfo
+    {
+        [SerializeField]
+        public string sample;
+
+        [SerializeField]
+        public string descPath;
+
+        [SerializeField]
+        public bool isGettingExported;
+    }
+
+    [Serializable]
+    public class SampleExportInfoList
+    {
+        [SerializeField]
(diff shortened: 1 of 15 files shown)


paulhazen:
It seems like there are still unresolved conflicts here.

Assets/Plugins/Source/Editor/Internal/EOSPackageExportConfiguration.cs
```diff
@@ -0,0 +1,71 @@
...
+    {
+        [SerializeField]
+        public string sample;
+
+        [SerializeField]
+        public string descPath;
+
+        [SerializeField]
+        public bool isGettingExported;
+    }
```
paulhazen, line 39: I think these definitions can be made into nested classes in places where they are utilized. This helps us avoid the pattern of having more than one definition per file, and also side-steps the issue of having a filename that doesn't properly represent the contents it contains.

Assets/Plugins/Source/Editor/Internal/UnityPackageCreationTool.cs
```diff
@@ -22,6 +22,8 @@
 
 using UnityEngine;
 using UnityEditor;
+using System.IO;
+using Playeveryware.Editor;
```
paulhazen, line 26: `Playeveryware.Editor` should be fixed to be `PlayEveryWare.EpicOnlineServices.Editor`. Despite the fact that this PR doesn't _introduce_ the creation of this `namespace`, when a namespace usage like this is discovered we should correct it. As it stands, when you fix the conflicts with `development`, it's possible that one of my PRs will have already fixed this.

Assets/Plugins/Source/Editor/Internal/UnityPackageCreationUtility.cs
```diff
@@ -160,7 +196,44 @@ public static string GetPackageVersion()
         }
     }
 
-    
+    public static void CreateUPMTarballV2(string preset)
```
paulhazen, line 199: Can you give some context regarding why this is a "V2"? This signature name only makes sense to me if UPM Tarballs themselves have a "version 2" - in which case there should be one function called `CreateUPMTarball`, and a parameter should be added that indicates the version of tarball to make. 

However, I suspect that your intention was to differentiate between the original `CreateUPMTarball` function and _this_ function - in which case you should have a function name and comments that make the distinction between the two functions clearer. Alternatively, if this function should replace the original function but there is good reason to not remove the original function, you should indicate that the old function is deprecated by adding `_DEPRECATED` or something to the function signature.

paulhazen (changes requested):
## Conflicts

I have a few concerns - first amongst them being the conflicts that exist between the `development` branch and the branch you are merging from.

## Purpose

It's not clear to me what functionality this tool provides, can you give some clear examples of how this tool will make things easier for either us or our users? It's very possible I'm missing an obvious benefit.

## Concerns regarding JSON usage

An additional concern I have is increasing the number of / utilization of JSON files whose contents make assumptions about where specific code files are. 

Reason being (as the code-base changes) these assumptions will become (at best) incomplete, or (at worst) completely wrong. It will require future developers to know that they need to alter the related JSON files if they make changes that affect how things work.

Is there another way we can identify the files that need to be moved in a more generic way? I'm aware that this approach is one that we take in other areas of the project - but those areas are _also_ concerning to me, and I'd hope that we avoid introducing more instances of this pattern. That said, I'd have to think a bit more about what a good alternative would be. Perhaps if I understood the purpose of this tool a bit better, I'd get why things are done this way.

paulhazen:
Closing this PR for the time being considering the extenuating conflicts with the branch, and further work that has been done on this feature.

PR #537, opened 2024-02-19
Config Editor

paulhazen:
Streamlines much of the configuration editing systems. Renames some files to maintain a consistent naming convention that is more succinct. Introduced system to compare sets of config values for changes and default-ness. Moved duplicated code into baser abstract classes where the functionality is shared between similar objects and processes.


Assets/Plugins/Source/Core/PlatformManager.cs
```diff
@@ -0,0 +1,294 @@
...
+            }
+        }
+
+        static PlatformManager()
+        {
+            AddPlatformInfo(Platform.Android, "Android", "eos_android_config.json", typeof(AndroidConfig), ".so");
+            AddPlatformInfo(Platform.iOS, "iOS", "eos_ios_config.json", typeof(IOSConfig), ".dylib");
+            AddPlatformInfo(Platform.Linux, "Linux", "eos_linux_config.json", typeof(LinuxConfig), ".so");
+            AddPlatformInfo(Platform.macOS, "macOS", "eos_macos_config.json", typeof(MacOSConfig), ".dylib");
+            AddPlatformInfo(Platform.Steam, "Steam", "eos_steam_config.json", typeof(EOSSteamConfig), "");
```
paulhazen, line 99: Confirm that `.so` is correct for Android.

paulhazen: Yes - it's correct - but irrelevant, because Android doesn't have native code that needs to be compiled.

Assets/Plugins/Source/Editor/ConfigHandler.cs
```diff
@@ -0,0 +1,139 @@
...
+            // If this is the first time we are saving the config, we need to create the directory
+            // If the directory already exists this will do nothing
+            FileInfo file = new(_configFilePath);
+            file.Directory?.Create();
+
+            File.WriteAllText(_configFilePath, configDataAsJSON);
+            //}
+
+            AssetDatabase.SaveAssets();
+            AssetDatabase.Refresh();
```
paulhazen, line 136: Best to have Default configurations not be written to disk.

andrew-hirata-playeveryware (approved):

arthur740212 (approved):

PR #546, opened 2024-02-26
test: Introduce Rudimentary Unit Testing

paulhazen:
This PR does the following:

- Introduces a utility for safely converting between signed and unsigned variants of numeric datatypes `long` and `int`. (The purpose of this class is to help validate configuration values that are entered by the user as strings, but must be converted into long values - we currently don't validate those converted correctly. The second purpose of this utility is to provide a good proof-of-concept for unit testing).
- Introduces a suite of tests to validate that the `SafeTranslator` class is functioning properly.
- Changes the assembly name `com.playeveryware.eos-Editor` to `com.playeveryware.eos.editor`, so that the name follows the naming convention used in all our other assemblies.


Assets/StreamingAssets/EOS/eos_steam_config.json
arthur740212, line 1: Just wondering if this file change is for another PR.

arthur740212 (approved):

Assets/Plugins/Source/Editor/com.playeveryware.eos.editor.asmdef
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "com.playeveryware.eos-Editor",
+    "name": "com.playeveryware.eos.editor",
```
andrew-hirata-playeveryware: I think this change is a no-go because it's a breaking change. Originally the idea was to keep asmdef files short and be 3-ish components long. I also originally had the idea of keeping the asmdef naming convention close to the UPM package naming convention, which would have it start with `com.playeveryware`, which leaves only with one component for it to change, hence the original use of `-`. If we're going to force a change that could break some asmdef ref files, I think we might need to have a longer discussion on how we name things. 

For example, we could change it so that the assembly name matches the namespace instead of the UPM name, which would help make _some_ of the assembly names shorter. 

Some notes for future me on the various naming conventions used by Unity for reference: 
https://docs.unity3d.com/Manual/cus-naming.html
https://docs.unity3d.com/Manual/cus-asmdef.html

andrew-hirata-playeveryware (changes requested):

Assets/Plugins/Source/Editor/com.playeveryware.eos.editor.asmdef
paulhazen: That's interesting, I didn't realize that it would be a breaking change. What causes the breakage? Things seemed to function properly, so I'm curious where my preliminary testing is lacking.

Assets/Plugins/Source/Editor/com.playeveryware-Editor.asmdef
```diff
@@ -24,4 +24,4 @@
         }
     ],
     "noEngineReferences": false
-}
+}
```
paulhazen: *sigh* it's so annoying when these changes get registered.

andrew-hirata-playeveryware (approved):

PR #547, opened 2024-02-27
build: Introduce BuildRunner, and PlatformSpecificBuilder classes

paulhazen:
Introduces a unified system for executing build tasks on a per-platform basis.

Unity honors all implementations of the interfaces `IPreprocessBuildWithReport`, and `IPostprocessBuildWithReport`. This means that across the project there are disparate tasks being carried out in a sort of mix-mash of circumstances. Some of the tasks happen for all "Standalone" targets, some happen for specific platforms, some happen only if EAC is enabled, etc. 

This PR introduces the concept of a `BuildRunner`, which is to be the only place where Preprocess and Postprocessing occurs. Primarily the build tasks are split amongst the various targeted platforms, but tasks that need to be carried out across all platforms are carefully considered as well.

The `BuildRunner` executes a class that implements the `PlatformSpecificBuilder` abstract class, the derivation of which specifies properties or tasks that are unique to the given platform.

This is a first part in a multi-part effort to consolidate all the build tasks into one pipeline so that we can tell our users a better story for our continuous integration approach / strategy.


andrew-hirata-playeveryware (changes requested):
Some changes might be needed for this PR

paulhazen:
Closing until work has been done to resolve some issues with compatibility.

paulhazen:
Re-opening following discussion and subsequent resolution of restricted platform-compatibility concerns.

Assets/Plugins/Source/Editor/Utility/ScriptingDefineUtility.cs
```diff
@@ -23,64 +23,57 @@
...
+        private static bool DoesScriptingDefineExist(BuildReport report, string defineAsString)
         {
 #if UNITY_2021_2_OR_NEWER
             var namedBuildTarget =
                 UnityEditor.Build.NamedBuildTarget.FromBuildTargetGroup(report.summary.platformGroup);
             var defineSymbolsForGroup = PlayerSettings.GetScriptingDefineSymbols(namedBuildTarget);
 #else
-        // Ensure that we don't do checks for config files if the current platform group has EOS disabled
-        var defineSymbolsForGroup = PlayerSettings.GetScriptingDefineSymbolsForGroup(report.summary.platformGroup);
+            // Ensure that we don't do checks for config files if the current platform group has EOS disabled
```
andrew-hirata-playeveryware, line 45: Are those tabs in the file?

andrew-hirata-playeveryware (approved):

Assets/Plugins/Source/Editor/Utility/ScriptingDefineUtility.cs
paulhazen: Something like that. I think I need to fix my `git.autocrlf` setting or something. When I turn off whitespace difference rendering there are no changes according to GitHub.

arthur740212 (approved):

PR #549, opened 2024-02-27
tool: Auto-copyright updater

paulhazen:
This GitHub action and associated python script will _only_ trigger when a PR is opened into the `development` branch. 

It scans the PR for non-binary altered files that contain copyright statements in the comment header. If the copyright statement is not up-to-date, it will updated it, then commit those changes in the branch being merged from.

>[!NOTE]
> While this workflow (like our "close inactive issues" workflow) runs on GitHub-hosted runners, usage should be negligible.
>
> A side-effect to be cognizant of with this tool is that if you push changes and rely upon it to update the copyright, when you then go to make subsequent changes you will be prompted to first pull the changes that this workflow makes before you can push your new changes.


arthur740212 (approved):

paulhazen:
Turns out this is legally not necessary, so it's a little more work than needs to be done.

PR #558, opened 2024-03-08
fix: resolve errors where types only availabe in the Unity Editor were

andrew-hirata-playeveryware:
# Intro 
While doing a quick audit of the state of the code in the development branch, I noticed that the standalone builds were not building.

# Overview of changes
Single commit with the changes needed to resolve the build errors, allowing standalone builds to work.


arthur740212 (approved):

paulhazen (approved):
Good catches, surprised these bugs slipped through; thanks for catching and fixing them! :)

PR #563, opened 2024-03-14
feat(networking): Change from copying byte arrays to using ArraySegments in EOSTransport SendPacket methods.

andrew-hirata-playeveryware:
# Intro 
While looking at the EOSTransport code, I noticed that we do a lot of converting from ArraySegments<bytes> into byte[] types, and also cloning those byte[] to other arrays.

# Overview of changes
These changes are targeted on the `EOSTransportManager.SendPacket` method, and focus only on decreasing the number of copies that occur between arrays to help speed up packet sending.

### DIFF
--- a/Assets/Scripts/Networking/EOSTransportManager.cs
+++ b/Assets/Scripts/Networking/EOSTransportManager.cs
@@ -983,20 +983,35 @@ namespace PlayEveryWare.EpicOnlineServices.Samples.Network
         /// <param name="reliability">Which level of reliability the packet should be sent with.</param>
         public void SendPacket(ProductUserId remoteUserId, string socketName, byte[] packet, byte channel = 0, bool allowDelayedDelivery = false, PacketReliability reliability = PacketReliability.ReliableOrdered)
         {
+            var packetAsArraySegment = new ArraySegment<byte>(packet);
+            SendPacket(remoteUserId, socketName, packetAsArraySegment, channel, allowDelayedDelivery, reliability);
+        }
+
+        /// <summary>
+        /// Sends a packet to a given remote user on an open connection using the specified reliability.
+        /// </summary>
+        /// <param name="remoteUserId">The id of the remote user to send a packet to.</param>
+        /// <param name="socketName">The name of the socket the connection is open on.</param>
+        /// <param name="packet">The packet to be sent.</param>
+        /// <param name="channel">Which channel the packet should be sent on.</param>
+        /// <param name="allowDelayedDelivery">If <c>false</c> and there is not an existing connection to the peer, the data will be dropped.</param>
+        /// <param name="reliability">Which level of reliability the packet should be sent with.</param>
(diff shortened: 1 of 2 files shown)


Assets/Scripts/Networking/EOSTransportManager.cs
```diff
@@ -982,21 +982,37 @@ private void TryHandleConnectionClosed(ProductUserId remoteUserId, string socket
...
+        /// <summary>
+        /// Sends a packet to a given remote user on an open connection using the specified reliability.
+        /// </summary>
+        /// <param name="remoteUserId">The id of the remote user to send a packet to.</param>
+        /// <param name="socketName">The name of the socket the connection is open on.</param>
+        /// <param name="packet">The packet to be sent.</param>
+        /// <param name="channel">Which channel the packet should be sent on.</param>
+        /// <param name="allowDelayedDelivery">If <c>false</c> and there is not an existing connection to the peer, the data will be dropped.</param>
+        /// <param name="reliability">Which level of reliability the packet should be sent with.</param>
+        // why is this a byte array instead of an array segment??????
```
paulhazen: Seems like this comment should be changed.

Assets/Scripts/Networking/EOSTransportManager.cs
```diff
@@ -1027,29 +1043,26 @@ public void SendPacket(ProductUserId remoteUserId, string socketName, byte[] pac
...
             int currentOffset = 0;
-            List<byte[]> messages = new List<byte[]>();
 
             ushort OutgoingFragmentedIndex = connection.GetNextMessageIndex();
+            byte[] fragmentBuffer = new byte[MaxPacketSize];
             for (ushort i = 0; i < numFragments; ++i)
             {
-                byte[] fragment = new byte[Mathf.Min( (packet.Length - currentOffset) + FragmentHeaderSize, MaxPacketSize)];
+                var fragment = new ArraySegment<byte>(fragmentBuffer, 0, Mathf.Min((packet.Count - currentOffset) + FragmentHeaderSize, MaxPacketSize));
+                //byte[] fragment = new byte[Mathf.Min( (packet.Length - currentOffset) + FragmentHeaderSize, MaxPacketSize)];
```
paulhazen: Unless it's here for historical purposes or to illustrate something, I think this comment should be removed.

paulhazen (changes requested):
Very minor things I think need rectifying. Ping me once they're resolved and I'll resubmit with an approval.

Assets/Scripts/Networking/EOSTransportManager.cs
andrew-hirata-playeveryware: Oh haha! I have removed this note to myself.

Assets/Scripts/Networking/EOSTransportManager.cs
andrew-hirata-playeveryware: Nope! I just missed this line when cleaning up the code. Good catch!

paulhazen (approved):
This was a great catch, and interesting insight. Learn something new everyday 😃

arthur740212 (approved):

PR #564, opened 2024-03-14
fix: Rectified regression concerning the EOSBootstrapper and the EOS overlay.

paulhazen:
This PR fixes a regression that was introduced at some point during the recent changes to the build system.

Those changes errantly coupled the configuration and execution of Easy-Anti-Cheat and the Bootstrapper. 

The regression was such that if EAC was not configured, then the bootstrapper tool would not be utilized, which would prevent the friends overlay from displaying on Windows.

### DIFF
--- a/Assets/Plugins/Source/Editor/Build/BuildRunner.cs
+++ b/Assets/Plugins/Source/Editor/Build/BuildRunner.cs
@@ -126,11 +126,8 @@ namespace PlayEveryWare.EpicOnlineServices.Build
                 return;
             }
 
-            // perform the post-build task for the platform using it's builder.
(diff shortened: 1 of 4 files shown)


arthur740212 (approved):

arthur740212:
Side question, was EAC implemented for mac or linux yet?

andrew-hirata-playeveryware (changes requested):
Needs to be merged into the release branch instead of the development branch.

paulhazen:
> Side question, was WAC implemented for mac or linux yet?

EAC should be implemented for all three - but we don't have any automated tests around it so I can't be certain.

andrew-hirata-playeveryware (approved):

PR #568, opened 2024-03-19
fix(eac,build) : Remove early return that skips copying files fo EAC

arthur740212:
This line bypasses the whole file copying step for EAC integration.
It is removed to allow normal functionality for Windows

### DIFF
--- a/Assets/Plugins/Source/Editor/Build/EACPostBuild.cs
+++ b/Assets/Plugins/Source/Editor/Build/EACPostBuild.cs
@@ -316,9 +316,6 @@ namespace PlayEveryWare.EpicOnlineServices.Build
             string destDir = Path.GetDirectoryName(report.summary.outputPath);
             string pathToInstallFrom = GetPathToPlatformSpecificAssets(report);
 
-            if (!string.IsNullOrEmpty(pathToInstallFrom))
-                return;
-
             List<string> filestoInstall = GetPostBuildFiles(report);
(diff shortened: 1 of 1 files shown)


paulhazen (commented):
This is actually changed in the release branch - there were much larger problems with the EAC stuff, and this got fixed along the way in the `release-3.1.0` branch.

arthur740212:
I think this still might be a problem, because it skips all the file copying if it finds the source folder. Which doesn't really make sense.

Assets/Plugins/Source/Editor/Build/EACPostBuild.cs
```diff
@@ -316,9 +316,6 @@ private static void InstallEACFiles(BuildReport report)
             string destDir = Path.GetDirectoryName(report.summary.outputPath);
             string pathToInstallFrom = GetPathToPlatformSpecificAssets(report);
 
-            if (!string.IsNullOrEmpty(pathToInstallFrom))
```
andrew-hirata-playeveryware, line 319: Isn't the code as it currently is written correct? If there is no path to install the files from, how would it work?

arthur740212: If it is not empty, it returns.
I see, the real problem is that there is an extra `!` .

arthur740212:
Issue solved in another branch. Closing the PR.

PR #570, opened 2024-03-20
fix: Config File Relocation & Static Access

paulhazen:
Simplified access to config files, and moved them out of streaming assets directory. Introduces some quality of life improvements.


Assets/Plugins/Source/Editor/Utility/UnityPackageCreationUtility.cs
```diff
@@ -66,7 +68,7 @@ public static class UnityPackageCreationUtility
         /// <summary>
         /// Contains section of package.json file pertaining to configuration
         /// </summary>
-        public static PackagingConfigEditor packageConfig;
+        //public static PackagingConfigEditor packageConfig;
```
paulhazen: Why is this commented out?

Assets/Plugins/Source/Editor/Windows/CheckDeploymentWindow.cs
```diff
@@ -28,6 +28,8 @@
 
 namespace PlayEveryWare.EpicOnlineServices.Editor.Windows
 {
+    using System.Threading.Tasks;
+
```
paulhazen: Unneeded `using` statement here.

Assets/Plugins/Source/Editor/Windows/EOSEditorWindow.cs
```diff
@@ -20,13 +20,15 @@
  * SOFTWARE.
  */
 
-    using System;
+using PlasticGui.WorkspaceWindow.BranchExplorer;
+using System;
```
paulhazen: What in the world is this, come on dude :D

Assets/Plugins/Source/Editor/Windows/PluginVersionWindow.cs
```diff
@@ -27,6 +27,8 @@
 
 namespace PlayEveryWare.EpicOnlineServices.Editor.Windows
 {
+    using System.Threading.Tasks;
```
paulhazen: Extraneous using statement.

Assets/Plugins/Source/Editor/Windows/PluginVersionWindow.cs
```diff
@@ -102,15 +104,14 @@ protected override void Setup()
...
-            GUILayout.EndHorizontal();
-
+                GUILayout.BeginHorizontal();
+                GUILayout.Label("Epic Online Services Version:", EditorStyles.boldLabel);
+                GUILayout.Label(eos_library_version);
+                GUILayout.EndHorizontal();
+                GUILayout.BeginHorizontal();
+                GUILayout.Label("Epic Online Services Plugin For Unity:", EditorStyles.boldLabel);
+                GUILayout.Label(eos_plugin_version);
+                GUILayout.EndHorizontal();
```
paulhazen: Fix indent.

Assets/Plugins/Source/Core/Config.cs
```diff
@@ -37,6 +41,73 @@ namespace PlayEveryWare.EpicOnlineServices
...
     public abstract class Config : ICloneable
     {
+        protected readonly string Filename;
+        protected readonly string Directory;
+
+        private string _lastReadJsonString;
+
+        protected Config(string filename) : this(filename, Path.Combine(Application.streamingAssetsPath, "EOS")) { }
+
+        protected Config(string filename, string directory)
```
paulhazen, line 53: Make sure that deriving classes do not have publicly accessible constructors, so that they are required to make use of the templated async Get function.

Assets/Plugins/Source/Core/Config.cs
```diff
@@ -37,6 +41,73 @@ namespace PlayEveryWare.EpicOnlineServices
...
+            }
+        }
+
+        protected virtual async Task ReadAsync()
+        {
+            bool configFileExists = File.Exists(FilePath);
+
+            if (configFileExists)
+            {
+                using (StreamReader reader = new(FilePath))
```
paulhazen: Try to make use of the declarative form of using instead of the one that causes an indent.

paulhazen: And check to make sure that this is something that is supported in the versions of C# we support.

Assets/Plugins/Source/Editor/Build/BuildRunner.cs
```diff
@@ -125,5 +130,55 @@ public void OnPostprocessBuild(BuildReport report)
...
+
+        /// <summary>
+        /// Completes all configuration tasks.
+        /// </summary>
+        private static void ConfigureVersions()
+        {
+            // TODO: DANGEROUS WAITING. There Isn't a reason I can think of that AutoSetProductVersion
+            //       task would fail, but that's just the kind of thing that could end up in an inf 
+            //       loop
+            AutoSetProductVersion().Wait(10000);
```
paulhazen: Fix this so that either all the calls to `Wait` have a default timeout, or this timeout is removed. "Waiting" is only really dangerous if you _don't_ have a timeout. As always, need to guarantee that exceptions are handled with grace.

Assets/Plugins/Source/Editor/Configs/EditorConfig.cs
```diff
@@ -0,0 +1,55 @@
...
+            EditorPrefs.SetString(Filename, configAsJSON);
+            return Task.CompletedTask;
+        }
+
+        protected override Task ReadAsync()
+        {
+            // TODO: Add JSON schemas for validation.
+            string configAsJSON = EditorPrefs.GetString(Filename);
+            JsonUtility.FromJsonOverwrite(configAsJSON, this);
+            return Task.CompletedTask;
```
paulhazen: Take a look at "Shared preferences" described here: https://blog.redbluegames.com/how-to-handle-data-for-custom-editor-tools-in-unity-6b85e9e17715

Users will want to share their config values and have them stored in source control.

Assets/Plugins/Source/Editor/Windows/EOSEditorWindow.cs
```diff
@@ -20,13 +20,15 @@
...
     using System.CodeDom;
```
paulhazen: Extraneous using?

Assets/Plugins/Source/Editor/Windows/EOSEditorWindow.cs
```diff
@@ -121,35 +152,47 @@ protected void SetAutoResize(bool autoResize)
...
+        Debug.Log("Window has been initialized.");
+    }
+
     /// <summary>
     /// Implement this method to initialize the contents of the editor window. Be sure to always call the base implementation if you override it.
     /// </summary>
     protected virtual void Setup()
     {
-        foreach (var fieldKeyPair in RetainableFields)
+        foreach (var (field, key) in RetainableFields)
```
paulhazen: Another thing to check to see if it's supported.

Assets/Plugins/Source/Editor/Windows/EOSSettingsWindow.cs
```diff
@@ -399,13 +400,15 @@ protected override void RenderWindow()
             GUIEditorUtility.AssigningBoolField("Save JSON in 'Pretty' Format", ref prettyPrint, 190);
             if (GUILayout.Button("Save All Changes"))
             {
-                SaveToJSONConfig(prettyPrint);
+                // Save NOTE: if this times out, then there will be problems.
+                Save(prettyPrint).Wait();
```
paulhazen: Again, make sure that exceptions are handled appropriately, and that there are reasonable.

Assets/Plugins/Source/Editor/Windows/EOSEditorWindow.cs
arthur740212: What does this Retainable Field do? I couldn't where the attribute was used

Assets/Plugins/Source/Editor/Utility/UnityPackageCreationUtility.cs
paulhazen: Fixed. The particular class member is no longer in use at all.

Assets/Plugins/Source/Editor/Windows/EOSEditorWindow.cs
paulhazen: If you add a `RetainableField` attribute to any field member on a class derived from `EOSEditorWindow`, it's value will be saved and loaded automatically when the window is closed and subsequently opened. 

It's currently not utilized for anything, but the use case that caused it's implementation in another project is that it allowed user preferences to persist for things like whether or not they had opened an expandable area or not. Basically: preferences related to how the window itself operates - nothing that's related directly to the plugin or the samples.

Assets/Plugins/Source/Core/Config.cs
paulhazen: This pattern is actually not possible to implement without using reflection.

Assets/Plugins/Source/Core/Config.cs
paulhazen: I'm going to close this - determining version support for C# is outside the scope of this PR and should be resolved using an automated testing flow.

Assets/Plugins/Source/Editor/Configs/EditorConfig.cs
paulhazen: Functionality was changed to write configuration values of an "editor" nature to `etc/config/`

andrew-hirata-playeveryware (approved):

Assets/Plugins/Source/Editor/Utility/EACUtility.cs
arthur740212, line 1: The early return on line 273,274 will cause EAC files to never get installed because the source path has to be not null.

arthur740212 (approved):
The line I commented on I have a fix in another PR.
We can add the fix here or have them merged separately
Other than that no problems :D

PR #571, opened 2024-03-20
fix(config) : Load platform config from streaming asset path

arthur740212:
The mac player streaming asset file is different than dataPath/StreamingAssets. It is easier to access directly with streaming assets path


paulhazen (commented):
~This is actually also changed in `release`.~

My mistake! I was looking at the wrong thing. Good catch!

paulhazen (approved):

andrew-hirata-playeveryware (approved):

PR #578, opened 2024-03-25
fix callback info comparisons in OnRTCRoomUpdateReceivingCompleted

DevMassive:


Assets/Scripts/EOSLobbyManager.cs
```diff
@@ -816,7 +816,7 @@ private void OnRTCRoomParticipantStatusChanged(ref ParticipantStatusChangedCallb
                 metadataCount);
 
             // Ensure this update is for our room
-            if (string.IsNullOrEmpty(CurrentLobby.RTCRoomName) || CurrentLobby.RTCRoomName.Equals(data.RoomName, StringComparison.OrdinalIgnoreCase))
+            if (string.IsNullOrEmpty(CurrentLobby.RTCRoomName) || !CurrentLobby.RTCRoomName.Equals(data.RoomName, StringComparison.OrdinalIgnoreCase))
```
paulhazen, line 819: This is a good catch! We will fold this change into an upcoming release.

Assets/Scripts/EOSLobbyManager.cs
```diff
@@ -1644,7 +1644,7 @@ private void OnRTCRoomUpdateReceivingCompleted(ref UpdateReceivingCallbackInfo d
 
             foreach(LobbyMember lobbyMember in CurrentLobby.Members)
             { 
-                if(lobbyMember.ProductId == data.LocalUserId)
+                if(lobbyMember.ProductId != data.ParticipantId)
```
paulhazen, line 1647: For this change in particular, someone with more experience in the lobby area of the project will need to weigh in - I don't feel like I can approve or request changes on this one. But we will continue to look into it, and keep yah posted :)

DevMassive: Ok.

https://github.com/PlayEveryWare/eos_plugin_for_unity/blob/a31af6e2b7b2b736d3db8e5fe37dbdce38487ffd/Assets/Plugins/Source/EOS_SDK/Generated/RTCAudio/UpdateReceivingCallbackInfo.cs#L24-L27

https://github.com/PlayEveryWare/eos_plugin_for_unity/blob/a31af6e2b7b2b736d3db8e5fe37dbdce38487ffd/Assets/Plugins/Source/EOS_SDK/Generated/RTCAudio/UpdateReceivingCallbackInfo.cs#L34-L37

LocalUserId is the ID of oneself, and ParticipantId is the ID of the user who is the target of the change.

andrew-hirata-playeveryware: Oh I see what the code was _trying_ to do before (I think).
Previously it was trying to not double set information that comes from the user setting their own settings. Which isn't _wrong_ but the way it does it is incorrect. Additionally that old code only "works" if there are 2 members. If there are greater than two members, and the second member in the list isn't the one being looked for, the current code will just set that user assuming it's the "correct" member to set. This fix appears to actually correctly identify the user that needs status update.

andrew-hirata-playeveryware (approved):

paulhazen (approved):

arthur740212 (approved):

PR #584, opened 2024-04-03
Mac Editor Fixes

arthur740212:
Fixes compile error on Mac Editor
Also removed the extra spaces in EOSManager.cs


paulhazen (approved):

Assets/Plugins/Source/Core/EOSManager.cs
```diff
@@ -521,6 +518,21 @@ private PlatformInterface CreatePlatformInterface(EOSConfig configData)
...
+#if !(UNITY_STANDALONE_WIN || UNITY_EDITOR_WIN || UNITY_STANDALONE_LINUX || UNITY_EDITOR_LINUX || UNITY_STANDALONE_OSX || UNITY_EDITOR_OSX)
+                integratedPlatformOptionsContainer.Release();
+#endif
+                if (platformInterface != null)
+                {
+                    return platformInterface;
+                }
+                else
+                {
+                    print($"Error creating platform interface:");
```
andrew-hirata-playeveryware: I don't believe this needs to be here, because we throw an exception if the platform interface is NULL int `Init`.

andrew-hirata-playeveryware (changes requested):

Assets/Plugins/Source/Core/EOSManager.cs
arthur740212: That is true.  Updated.

andrew-hirata-playeveryware (approved):

PR #586, opened 2024-04-03
Various Changes to the Create Package Process

paulhazen:
This PR contains some changes to the create package process. 

A lot of the changes are quality of life issues - but some of them are critical - for example, caching the existing binary files in case the building of them doesn't work.

A myriad of (more recent) commits are fixing one-off bugs with the release.


Assets/Plugins/Android/Editor/AndroidBuilder.cs
```diff
@@ -265,7 +266,7 @@ private static string GetBuildTools()
 
         private static void WriteConfigMacros(string filepath)
         {
-            var contents = File.ReadAllText(filepath);
+            var contents = FileUtility.ReadAllText(filepath);
```
paulhazen, line 281: Just changes where the file reading is happening to a central place, to be able to do platform-dependent things with the implementation.

Assets/Plugins/Source/Core/EOSManager.cs
```diff
@@ -555,34 +555,13 @@ public void Init(IEOSCoroutineOwner coroutineOwner)
...
-                    throw new Exception("Couldn't find EOS Config file: Please ensure " + eosFinalConfigPath +
-                                        " exists and is a valid config");
-                }
-
-                configDataAsString = File.ReadAllText(eosFinalConfigPath);
-#endif
-                var configData = JsonUtility.FromJson<EOSConfig>(configDataAsString);
-
-                print("Loaded config file: " + configDataAsString);
-                return configData;
```
paulhazen, line 576: This logic is moved to the `FileUtility` class's `ReadAllText` method, which deals with the change in implementation that is unique to Android.

Assets/Plugins/Source/Core/EOSManager.cs
```diff
@@ -555,34 +555,13 @@ public void Init(IEOSCoroutineOwner coroutineOwner)
...
-            }
-
             //-------------------------------------------------------------------------
             public void Init(IEOSCoroutineOwner coroutineOwner, string configFileName)
             {
                 string eosFinalConfigPath = Path.Combine(Application.streamingAssetsPath, "EOS", configFileName);
                 if (loadedEOSConfig == null)
                 {
-                    loadedEOSConfig = LoadEOSConfigFileFromPath(eosFinalConfigPath);
+                    loadedEOSConfig = JsonUtility.FromJsonFile<EOSConfig>(eosFinalConfigPath);
```
paulhazen, line 567: 90% of the time we are doing a `FromJson` function, we are doing it immediately after reading a file. This helper method enables us to consolidate that behavior to one place, unifying the way we deal with things like IO and thereby reducing points of failure.

Assets/Plugins/Source/Core/Utility/JsonUtility.cs
```diff
@@ -22,21 +22,106 @@
...
+#if UNITY_EDITOR
+                throw;
+#endif
+            }
+            catch (Exception ex)
+            {
+                Debug.LogError($"An error occurred: {ex.Message}");
+#if UNITY_EDITOR
+                throw;
+#endif
```
paulhazen, line 54: If the JSON is invalid, we want to throw an exception if we are in the editor. However, if we are running outside the editor, then we want to fail more silently.

Assets/Plugins/Source/Core/Utility/JsonUtility.cs
```diff
@@ -22,21 +22,106 @@
...
+        /// Determines if the given json string is valid json or not. If JSON is invalid, errors will be logged, but exceptions will be suppressed.
+        /// </summary>
+        /// <param name="json">The JSON string to validate.</param>
+        /// <returns>True if json is valid, false otherwise.</returns>
+        private static bool IsJsonValid(string json)
+        {
+            try
+            {
+                ValidateJson(json);
+                return true;
```
paulhazen, line 68: Note that as a consequence of the decision to not throw exceptions during non-editor runtime, the `IsJsonValid` function will always return `true` when not in Editor. This may or may not be desirable behavior.

Assets/Plugins/Android/Core/AndroidFileIOHelper.cs
paulhazen, line 1: This file existed to provide functionality for Android that was analogous to C#s `File.ReadAllText`. Instead of having it exist in its own file, I created a proxy function in `FileUtility` called `ReadAllText` that has `#if` conditionals around its implementation to provide the same thing.

Assets/Plugins/Source/Core/Utility/FileUtility.cs
```diff
@@ -32,27 +32,58 @@ namespace PlayEveryWare.EpicOnlineServices.Utility
     /// </summary>
     public static class FileUtility
     {
+        // Note: "\r\n" (Windows), "\n" (Unix/Linux), "\r" (older Macs)
+        private static readonly string[] LineEndings = new[] { "\r\n", "\n", "\r" };
```
paulhazen, line 36: This exists to emulate the same functionality provided by C#'s `File.ReadAllLines` functions. Since Android (and perhaps later other platforms) may need special implementations surrounding IO, the function is built off of the existing `ReadAllText` function within `FileUtility`. The way that C# implements `File.ReadAllLines` is that it breaks lines on any and all of these characters (even if the file contents have mixed line endings).

Assets/Plugins/Source/Core/Utility/FileUtility.cs
```diff
@@ -32,27 +32,58 @@ namespace PlayEveryWare.EpicOnlineServices.Utility
...
-        public static string GenerateTempDirectory()
+        public static bool TryGetTempDirectory(out string path)
         {
             // Generate a temporary directory path.
-            string tempDirectory = Path.Combine(Application.temporaryCachePath, $"/Output-{Guid.NewGuid()}/");
+            string tempPath = Path.Combine(Application.temporaryCachePath, $"/Output-{Guid.NewGuid()}/");
 
             // If (by some crazy miracle) the directory path already exists, keep generating until there is a new one.
-            while (Directory.Exists(tempDirectory))
+            if (Directory.Exists(tempPath))
```
paulhazen, line 49: Instead of opening up the possibility of an infinite loop, I limited the "retry" count to just 1. If, after that there is still a conflict, then an error is logged and the function returns false.

Assets/Plugins/Source/Core/Utility/FileUtility.cs
```diff
@@ -71,7 +102,46 @@ public static string GetProjectPath()
...
+#else
+            if (request.isNetworkError || request.isHttpError)
+            {
+                Debug.Log("Requesting " + filePath + ", please make sure it exists and is a valid config");
+                throw new Exception("UnityWebRequest didn't succeed : Network or HTTP Error");
+            }
+#endif
+            text = request.downloadHandler.text;
+#else
+            text = File.ReadAllText(path);
```
paulhazen: This is the implementation that is unique to Android regarding `File.ReadAllText`.

Assets/Plugins/Source/Editor/Controls/CancellableAsyncProgressBar.cs.meta
paulhazen, line 1: While GitHub is treating this as a file rename, the contents are in-fact different, and the project has been checked for broken references.

Assets/Plugins/Source/Editor/Controls.meta
paulhazen, line 1: While GitHub is treating this as a file rename, the contents are in-fact different, and the project has been checked for broken references.

Assets/Plugins/Source/Editor/Utility/UnityPackageCreationUtility.cs
```diff
@@ -75,15 +75,6 @@ public enum PackageType
...
         public static CoroutineExecutor executorInstance;
-        
-        private static PackageDescription ReadPackageDescription(string pathToJSONPackageDescription)
-        {
-            var JSONPackageDescription = File.ReadAllText(pathToJSONPackageDescription);
-
-            var packageDescription = JsonUtility.FromJson<PackageDescription>(JSONPackageDescription);
-
-            return packageDescription;
-        }
```
paulhazen, line 86: Calls to this function were replaced by a simple call to 
```csharp
JsonUtility.FromJsonFile<PackageDescription>(pathToJSONPackageDesription);
```

Assets/Plugins/Source/Editor/Utility/PackageFileUtility.cs
```diff
@@ -24,187 +24,82 @@
...
 
-                // truncate trailing garbage.
-                filestream.SetLength(filestream.Position);
-            }
-        }
-        
-        public static void Dos2UnixLineEndings(string filename)
-        {
-            Dos2UnixLineEndings(filename, filename);
-        }
```
paulhazen, line 77: These functions (line ending conversion functions) were moved into an extensions class and greatly simplified.

etc/PackageConfigurations/eos_package_description.json
```diff
@@ -49,13 +49,33 @@
 
         {"src":"Assets/Plugins/Source/Editor/*.cs", "dest": "Editor/"},
         {"src":"Assets/Plugins/Source/Editor/*.cs.meta", "dest": "Editor/"},
+        {"src":"Assets/Plugins/Source/Editor/Controls.meta", "dest":"Editor/"},
+        {"src":"Assets/Plugins/Source/Editor/Controls/*.cs", "dest":"Editor/Controls/"},
+        {"src":"Assets/Plugins/Source/Editor/Controls/*.cs.meta", "dest":"Editor/Controls/"},
         
         {"src":"Assets/Plugins/Source/Editor/Utility.meta", "dest": "Editor/"},
         {"src":"Assets/Plugins/Source/Editor/Utility/com.playeveryware.eos-Editor.steam.utility.asmdef", "dest": "Editor/Utility/"},
```
arthur740212: The steam version button will depend on this to work. 
I can make a separate PR to explain.
For the purpose of this PR, it shouldn't matter.

arthur740212 (approved):
Steam Editor Utility missing (low prio)
Package builder cleans whole output path (dangerous for us, doesn't affect upm users)

Other than those, the apps are building and running great.

arthur740212 (changes requested):
On a project with no `config.json` s
The config editor loads one platform every time opening the window.
Because it creates a json for one and only one platform each opening.

pbtest - SampleScene - Windows, Mac, Linux - Unity 2021 3 8f1 _DX11_ 4_4_2024 5_38_57 PM

paulhazen:
> On a project with no `config.json` s The config editor loads one platform every time opening the window. Because it creates a json for one and only one platform each opening.

To clarify - this only happens when the UPM is imported into a fresh project, correct?

paulhazen:
> On a project with no `config.json` s The config editor loads one platform every time opening the window. Because it creates a json for one and only one platform each opening.

I've tried multiple times now to reproduce this issue, but don't seem to be able to. Can please you double check and provide detailed reproduction steps including platfom?

arthur740212:
Based on how the bug works, I'll expect the repo to also have problems, but a project with upm imported does have this issue.

So if we have a empty project with the plugin upm imported.
And the `StreamingAssets` does not contain any `eos_config_[platform].json`
Open the config editor and you'll find that it shows nothing, and logs errors in the log window.
At this moment the main config file will be created in `StreamingAssets` folder
Close the window and reopen again you'll see the Main config label is created but all other labels are empty.
You'll also notice that the steam coonfig .json is also created in `StreamingAssets` folder

paulhazen:
Closing this PR in favor of a narrower approach to changes.

PR #588, opened 2024-04-04
EOSManager related fixes.

arthur740212:
Same as #584 but instead the branch is made from `release-3.1.0 instead` of `development`.
Where there is a change in the `CreatePlatformInterface` function

Includes :
Mac compile fixes
Remove extra whitespaces
Release IntegratedPlatformContainer before first tick to prevent garbage collector crash. (mobile, console)


Assets/Plugins/Source/Core/IPlatformSpecifics.cs
```diff
@@ -79,7 +79,7 @@ public interface IEOSCreateOptions
         Utf8String CacheDirectory { get; set; }
         uint TickBudgetInMilliseconds { get; set; }
 
-#if !(UNITY_STANDALONE_WIN || UNITY_EDITOR_WIN || UNITY_STANDALONE_LINUX|| UNITY_EDITOR_LINUX)
+#if !(UNITY_STANDALONE_WIN || UNITY_EDITOR_WIN || UNITY_STANDALONE_LINUX || UNITY_EDITOR_LINUX || UNITY_STANDALONE_OSX || UNITY_EDITOR_OSX)
         Epic.OnlineServices.IntegratedPlatform.IntegratedPlatformOptionsContainer IntegratedPlatformOptionsContainerHandle { get; set; }
 #endif
```
paulhazen, line 84: I know this might seem super pedantic - but couldn't / shouldn't we change this to the more easily read:

```C#
#if !(UNITY_STANDALONE || UNITY_EDITOR)
```

?

Same question applies to the other file that's been changed in this PR.

paulhazen (changes requested):

Assets/Plugins/Source/Core/IPlatformSpecifics.cs
arthur740212: If it doesn't affect mobile platforms then should be squashable.

andrew-hirata-playeveryware (approved):

paulhazen (approved):

PR #590, opened 2024-04-05
Narrower fixes to release branch

paulhazen:
This PR is a more targeted version of PR #586. 

It contains _only_ the changes that are strictly needed to stabilize the release.

This gives us the option of being more incremental with changes if need be.

If we deem that doing so is appropriate: other improvements can be introduced in smaller pull requests later.


Assets/Plugins/Source/Core/EOSManager.cs
```diff
@@ -471,7 +471,7 @@ private PlatformInterface CreatePlatformInterface(EOSConfig configData)
                 IPlatformSpecifics platformSpecifics = EOSManagerPlatformSpecificsSingleton.Instance;
 
                 IEOSCreateOptions platformOptions = new EOSCreateOptions();
-                platformOptions.CacheDirectory = platformSpecifics.GetTempDir();
+                platformOptions.CacheDirectory = Application.temporaryCachePath;
```
andrew-hirata-playeveryware: Why was this changed to temporaryCachePath?

Assets/Plugins/Source/Core/Utility/FileUtility.cs
```diff
@@ -64,6 +64,73 @@ public static string GetProjectPath()
...
+        {
+            // Generate a temporary directory path.
+            string tempPath = Path.Combine(Application.temporaryCachePath, $"Output-{Guid.NewGuid()}/");
+
+            // If (by some crazy miracle) the directory path already exists, keep generating until there is a new one.
+            if (Directory.Exists(tempPath))
+            {
+                Debug.LogWarning(
+                    $"The temporary directory created collided with an existing temporary directory of the same name. This is very unlikely.");
+                tempPath = Path.Combine(Application.temporaryCachePath, $"Output-{Guid.NewGuid()}/");
```
andrew-hirata-playeveryware: `Application.temporaryCachePath` Doesn't always work (raises an exception), on some platforms. On a few platforms it returns an empty string.

Assets/Plugins/Source/Core/Utility/FileUtility.cs
```diff
@@ -136,5 +203,29 @@ public static void CleanDirectory(string directoryPath, bool ignoreGit = true)
                 Debug.Log($"An error (which was ignored) occurred while cleaning \"{directoryPath}\": {ex.Message}");
             }
         }
+
+#if UNITY_EDITOR
+        public static void OpenFolder(string path)
```
andrew-hirata-playeveryware: Small naming convention issue: I think this should be 'OpenDirectory' to match the rest of the usage.

Assets/Plugins/Source/Core/Utility/FileUtility.cs
```diff
@@ -64,6 +64,73 @@ public static string GetProjectPath()
...
+
+        public static void ConvertDosToUnixLineEndings(string filename)
+        {
+            ConvertDosToUnixLineEndings(filename, filename);
+        }
+
+        public static void ConvertDosToUnixLineEndings(string srcFilename, string destFilename)
+        {
+            string fileContents = ReadAllText(srcFilename);
+            fileContents = fileContents.Replace("\r\n", "\n");
```
andrew-hirata-playeveryware: This new method will end up with two copies of the string in memory, which is not as kind to the GC as the previous code.

Assets/Plugins/Source/Editor/Utility/UnityPackageCreationUtility.cs
```diff
@@ -116,29 +116,6 @@ public static async Task CreatePackage(PackageType packageType, IProgress<Create
             ValidatePackage(packagingConfig.pathToOutput);
         }
 
-        private static void CreateDotUnityPackage(string outputPath, string json_file,
```
andrew-hirata-playeveryware, line 119: Why was this deleted?

andrew-hirata-playeveryware (changes requested):

Assets/Plugins/Source/Core/EOSManager.cs
paulhazen: If memory serves, it's because that's what the `GetTempDir` function returned, and this was the only place it was being used.

Assets/Plugins/Source/Editor/Utility/UnityPackageCreationUtility.cs
paulhazen: For a few reasons. The json file that we have for it is wildly outdated, and we don't make the option to create a `.unitypackage` available anymore in the UI. The short answer is that it was removed because it hasn't worked for a while now.

Assets/Plugins/Source/Core/Utility/FileUtility.cs
paulhazen: Fixed this.

Assets/Plugins/Source/Core/EOSManager.cs
paulhazen: I'll restore the previous functionality so certain platforms have the option of returning a different temporary directory.

Assets/Plugins/Source/Core/Utility/FileUtility.cs
paulhazen: Interesting. In this context, it is utilized for the build process, so it should never be used by any platform that is not capable of _building_ the plugin.

Assets/Plugins/Source/Core/Utility/FileUtility.cs
paulhazen: Ah that makes sense. Would you like me to restore previous functionality? This is only used while in the editor, and while copying files during the create package process, seems like a good trade-off for increased readability, but it's far from a hill that I'd die on 😄

andrew-hirata-playeveryware: I would prefer to restore the previous functionality.

Assets/Plugins/Source/Core/Utility/FileUtility.cs
andrew-hirata-playeveryware: That is true, but I believe this code is now in a share place where it could be used by someone on accident and break one of those platforms now.

andrew-hirata-playeveryware (changes requested):

Assets/Plugins/Source/Core/Utility/FileUtility.cs
paulhazen: Previous functionality has been restored.

Assets/Plugins/Source/Core/Utility/FileUtility.cs
paulhazen: I now have it (and some other functions only used in editor) enclosed in `#if UNITY_EDITOR` conditionals.

andrew-hirata-playeveryware (approved):

Assets/Plugins/Source/Core/Utility/FileUtility.cs
paulhazen: I actually moved this function into the `PackageFileUtility` to be extra certain it doesn't get utilized inappropriately.

arthur740212 (approved):

PR #595, opened 2024-04-10
Script to help identify missing scripts

paulhazen:
Helper script that I used to correct some issues with missing scripts in scenes and prefabs. I made it conditionally active based on a scripting define that is currently commented out - so this code file does nothing unless you uncomment the `#define PROJECT_UTILITY_ENABLED` line at the top.

Adds menu items:

`Tools / Find Missing Scripts / In Scenes`
`Tools / Find Missing Scripts / In Prefabs`

Self-explanatory, but identifies any missing scripts.

`Tools / Find Unreferenced Prefabs`

Finds any prefabs that are not referenced by a scene directly, or by another prefab. Basically, if there are prefab not referenced by a scene that are themselves referencing prefabs referencing prefabs etc. then all the descendant prefabs will be logged as being unreferenced.

This does not need to be in release, but is opened against it, nonetheless.


arthur740212 (approved):

paulhazen:
Closing PR in favor of opening it into the development branch at a later time.

PR #599, opened 2024-04-12
Unity Project Utility

paulhazen:
This PR is the result of two changes:

- Implementation of a script that makes accessible via the `Tools` menu in the Unity Editor various operations that allow for checking of some of the most common errors and warnings that get introduced to the project.
- Using aforementioned script to identify `*.prefab` files that were not referenced by any of the sample scenes, and deleting them along with their corresponding `.meta` files.

By default, the script is conditionally excluded unless you define `ENABLE_PROJECT_UTILITY`, which is near the top of the `UnityProjectUtility.cs` file - consequently there should not be any noticeable changes to the project.


Assets/EOS/EOSManager.prefab
andrew-hirata-playeveryware, line 1: Isn't this EOSManager.prefab part of the steps we recommend for adding integrating the Plugin?

Assets/Prefab/P2P_Particle.prefab
andrew-hirata-playeveryware, line 1: The fact that these particle prefabs are not being used is more of a sign that the refactor/change to the chat system to allow it to send non-text data (particles, in this case) might be broken. I.e. in a separate PR, we either remove that functionality, along with these prefabs, or we fix it.

andrew-hirata-playeveryware (changes requested):

Assets/Prefab/P2P_Particle.prefab
paulhazen: I think this will be a good tool to catch things like this. I'll revert all the removed prefabs.

Assets/EOS/EOSManager.prefab
paulhazen: It is / was, but there are two of them (albeit in different directories) but we only ever direct users to use one of them. Regardless, I think it makes more sense to have the removal of possibly extraneous prefabs be a component of another PR, and have this one stick to introducing the script.

andrew-hirata-playeveryware (approved):

arthur740212 (approved):

PR #600, opened 2024-04-12
Unify implementation of ReadAllText

paulhazen:
This PR removes `AndroidFileIOHelper.cs` and the various places it is utilized, in favor of including the same logic conditionally compiled into the implementation of the `FileUtility.ReadAllText` method.


andrew-hirata-playeveryware:
This merge is being made into the wrong branch :/

andrew-hirata-playeveryware:
I like the idea of the FileUtility being the main point of entry for doing `ReadAllText`, but I think it's a better pattern for this project to keep the android code in it's own specific file, similar to how we segment other code by platform by putting it in it's own file, when it's possible.

andrew-hirata-playeveryware (changes requested):
I like the idea of the FileUtility being the main point of entry for doing ReadAllText, but I think it's a better pattern for this project to keep the android code in it's own specific file, similar to how we segment other code by platform by putting it in it's own file, when it's possible. Besides that I'm fine with the change.

andrew-hirata-playeveryware (approved):

arthur740212 (approved):

PR #601, opened 2024-04-12
JsonUtility improvements

paulhazen:
This PR addresses a frequent issue in the EOS Plugin where failures during JSON parsing result in unclear error messages. To enhance clarity, a new `ValidateJson` method has been introduced. This method is now employed across all functions that process JSON data. It ensures that errors outside the editor are logged as warnings or errors, while within the editor, it throws descriptive exceptions to aid in identifying and resolving JSON issues.

Additionally, to streamline operations that involve reading and parsing JSON files, a new method `FromJsonFile<T>(string path)` has been added. This method combines reading the file and parsing its contents into JSON, simplifying the process for developers.

### DIFF
--- a/Assets/Plugins/Source/Core/Utility/JsonUtility.cs
+++ b/Assets/Plugins/Source/Core/Utility/JsonUtility.cs
@@ -22,21 +22,112 @@
 
 namespace PlayEveryWare.EpicOnlineServices.Utility
 {
+    using Newtonsoft.Json;
+    using Newtonsoft.Json.Linq;
+    using System;
+    using UnityEngine;
+
+    /// <summary>
+    /// Contains functions to interact with various json data.
+    /// </summary>
     public static class JsonUtility
     {
+        /// <summary>
+        /// Attempts to parse the given JSON string. If problems are encountered, an error will be logged and the exception will continue.
+        /// </summary>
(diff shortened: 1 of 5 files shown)


andrew-hirata-playeveryware:
I think you accidentally made this PR against `stable` instead of against `development`.

andrew-hirata-playeveryware (approved):

Assets/Plugins/Source/Core/Utility/JsonUtility.cs
```diff
@@ -22,21 +22,112 @@
 
 namespace PlayEveryWare.EpicOnlineServices.Utility
 {
+    using Newtonsoft.Json;
+    using Newtonsoft.Json.Linq;
```
arthur740212, line 26: I am kind of unsure about these extra dependencies. IIs there a way to work around this?

arthur740212 (changes requested):

Assets/Plugins/Source/Core/Utility/JsonUtility.cs
paulhazen: The Newtonsoft stuff is already included in the project, so no new dependencies should be added.

arthur740212 (approved):

PR #604, opened 2024-04-15
Workflow for Documentation Publishing

paulhazen:
## Document Publishing Automation

> [!WARNING]
> While it _looks_ like a lot of files are changed, this is because of a template being added to our `etc/docfx/` directory, which is necessary to appropriately place our image asset.

All files in the path `etc/docfx/templates/` were added because of this. I will annotate / comment on the relevant changes.

This PR creates a GitHub action workflow that in concert with the configured `docfx` directory will generate documentation and publish the changes to our website.

> [!NOTE]
> Here is an overview of the security risks and how they are mitigated:
> - Exposing credentials for SFTP: The credentials, website information, and even the directory are stored as GitHub "repository secrets," so they are not exposed.
> - Bandwidth costs: The workflow can only be manually triggered, so no outside collaborators can inadvertently (or otherwise) cause changes to be made to the website.

A few changes were made to some of the `*.cs` files in the project, because the manner in which `docfx` functions is such that it runs a sort-of compilation on the scripts. Since the scripting defines exigent in our project were never designed for a scenario where something _other_ than unity was compiling them, a few gaps were discovered and closed.


Assets/Plugins/Source/Core/EOSCreateOptions.cs
```diff
@@ -44,7 +44,7 @@ public class EOSCreateOptions : IEOSCreateOptions
 
         #region Platform-Specific properties
 
-#if UNITY_PS5 || UNITY_GAMECORE || UNITY_SWITCH || UNITY_ANDROID || UNITY_IOS
+#if !(UNITY_STANDALONE_WIN || UNITY_EDITOR_WIN || UNITY_STANDALONE_LINUX || UNITY_EDITOR_LINUX || UNITY_STANDALONE_OSX || UNITY_EDITOR_OSX)
         public IntegratedPlatformOptionsContainer IntegratedPlatformOptionsContainerHandle { get => options.IntegratedPlatformOptionsContainerHandle; set => options.IntegratedPlatformOptionsContainerHandle = value; }
 #endif
```
paulhazen: This was changed so that the conditional for the existence of the property as defined on the _interface_ would be the same as the conditional for the _implementation_ of the property.

andrew-hirata-playeveryware (approved):

arthur740212 (approved):

PR #605, opened 2024-04-15
chore(wsa, dynamic_loading): remove code associated with wsa dynamic

andrew-hirata-playeveryware:
# Intro 
Early in the project's history, there was some hopeful support for Windows Store App (WSA) added to the plugin. This was done to round out the number of platforms our plugin supported. Eventually, support for this platform fell by the wayside, and eventually was dropped. It actually might be more accurate to say that it wasn't ever really supported, and the code added to the plugin to support that platform doesn't _really_ work.

# Overview of changes
This particular PR removes the code that attempted to do 'clever' dynamic loading things so it could determine which version of Windows the plugin was running on, 32-bit or 64-bit. Given that it doesn't actually work, and makes the dynamic library loading code even more difficult to read, I believe it is time to get rid of the code

### DIFF
--- a/Assets/Plugins/Source/Core/EOSManager_DynamicLoading.cs
+++ b/Assets/Plugins/Source/Core/EOSManager_DynamicLoading.cs
@@ -24,7 +24,7 @@
 #define USE_EOS_GFX_PLUGIN_NATIVE_RENDER
 #endif
 
-#if UNITY_64 || UNITY_WSA
+#if UNITY_64
 #define PLATFORM_64BITS
 // As far as I know, on Windows, if it isn't 64 bit, it's 32 bit
 #elif (UNITY_EDITOR_WIN || UNITY_STANDALONE_WIN)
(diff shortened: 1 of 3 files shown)


Assets/Plugins/Source/Core/EOSManager_DynamicLoading.cs
```diff
@@ -20,15 +20,10 @@
...
-//#define USE_MANUAL_METHOD_LOADING
-
-// TODO: confirm that the overlay doesn't work on macOS
 #if UNITY_STANDALONE_WIN && !UNITY_EDITOR && !UNITY_IOS && !UNITY_STANDALONE_OSX
 #define USE_EOS_GFX_PLUGIN_NATIVE_RENDER
 #endif
 
-//#define USE_WSA_DYNAMIC_LOAD
-
 #if UNITY_64 || UNITY_WSA
```
paulhazen: If we are removing Windows Store Application support, we probably want to get rid of the second part of this conditional.

Assets/Plugins/Source/Core/EOSManager_DynamicLoading.cs
```diff
@@ -81,9 +76,6 @@ public partial class EOSSingleton
                 "GfxPluginNativeRender-unknown";
 #endif
 
-            [DllImport(GfxPluginNativeRenderPath)]
-            static extern void UnloadEOS();
```
paulhazen, line 85: Was this only ever used by WSA? The generic name is throwing me off I think.

paulhazen (changes requested):
Very minor change requested. Otherwise looks good.

arthur740212 (approved):
As long as these removed functions are actually WSA specific, should be good.

paulhazen (approved):

PR #614, opened 2024-04-23
Hotfix 3.1.2

paulhazen:
## Hotfixes
- Corrects version of Unity referenced within `package.json` generated by the Create Package Window.
- Moves platform-specific implementations of `EOSCreateOptions` and `EOSInitializeOptions` into platform-specific files.
- Remove the centralized means by which each platform's config type and dynamic library extension is defined, allowing the specific platform files to describe those details.
- Replaces explicit instantiation of platform-specific `PlatformConfigEditor` classes with a reflection-based approach that instantiates a config editor for each platform that exists (allowing for future platform PlatformConfigEditors to _automagically_ just work).
- Removes dependency errantly introduced within the `BuildUtility` class.
- Reorders the order of the conditional branch within `BuildRunner`.
- Updates the version number accordingly.


Assets/Plugins/Android/Core/AndroidPlatformSpecifics.cs
```diff
@@ -30,24 +30,36 @@
...
+
+    public class EOSInitializeOptions
+    {
+        public AndroidInitializeOptions options;
+    }
+
     //-------------------------------------------------------------------------
     // Android specific Unity Parts.
-    public class EOSPlatformSpecificsAndroid : PlatformSpecifics<AndroidConfig>
+    public class AndroidPlatformSpecifics : PlatformSpecifics<AndroidConfig>
```
andrew-hirata-playeveryware: I would argue that as a hotfix this should not rename any classes unless it's essential that they be renamed.

Assets/Plugins/Source/Core/PlatformManager.cs
```diff
@@ -99,27 +99,40 @@ public static Platform CurrentPlatform
...
-            AddPlatformInfo(Platform.Linux, "Linux", "eos_linux_config.json", typeof(LinuxConfig), ".so");
-            AddPlatformInfo(Platform.macOS, "macOS", "eos_macos_config.json", typeof(MacOSConfig), ".dylib");
-            AddPlatformInfo(Platform.Steam, "Steam", "eos_steam_config.json", typeof(EOSSteamConfig), "");
+            AddPlatformInfo(Platform.Android,     "Android",       "eos_android_config.json"    );
+            AddPlatformInfo(Platform.iOS,         "iOS",           "eos_ios_config.json"        );
+            AddPlatformInfo(Platform.Linux,       "Linux",         "eos_linux_config.json"      );
+            AddPlatformInfo(Platform.macOS,       "macOS",         "eos_macos_config.json"      );
+            AddPlatformInfo(Platform.Steam,       "Steam",         "eos_steam_config.json"      );
+            AddPlatformInfo(Platform.XboxOne,     "Xbox One",      "eos_xb1_config.json"        );
+            AddPlatformInfo(Platform.XboxSeriesX, "Xbox Series X", "eos_xboxseriesx_config.json");
```
andrew-hirata-playeveryware: This config file name is incorrect: it should be `eos_xsx_config.json`. Changing the name of the config file would be a breaking change, i.e. can't be in a hotfix.

andrew-hirata-playeveryware (changes requested):

Assets/Plugins/Android/Core/AndroidPlatformSpecifics.cs
```diff
@@ -30,24 +30,36 @@
 using JavaVM = System.IntPtr;
 using System.Diagnostics;
 
-#if UNITY_ANDROID && !UNITY_EDITOR
+#if UNITY_ANDROID
```
arthur740212: !UNITY_EIDTOR can't be removed, it causes a compile error

Assets/Plugins/Android/Core/AndroidPlatformSpecifics.cs
paulhazen: Yep! Good catch; I did this out of habit.

Assets/Plugins/iOS/Core/IOSPlatformSpecifics.cs
```diff
@@ -24,13 +24,25 @@
 using UnityEngine.Scripting;
 using System.Runtime.InteropServices;
 
-#if UNITY_IOS && !UNITY_EDITOR
+#if UNITY_IOS
```
arthur740212: Similarly, !UNITY_EDITOR can't be removed

Assets/Plugins/Source/Editor/Build/BuildRunner.cs
arthur740212, line 1: On line 60, because BuildRunner is an Editor only script, the process will never get pass this check, blocking any platform after this

arthur740212: One temporary fix is to move all platforms before Windows

Assets/Plugins/Source/Core/EOSManager.cs
arthur740212, line 1: The networkStatusUpdate call needs to be changed too

andrew-hirata-playeveryware (approved):

arthur740212 (approved):

PR #615, opened 2024-04-23
Config Caching

paulhazen:
Based on recent development, I noticed that the config files often get read many times. In non-editor scenarios, the config files are never changed, so this PR includes a caching mechanism that allows the config files to be read only once, then stored in memory for future reference. The memory footprint of these config files is very (very) small, so the performance benefit is an over-all improvement.

Along the way I added comments to functions that were missing them.


andrew-hirata-playeveryware:
The only worry I have about this, is I'm not sure how the caching would play with doing git-checkouts of the config file or with manual commits to the config file: i.e. how will it reload the config file if the only thing that changes is the config file?

paulhazen:
> The only worry I have about this, is I'm not sure how the caching would play with doing git-checkouts of the config file or with manual commits to the config file: i.e. how will it reload the config file if the only thing that changes is the config file?

If I'm understanding your concern correctly - you are describing the scenario where this caching mechanism is used, and then outside of the Unity Editor (either manually or as a result of a git pull) the config file is updated. Your concern is that when that happens, the update will not be reflected because the config values that were cached before the edit would still be used.

Unless I misunderstand things (a distinct possibility) I don't think this is a possible scenario. 

The caching mechanism is `#if` blocked around `!UNITY_EDITOR`, meaning it only caches things when it is running standalone or on a console - and both scenarios would require re-deployment / relaunch to update the config file to take effect regardless of the caching mechanism.

andrew-hirata-playeveryware:
Oh I misread the code with the `#ifdefs`: that's my mistake. Which scenarios are reading the config files multiple times? It should be the case that they either refer to the cached EOSConfig we have in EOSManger, or if each platform needs it, it can store values in their own PlatformSpecific implementation.

paulhazen:
> Oh I misread the code with the `#ifdefs`: that's my mistake. Which scenarios are reading the config files multiple times? It should be the case that they either refer to the cached EOSConfig we have in EOSManger, or if each platform needs it, it can store values in their own PlatformSpecific implementation.

There are a lot of places where we were being inconsistent about how we obtained the config information. It's within the `[X]PlatformSpecifics` implementations that re-reading the file most often takes place. 

Certainly the `EOSManager` was doing the best job of using the cached value, but over time I think the existence of that cached value was not obvious to some developers, and reading in the config file was re-implemented in several places. This feature is designed to provide the caching at the data layer instead of at the logic layer - that way it can be abstracted and we can "Get" the config value by just going:

```c#
AndroidConfig androidConfig = Config.Get<AndroidConfig>();
```
or
```c#
WindowsConfig windowsConfig = Config.Get<WindowsConfig>();
```

and while running outside of the context of an editor there will be no possibility that the file can even _accidentally_ be read twice. This seemed better than adding a caching mechanism to the `PlatformSpecifics` implementing class - since the responsibility for caching the data is common to all `Config` classes, not just `Platform` config classes. This also addresses the scenario where multiple `PlatformSpecific` classes could be instantiated over the course of the runtime - if the cached value was local to the instance of the config, then the cache would only be valid for the lifetime of the object.

andrew-hirata-playeveryware:
Okay, those are reasonable reasons to add this caching.

andrew-hirata-playeveryware (approved):

arthur740212 (approved):

PR #626, opened 2024-04-29
fix: ensure that correct EpicAccountId is being used for connect login.

### LINKED ISSUE #576: Samples, steam account linking, loginCallbackInfo.LocalUserId is null
I have encountered a null exception and I've fixed it in my project, so I'm sharing the fix.

The error log:
image

In the sample "Standard Pack : EOS Plugin for Unity Samples", a script named UILoginMenu.cs (I used it as reference for steam login, steam account linking with EOS on the first session).

The method named StartConnectLoginWithLoginCallbackInfo(LoginCallbackInfo loginCallbackInfo), the paramater LoginCallbackInfo is used to get loginCallbackInfo.LocalUserId. 
But in the case of "loginCallbackInfo.ResultCode == Epic.OnlineServices.Result.InvalidUser" (User is linking steam with an active Epic account), the loginCallbackInfo.LocalUserId was null in my case, so I needed to use linkAccountCallbackInfo.LocalUserId instead (callback of AuthLinkExternalAccountWithContinuanceToken).

I changed StartConnectLoginWithLoginCallbackInfo parameter of LoginCallbackInfo to EpicAccountId, and to send the LocalUserId accordingly.

Like so:

image

andrew-hirata-playeveryware:
# Intro 
The previous code was using a stale EpicAccountId that is sometimes an invalid one. This issue was reported in #576. 

# Overview of changes
The new one ensures that the most recently used EpicAccountId is the one passed back via the callback, which should ensure that it isn't every NULL.

### DIFF
--- a/Assets/Scripts/UI/Login/UILoginMenu.cs
+++ b/Assets/Scripts/UI/Login/UILoginMenu.cs
@@ -1258,9 +1258,9 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
         }
 
         //-------------------------------------------------------------------------
-        private void StartConnectLoginWithLoginCallbackInfo(LoginCallbackInfo loginCallbackInfo)
+        private void StartConnectLoginWithEpicAccount(EpicAccountId LocalUserId)
         {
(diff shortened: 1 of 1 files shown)


paulhazen (approved):
Makes sense. Good catch.

arthur740212 (approved):

PR #631, opened 2024-04-30
Hotfix 3.1.3

### LINKED ISSUE #620: Fail to find RTCOptions in 3.1.2 UNITY_EDITOR_OSX (3.1.1 works)
Library/PackageCache/com.playeveryware.eos@c13536130d/Runtime/Core/PlatformSpecifics.cs(91,55): error CS1061: 'Options' does not contain a definition for 'RTCOptions' and no accessible extension method 'RTCOptions' accepting a first argument of type 'Options' could be found (are you missing a using directive or an assembly reference?)

arthur740212:
Hotfix 3.1.3

- Mac create option fix
- plugin deployment check

### DIFF
--- a/Assets/Plugins/Source/Editor/Utility/BuildUtility.cs
+++ b/Assets/Plugins/Source/Editor/Utility/BuildUtility.cs
@@ -70,25 +70,17 @@ namespace PlayEveryWare.EpicOnlineServices.Build
         /// <returns>True if the plugin is deployed as UPM, false otherwise.</returns>
         private static bool IsDeployedAsUPM()
         {
-            try
+            string packagePathname = Path.GetFullPath(Path.Combine("Packages", EOSPackageInfo.PackageName));
+
+            if (Directory.Exists(packagePathname))
             {
-                string manifestJson = FileUtility.ReadAllText(
-                    Path.Combine(FileUtility.GetProjectPath(), "Packages", "manifest.json")
-                );
-
-                Manifest manifest = JsonUtility.FromJson<Manifest>(manifestJson);
-
-                if (manifest != null && manifest.dependencies.ContainsKey(EOSPackageInfo.PackageName))
-                {
-                    return true;
-                }
+                Debug.Log("Deployed via UPM");
+                return true;
             }
-            catch (Exception e)
+            else
             {
-                Debug.LogWarning($"There was a problem determining if deployed via UPM: \"{e.Message}\".");
+                return false;
             }
-
-            return false;
         }
 
         /// <summary>
--- a/Assets/Plugins/Source/Core/EOSPackageInfo.cs
+++ b/Assets/Plugins/Source/Core/EOSPackageInfo.cs
@@ -35,7 +35,7 @@ namespace PlayEveryWare.EpicOnlineServices
          * not involve editing source code files.
          */
 
-        public const string Version = "3.1.2";
+        public const string Version = "3.1.3";
 
         public const string PackageName = "com.playeveryware.eos";
     }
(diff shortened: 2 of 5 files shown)


Assets/Plugins/Source/Editor/Utility/BuildUtility.cs
```diff
@@ -70,25 +70,39 @@ public static bool DeployedAsUPM
...
+
+            //    if (manifest != null && manifest.dependencies.ContainsKey(EOSPackageInfo.PackageName))
+            //    {
+            //        return true;
+            //    }
+            //}
+            //catch (Exception e)
+            //{
+            //    Debug.LogWarning($"There was a problem determining if deployed via UPM: \"{e.Message}\".");
+            //}
```
paulhazen: Good catch. This was working properly when we were using `Newtonsoft.Json`, because it _does_ support dictionaries.

Assets/Plugins/Source/Editor/Utility/BuildUtility.cs
```diff
@@ -70,25 +70,39 @@ public static bool DeployedAsUPM
...
+
+            string packagePathname = Path.GetFullPath(Path.Combine("Packages", EOSPackageInfo.PackageName));
+
+            if (Directory.Exists(packagePathname))
+            {
+                Debug.Log("Deployed via UPM");
+                return true;
```
paulhazen: This is an insufficient test to determine whether or not the project is running as a UPM. For instance, the `manifest.json` file can define a local path or a Git URL to the package - in which case this particular test will return a false negative since the corresponding directory in `Packages` may not exist.

My suggestion would be to alter the original approach to read the `manifest.json` file line by line and check if there is a line that contains `EOSPackageInfo.PackageName`.

Only, that can't be guaranteed to work either, since the `manifest.json` file does not necessarily have entries on new-lines (the json file could all be on one line and would be valid).

So probably the best approach would be to read the contents of `manifest.json` into a string, split that string by `;`, then check each entry to see if it contains `EOSPackageInfo.PackageName`.

Assets/Plugins/Source/Editor/Utility/BuildUtility.cs
```diff
@@ -70,25 +70,39 @@ public static bool DeployedAsUPM
...
-                if (manifest != null && manifest.dependencies.ContainsKey(EOSPackageInfo.PackageName))
-                {
-                    return true;
-                }
+            //try
+            //{
+            //    /////////////////////////////////////////////////////////
+            //    /// JsonUtility.FromJson does not support Dictionaries///
+            //    /// Causing this check to always return false         ///
+            //    /////////////////////////////////////////////////////////
```
paulhazen: I think as a practice we should try and have this articulation made in the body of the pull request, and the code that needs to be changed or removed should be removed instead of just commented out.

paulhazen (changes requested):

Assets/Plugins/Source/Editor/Utility/BuildUtility.cs
andrew-hirata-playeveryware: Path.GetFullPath is the Unity 'magic' way to resolve UPM based paths to absolute paths, at least according to their [documentation](https://docs.unity3d.com/Manual/upm-assets.html#GetFullPath). I'm fairly confident that it should work in this situation as well.

Assets/Plugins/Source/Editor/Utility/BuildUtility.cs
andrew-hirata-playeveryware: I can agree with that.

andrew-hirata-playeveryware (approved):

Assets/Plugins/Source/Editor/Utility/BuildUtility.cs
paulhazen: > Path.GetFullPath is the Unity 'magic' way to resolve UPM based paths to absolute paths, at least according to their [documentation](https://docs.unity3d.com/Manual/upm-assets.html#GetFullPath). I'm fairly confident that it should work in this situation as well.

Oh that's right! Dang, I keep forgetting that Unity be like that. I'll mark this conversation as resolved. My mistake.

paulhazen (approved):

PR #634, opened 2024-04-30
chore: Prune unreferenced files

paulhazen:
Prior to deleting each file, the files in the `Assets` directory were recursively inspected using the `grep` utility to determine whether the function signatures and/or class names contained in the file to delete were referenced anywhere. 

It was confirmed that they were not referenced in this project, or in any other related projects.


andrew-hirata-playeveryware:
Posting this for posterity: 
The AndroidInitOptions exists due to a few releases where we didn't have C# interfaces / code for some of the android code, so I hand rolled some solutions to get it to work. Since we have had C# code for Android for many releases now, it makes sense to get rid of it.

The 'Default platform' was originally there so that we could have a 'fallback' that could log different kinds of errors if a new platform was added (i.e. give better errors than just throwing a null ptr exception), and additionally work as a template for new platforms. It's also reasonable to get rid of it now.

andrew-hirata-playeveryware (approved):

arthur740212 (approved):

PR #637, opened 2024-04-30
fix(steam): resize buffer before fetching App ticket and Session ticket.

### LINKED ISSUE #581: Possible typos in SteamManager edge case?
Minor things, but in my travels exploring the sample code (while working on our implementation) I believe there may be some accidental typos in edge case code inside the sample SteamManager.cs.

I may be wrong... and even if I'm right, it might be extremely unlikely this code even gets called, but still thought I'd point it out.

1) In RequestAppTicketResponse, where it realises it requires a larger buffer, it's calling CancelAuthTicket.

It would seem this was copy pasted from the similar code related to Auth Tickets, but doesn't appear to be valid in this context.

2) In both RequestAppTicketResponse and GetSessionTicket, where they determine the ticket is too larger for their buffer, they increase their bufferSize int variable, but fail to actually increase the size of the buffer byte array itself.

It would seem to me that both these code paths should have:

buffer = new byte[bufferSize];

Inserted before they retry their 'get'.

andrew-hirata-playeveryware:
# Intro 
This is an interesting find by a [user](https://github.com/PlayEveryWare/eos_plugin_for_unity/issues/581). The existing code, if it were to fail to grab either ticket type, would potentially cause a buffer overflow. Because 'most' tickets are the default size of 1024, this code defect probably hasn't been seen in the wild yet.

# Overview of changes
There are three changes:
* Remove unneeded CancelAuthTicket, since it was added to a call before getting the Encrypted App ticket (not needed)
* Allocate a new byte Array to fit the Encrypted App ticket when our default array isn't large enough.
* Allocate a new byte Array to fit the Auth Session ticket when our default array isn't large enough.

### DIFF
--- a/Assets/Scripts/Steam/SteamManager.cs
+++ b/Assets/Scripts/Steam/SteamManager.cs
@@ -81,10 +81,11 @@ namespace PlayEveryWare.EpicOnlineServices.Samples.Steam
             byte[] buffer = new byte[bufferSize];
             uint ticketSize = 0;
             bool success = SteamUser.GetEncryptedAppTicket(buffer, bufferSize, out ticketSize);
+
             if (!success && (int)ticketSize > bufferSize)
             {
-                SteamUser.CancelAuthTicket(sessionTicketHandle);
                 bufferSize = (int)ticketSize;
+                buffer = new byte[bufferSize];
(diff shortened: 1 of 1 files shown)


paulhazen (approved):
Nice catch on the user's part. Good fix on yours :)

arthur740212 (approved):

PR #641, opened 2024-05-02
Enforce factory pattern for classes that derive from `Config`

paulhazen:
In light of the changes made in PR #615 that introduced Config caching, it occurred to me that we should enforce the factory pattern with Config classes to make sure that they are all properly kept track of. 

To that end, I changed the access modifier on the constructor of all implementing classes from `public` to `protected`, introduced static methods on the baser abstract `Config` class that handle the factory aspects, and added a static constructor to each so that the registration of the factory methods can be handled by the implementing class (this allows flexibility for adding new platforms in the future because platform configs can register their own constructors instead of having the factory (static context within `Config`) keep track of the constructors for each in a discrete list.

While doing this work, I also renamed `EOSSteamConfig` and `EOSSteamInternal` to `SteamConfig` and `SteamInternal` respectively, which reflects the naming convention we use with other `Config` and Config-related objects.

In addition, I added function comments where necessary, and re-ordered the various `Config` class member variables to precede the static constructor within each.


Assets/Plugins/Source/Core/Config.cs
```diff
@@ -83,26 +89,76 @@ protected Config(string filename, string directory)
...
         }
 
+        /// <summary>
+        /// Allows deriving classes to register their constructor method in order to enforce the
+        /// factory pattern. This requires that each class that derives from Config must implement
+        /// a static method registering it's constructor.
+        /// </summary>
+        /// <typeparam name="T">The config type.</typeparam>
+        /// <param name="factory">The function to create the config type</param>
+        protected static void RegisterFactor<T>(Func<Config> factory) where T : Config
```
WispyMouse: What leads towards wanting to name this "RegisterFactor" instead of "RegisterFactory"?

Assets/Plugins/Source/Core/Config.cs
```diff
@@ -83,26 +89,76 @@ protected Config(string filename, string directory)
             Directory = directory;
         }
 
+        /// <summary>
+        /// Allows deriving classes to register their constructor method in order to enforce the
+        /// factory pattern. This requires that each class that derives from Config must implement
+        /// a static method registering it's constructor.
```
WispyMouse: nit: "it's constructor" => "its constructor"

Assets/Plugins/Source/Core/Config.cs
```diff
@@ -83,26 +89,76 @@ protected Config(string filename, string directory)
...
+        /// Try to retrieve the factory method for the indicated type that can be used to
+        /// create a new instance of the given config type.
+        /// </summary>
+        /// <typeparam name="T">The type of config to get the factory method for.</typeparam>
+        /// <param name="factory">The factory method that instantiates the config indicated.</param>
+        /// <returns>True if the factory is registered, false otherwise.</returns>
+        /// <exception cref="InvalidOperationException">
+        /// If the indicated type does not have a corresponding constructor function
+        /// registered, a verbose exception will be thrown indicating how to properly
+        /// implement the Config implementing class such that it's constructor is
```
WispyMouse: nit: "it's constructor" => "its constructor"

WispyMouse: I'll leave this here as a general reminder of this small grammatical change, it applies in some other places.

Assets/Plugins/Source/Core/Config.cs
```diff
@@ -83,26 +89,76 @@ protected Config(string filename, string directory)
...
+            //       not be cached.
 #if !UNITY_EDITOR
+            // Determine if there is a cached copy of the config, return that if it exists.
             if (s_cachedConfigs.TryGetValue(typeof(T), out Config config))
             {
                 return (T)config;
             }
 #endif
+            // Try to get the factory method used to instantiate the config.
+            TryGetFactory<T>(out Func<Config> factory);
```
WispyMouse, line 179: This looks like it should check to see if the factory is successfully gotten from the TryGetFactory function, before using the factory below.

Assets/Plugins/Source/Core/Config.cs
```diff
@@ -83,26 +89,76 @@ protected Config(string filename, string directory)
...
+
+            // Use the factory method to create the config.
+            T instance = (T)factory();
 
-            T instance = new();
+            // Asynchronously read the config values from the corresponding file.
             await instance.ReadAsync();
```
WispyMouse, line 185: Not related to this review: I'm surprised a method named "ReadAsync" just passes back a Task, without having it return what the last read string was.

Assets/Plugins/Source/Core/Config.cs
```diff
@@ -111,21 +167,34 @@ public static async Task<T> GetAsync<T>() where T : Config, new()
...
 #if !UNITY_EDITOR
+            // Determine if there is a cached copy of the config, return that if it exists.
             if (s_cachedConfigs.TryGetValue(typeof(T), out Config config))
             {
                 return (T)config;
             }
 #endif
-            T instance = new();
+            // Try to get the factory method used to instantiate the config.
+            TryGetFactory<T>(out Func<Config> factory);
```
WispyMouse, line 215: Similar to GetAsync above, should this check to see that the factory is retrieved before being used?

WispyMouse (approved):
nits on "it's" grammar;
a question about the named "Factor" versus "Factory;
wanting to use the boolean return on TryGet;
and unrelated musing on "Read" not passing back values.

None of these are essential changes. Review marked as Approve.

Assets/Plugins/Source/Core/Config.cs
paulhazen: Typo - I actually had the function signature fixed on my local branch, looks like I had some unpushed changes. Should be fixed now.

Assets/Plugins/Source/Core/Config.cs
paulhazen: The reason I did not check for this is that the `TryGetFactory` method throws an `InvalidOperationException` if the factory is not found. I did not want to design it in such a way where it would ever be "okay" that `TryGetFactory` returned false - but I wanted to follow the naming convention and pattern that usually accompanies a "`TryGet[Insert Noun Here]`" method in other places of C#.

Assets/Plugins/Source/Core/Config.cs
paulhazen: I can see how that could be misleading. The `Read` and `ReadAsync` functions do not actually return anything, they perhaps more appropriately could be named `ReadValuesFromFile` or something like that - because as the function comments explain:
```c#
/// <summary>
/// Synchronously reads the contents of a Config from the json file that defines it.
/// </summary>
protected virtual void Read() { /*...*/ }
```

The reason that the `ReadAsync` function returns a `Task` is that it is required to do so in-order to fulfill the `async` modifier - which makes the method `awaitable`. 

```c#
/// <summary>
/// Asynchronously read the values from the JSON file associated with this Config
/// </summary>
/// <returns>Task.</returns>
protected virtual async Task ReadAsync() { /*...*/ }
```

That said (and within the context of the code) the function names `ReadAsync` and `Read` seemed sufficiently descriptive and certainly more succinct than `ReadValuesFromFileIntoMemory` or something of similar ilk.

Assets/Plugins/Source/Core/Config.cs
paulhazen: Basically, I wanted to require a `try` block to suppress the lack of a factory, that way it cannot fail silently.

Assets/Plugins/Source/Core/Config.cs
paulhazen: I pride myself on my grammar - but I continue to make this mistake 😆

Assets/Plugins/Source/Core/Config.cs
paulhazen: I thought that was the case, but it was not. _Now_ it should be fixed.

Assets/Plugins/Source/Core/Config.cs
paulhazen: Addressed in first comment. Marking conversation as "closed."

andrew-hirata-playeveryware (changes requested):
@paulhazen, Could you kindly add the unit tests that we spoke about earlier?

Assets/Plugins/Source/Editor/Configs/SigningConfig.cs
```diff
@@ -24,14 +24,45 @@
...
+        /// </summary>
         public List<string> dllPaths;
+
+        static SigningConfig()
+        {
+            RegisterFactory(() => new SigningConfig());
+        }
+
+        protected SigningConfig() : 
+            base("eos_plugin_signing_config.json") { }
```
andrew-hirata-playeveryware: I believe that this should be on the same line as the `:`

paulhazen:
> @paulhazen, Could you kindly add the unit tests that we spoke about earlier?

Donzo! :)

Assets/Plugins/Source/Editor/Configs/AndroidBuildConfig.cs
```diff
@@ -26,8 +26,14 @@ namespace PlayEveryWare.EpicOnlineServices.Editor.Config
...
-
         public bool DynamicallyLinkEOSLibrary;
+
+        static AndroidBuildConfig()
+        {
+            RegisterFactory(() => new AndroidBuildConfig());
+        }
+
+        protected AndroidBuildConfig() : 
+            base("eos_plugin_android_build_config.json") { }
```
andrew-hirata-playeveryware: This 'base' should be on the same line as the class.

Assets/Plugins/Source/Editor/Configs/LibraryBuildConfig.cs
```diff
@@ -23,10 +23,16 @@ namespace PlayEveryWare.EpicOnlineServices.Editor.Config
...
         public string makePath;
         public bool msbuildDebug;
+
+        static LibraryBuildConfig()
+        {
+            RegisterFactory(() => new LibraryBuildConfig());
+        }
+
+        protected LibraryBuildConfig() : 
+            base("eos_plugin_library_build_config.json") { }
```
andrew-hirata-playeveryware: This invocation of the base constructor should be on the same line as the `:`.

Assets/Plugins/Source/Editor/Configs/PackagingConfig.cs
```diff
@@ -24,9 +24,16 @@ namespace PlayEveryWare.EpicOnlineServices.Editor.Config
...
         public string pathToJSONPackageDescription;
         public string pathToOutput;
+
+        static PackagingConfig()
+        {
+            RegisterFactory(() => new PackagingConfig());
+        }
+
+        protected PackagingConfig() :
+            base("eos_plugin_packaging_config.json") { }
```
andrew-hirata-playeveryware: This invocation of the base constructor should be on the same line as the `:`

Assets/Plugins/Source/Editor/Configs/PrebuildConfig.cs
```diff
@@ -27,7 +27,19 @@ namespace PlayEveryWare.EpicOnlineServices.Editor.Config
...
+        /// </summary>
         public bool useAppVersionAsProductVersion;
+
+        static PrebuildConfig()
+        {
+            RegisterFactory(() => new PrebuildConfig());
+        }
+
+        protected PrebuildConfig() : 
+            base("eos_plugin_version_config.json") { }
```
andrew-hirata-playeveryware: This invocation of the base constructor should be on the same line as the `:`

Assets/Plugins/Source/Editor/Configs/ToolsConfig.cs
```diff
@@ -31,20 +31,34 @@ namespace PlayEveryWare.EpicOnlineServices.Editor.Config
...
+        /// </summary>
         public bool useEAC;
+
+        static ToolsConfig()
+        {
+            RegisterFactory(() => new ToolsConfig());
+        }
+
+        protected ToolsConfig() : 
+            base("eos_plugin_tools_config.json") { }
```
andrew-hirata-playeveryware: This invocation of the base constructor should be on the same line as the `:`

andrew-hirata-playeveryware (changes requested):

arthur740212:
There is a conflict that isn't resolved yet

PR #648, opened 2024-05-03
Correct `MacOSBuilder`'s `PreBuild` step

paulhazen:
While investigating a related issue I discovered that the `MacOSBuilder`'s `PreBuild` function was inappropriately overridden. Per the documentation on the base implementation of `PreBuild` within `PlatformSpecificBuilder`:

```c#
/// <summary>
/// Implement this function on a per-platform basis to provide custom logic for the platform being compiled.
/// Any overriding implementations should first call the base implementation.
/// </summary>
/// <param name="report"></param>
public virtual void PreBuild(BuildReport report)
{
    // Check to make sure that the platform configuration exists
    CheckPlatformConfiguration();

    // Configure the version numbers per user defined preferences
    ConfigureVersion();

    // Check to make sure that the binaries for the platform exist, and build them if necessary.
    CheckPlatformBinaries();
}
```

I'm not certain how this got introduced - very likely it's my oversight. But it's a good reminder to make sure you read the documentation for any function you are overriding. In this particular case the overridden function was using an old means of checking to see that the binaries existed - but as you'll note from the base implementation above there is more that this function does, and those steps were being ignored.


arthur740212 (approved):
Does the checks check from both the package path and the project path?

Assets/Plugins/macOS/Editor/MacOSBuilder.cs
```diff
@@ -29,26 +29,14 @@ namespace PlayEveryWare.EpicOnlineServices.Build
...
-
-            string macOSPluginFolder = Path.Combine(Application.dataPath, "Plugins", "macOS");
-            if (!File.Exists(Path.Combine(macOSPluginFolder, "libDynamicLibraryLoaderHelper.dylib")) ||
-                !File.Exists(Path.Combine(macOSPluginFolder, "MicrophoneUtility_macos.dylib")))
-            {
-                // TODO: Implement functionality to actually do this for the user, instead of prompting them to do it.
-                Debug.LogError(
-                    "Custom native libraries missing for mac build, use the makefile in lib/NativeCode/DynamicLibraryLoaderHelper_macOS to install the libraries");
-            }
+            AddProjectFileToBinaryMapping(
```
andrew-hirata-playeveryware, line 36: I'm having deja vu here, but It looks like the `AddProjectFileToBinaryMapping` adds the information to an array that is used in a way that does not handle when the code is run from `packages` (i.e. when it's installed as a UPM).

andrew-hirata-playeveryware (changes requested):

Assets/Plugins/macOS/Editor/MacOSBuilder.cs
paulhazen: I will look into this - I believe that at the moment in time when the code actually checks / resolves the filepath it will handle the situation you mentioned, but I will double check and get back to you regarding that here in this thread.

paulhazen: > I'm having deja vu here, but It looks like the `AddProjectFileToBinaryMapping` adds the information to an array that is used in a way that does not handle when the code is run from `packages` (i.e. when it's installed as a UPM).

This is correct - because the UPM ships with the native code binaries already compiled - so there is no need to compile them.

andrew-hirata-playeveryware (approved):

PR #659, opened 2024-05-08
fix: Remove `Newtonsoft` dependency errantly introduced.

paulhazen:
As the title indicates, this PR removes the dependency on the `Newtonsoft` library used for JSON parsing that was errantly introduced without properly marking it as a dependency. The decision was made to remove the dependency until a time that we decide with more intentionality to add the dependency.

The change in functionality is that if invalid JSON is detected, an error will be logged, however that error will no longer contain detailed information regarding the nature of the invalidity of the JSON in question. 

This PR was originally merged before a subtle change was observed that warranted reverting and re-opening. The original PR was approved by @andrew-hirata-playeveryware, and @WispyMouse was the one to discover what was previously overlooked - so they are the only ones whose review is requested this time around.


WispyMouse (approved):
LGTM, all concerns are addressed.

PR #661, opened 2024-05-08
fix: Remove `Newtonsoft` dependency errantly introduced.

paulhazen:
As the title indicates, this PR removes the dependency on the `Newtonsoft` library used for JSON parsing that was errantly introduced without properly marking it as a dependency. The decision was made to remove the dependency until a time that we decide with more intentionality to add the dependency.

The change in functionality is that if invalid JSON is detected, an error will be logged, however that error will no longer contain detailed information regarding the nature of the invalidity of the JSON in question. 

This PR was originally merged before a subtle change was observed that warranted reverting and re-opening. The original PR was approved by @andrew-hirata-playeveryware, and @WispyMouse was the one to discover what was previously overlooked - so they are the only ones whose review is requested this time around.

The PR was THEN reverted because it was causing some errors in the editor. 

All issues should now be resolved. I am not going to merge into `development` until I get two approving reviews - and even then, will not merge it in until closer to the end of the day - in order to continue work on the next release and incorporate the changes.

### DIFF
--- a/Assets/Plugins/Source/Core/Utility/JsonUtility.cs
+++ b/Assets/Plugins/Source/Core/Utility/JsonUtility.cs
@@ -22,7 +22,6 @@
 
 namespace PlayEveryWare.EpicOnlineServices.Utility
 {
-    using Codice.Client.ChangeTrackerService;
     using UnityEngine;
 
     /// <summary>
--- a/Assets/Plugins/Source/Core/Utility/JsonUtility.cs
+++ b/Assets/Plugins/Source/Core/Utility/JsonUtility.cs
@@ -8,8 +8,8 @@
  * copies of the Software, and to permit persons to whom the Software is
  * furnished to do so, subject to the following conditions:
  *
- * The above copyright notice and this permission notice shall be included in all
- * copies or substantial portions of the Software.
+ * The above copyright notice and this permission notice shall be included in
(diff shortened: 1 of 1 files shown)


Assets/Plugins/Source/Core/Utility/JsonUtility.cs
```diff
@@ -18,13 +18,11 @@
...
  * SOFTWARE.
- */
+*/
 
 namespace PlayEveryWare.EpicOnlineServices.Utility
 {
-    using Newtonsoft.Json;
-    using Newtonsoft.Json.Linq;
-    using System;
+    using Codice.Client.ChangeTrackerService;
```
arthur740212: I wonder what is this for. Maybe another resharper thing

arthur740212 (approved):

WispyMouse (approved):
Compiles, samples have no visible errors.

Assets/Plugins/Source/Core/Utility/JsonUtility.cs
andrew-hirata-playeveryware: I believe this needs to be removed.

andrew-hirata-playeveryware (changes requested):

andrew-hirata-playeveryware (approved):

PR #662, opened 2024-05-08
fix: Properly support behavior when `EOS_DISABLE` is defined.

### LINKED ISSUE #653: Error when using EOS_DISABLE define
Error when using EOS_DISABLE define:
`Library/PackageCache/com.playeveryware.eos@74682501a1/Runtime/macOS/Core/MacOSPlatformSpecifics.cs(47,43): error CS0246: The type or namespace name 'PlatformSpecifics<>' could not be found (are you missing a using directive or an assembly reference?)`

paulhazen:
In some circumstances (some build configurations) if the scripting define `EOS_DISABLE` has been added within the Player Settings window of the Unity project, compilation errors will result.

This PR represents changes that affect the correct implementation of that define. Along the way, missing comments were added to the `IOSBuilder` class, and some very minor formatting changes were made to increase readability.

This closes issue #653.

### DIFF
--- a/Assets/Plugins/iOS/Editor/IOSBuilder.cs
+++ b/Assets/Plugins/iOS/Editor/IOSBuilder.cs
@@ -8,8 +8,8 @@
  * copies of the Software, and to permit persons to whom the Software is
  * furnished to do so, subject to the following conditions:
  *
- * The above copyright notice and this permission notice shall be included in all
- * copies or substantial portions of the Software.
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
  *
  * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
  * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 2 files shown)


Assets/Plugins/iOS/Editor/IOSBuilder.cs
```diff
@@ -46,14 +64,27 @@ public override void PostBuild(BuildReport report)
             string targetGUID = proj.GetUnityMainTargetGuid();
             string unityTargetGUID = proj.GetUnityFrameworkTargetGuid();
 
-            proj.SetBuildProperty(targetGUID, "ENABLE_BITCODE", "false");
-            proj.SetBuildProperty(unityTargetGUID, "ENABLE_BITCODE", "false");
+            proj.SetBuildProperty(
```
andrew-hirata-playeveryware: I'm not certain if I like this style change.

paulhazen: No problem, I reverted to the previous style of having it all on one line. I only realized in retrospect that I had not made separate commits between the style change and the actual fix. Ideally, they would be two different tasks - but decoupling them seemed like more work than it was worth. Easy to fix small preferences like this though.

andrew-hirata-playeveryware (approved):

PR #665, opened 2024-05-09
Log Level Config

arthur740212:
The log levels on startup used to be hard coded.
Now it is loaded from a config file to allow flexibility when debugging with logs.
eos_plugin_for_unity - Lobbies - Windows, Mac, Linux - Unity 2021 3 8f1 _DX11_ 3_15_2024 4_07_36 PM
eos_plugin_for_unity - Lobbies - Windows, Mac, Linux - Unity 2021 3 8f1 _DX11_ 3_15_2024 4_07_41 PM


Assets/Plugins/Source/Core/EOSManager.cs
```diff
@@ -588,19 +590,22 @@ public void Init(IEOSCoroutineOwner coroutineOwner, string configFileName)
...
                     return;
                 }
 
 #if !UNITY_EDITOR && !UNITY_SWITCH
                 // Set logging to VeryVerbose on EOS SDK bootstrap so we get the most logging information
-                SetLogLevel(LogCategory.AllCategories, LogLevel.VeryVerbose);
+                for (int logCategoryIndex = 0; logCategoryIndex < logLevelList.Count; logCategoryIndex++)
+                {
+                    SetLogLevel((LogCategory)logCategoryIndex, logLevelList[logCategoryIndex]);
+                }
```
paulhazen: It's unclear to me why this loop is the same in both places here - does it really need to be repeated?

Assets/Plugins/Source/Core/EOSManager.cs
```diff
@@ -675,12 +680,10 @@ public void Init(IEOSCoroutineOwner coroutineOwner, string configFileName)
...
-                // Default back to quiet logs
-#if UNITY_EDITOR
-                SetLogLevel(LogCategory.AllCategories, LogLevel.VeryVerbose);
-#else
-                SetLogLevel(LogCategory.AllCategories, LogLevel.Warning);
-#endif
+                for (int logCategoryIndex = 0; logCategoryIndex < logLevelList.Count; logCategoryIndex++)
+                {
+                    SetLogLevel((LogCategory)logCategoryIndex, logLevelList[logCategoryIndex]);
+                }
```
paulhazen: Same question here as above; it does not seem like between each of these loops the log level is changing, so I'm confused about why each needs to be run.

Assets/Plugins/Source/Core/LogLevelConfig.cs
```diff
@@ -0,0 +1,48 @@
...
+    public class LogCategoryLevelPair
+    {
+        public string category;
+        public string level;
+
+        public LogCategoryLevelPair(string _category, string _level)
+        {
+            category = _category;
+            level = _level;
+        }
```
paulhazen: I don't think this is a naming convention I have seen before. I would suggest having the parameters named the same, but omit the preceding `_`.

To differentiate between the parameters and the member variables you could either add `this.` before each assignment, or you could make the member variables capitalized (probably the better way to do it - since they are public members).

Assets/Plugins/Source/Core/LogLevelConfig.cs
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    [Serializable]
+    public class LogLevelConfig : Config
```
paulhazen, line 42: Is `LogLevelConfig` only ever utilized within the editor? If so, it should extend from `EditorConfig`.

Assets/Plugins/Source/Core/LogLevelHelper.cs
```diff
@@ -0,0 +1,86 @@
...
+                throw new Exception("Couldn't find Log Level Config file: Please ensure log_level_config.json exists and is a valid config");
+            }
+
+            configDataAsString = File.ReadAllText(logLevelConfigPath);
+#endif
+            var configData = JsonUtility.FromJson<LogLevelConfig>(configDataAsString);
+
+            Debug.Log("Loaded log level config file: " + configDataAsString);
+            return configData;
+        }
```
paulhazen: This function is no longer needed since we have implemented the `Config.Get<T>()` method for all classes that derive from `Config`. See my next comment for what to do so that this whole function can be deleted.

Assets/Plugins/Source/Core/LogLevelHelper.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        public static List<LogLevel> LogLevelList
+        {
+            get
+            {
+                LogLevelConfig logLevelConfig = LoadLogLevelConfigFileFromPath(Path.Combine(Application.streamingAssetsPath, "EOS", "log_level_config.json"));
```
paulhazen: This should be replaced with the following:

```c#
LogLevelConfig logLevelConfig = Config.Get<LogLevelConfig>();
```

Implementing the `Get` function was done so that we don't have to repeat the code written in the function `LoadLogLevelConfigFileFromPath` (see earlier comment).

paulhazen (changes requested):
I think there are a few areas that need some attention. Additionally, I would suggest that instead of having a dropdown for each section that allows us to select what log level to use, have a set of lists - one for each log level, and within each list you can have the category, then the user can drag the category to whichever level they want. This would make the UI a lot more intuitive, and far less noisy.

Assets/Plugins/Source/Core/EOSManager.cs
arthur740212: They are ran under different conditions. One for running in the Editor, the other for Player.

Assets/Plugins/Source/Core/EOSManager.cs
arthur740212: These are for the platforms that uses dynamic binding, which are excluded in the earlier call.
Those using dynamic binding can not use SetLogLevel before the LogInterface gets initialized, or else it would crash.

Assets/Plugins/Source/Core/LogLevelConfig.cs
arthur740212: The log window in the samples will use them in runtime eventually.

Assets/Plugins/Source/Core/LogLevelConfig.cs
andrew-hirata-playeveryware: I have seen a similar naming convention before, but only for the other way around: a leading underscore being used to mark a private variable of a class. I would prefer changing the parameter names to something like `aCategory` and `aLevel` , or using `this` to disambiguate the code.

andrew-hirata-playeveryware (changes requested):

Assets/Plugins/Source/Core/LogLevelHelper.cs
arthur740212: I would mostly agree, but the `Config.Get<T>()` function does not handle file reading on Android.

Assets/Plugins/Source/Core/LogLevelHelper.cs
arthur740212: Used `JsonUtility.FromJson<>()` instead, which handles Android file reading

Assets/Plugins/Source/Core/LogLevelConfig.cs
arthur740212: Fixed!

andrew-hirata-playeveryware (approved):

Assets/Plugins/Source/Editor/Windows/LogLevelWindow.cs
```diff
@@ -0,0 +1,118 @@
...
+
+#if !EOS_DISABLE
+namespace PlayEveryWare.EpicOnlineServices.Editor.Windows
+{
+    public class LogLevelWindow : EOSEditorWindow
+    {
+        LogLevelConfig currentLogLevelConfig;
+
+        string[] categories = LogLevelHelper.LogCategoryStringArray;
+        string[] levels = LogLevelHelper.LogLevelStringArray;
```
paulhazen: I don't think these should be stored as member variables - they are only ever needed once in the lifecycle of the window, so you should be able to query them when you need them.

Assets/Plugins/Source/Editor/Windows/LogLevelWindow.cs
```diff
@@ -0,0 +1,118 @@
...
+            if (currentLogLevelConfig.logCategoryLevelPairs.Count != categories.Length)
+            {
+                currentLogLevelConfig.logCategoryLevelPairs.Clear();
+
+                foreach (var category in categories)
+                {
+                    currentLogLevelConfig.logCategoryLevelPairs.Add(new LogCategoryLevelPair(category, "Info"));
+                }
+
+                SaveToJSONConfig(true);
```
paulhazen: Why are you saving the JSON inside the `Setup` function?

Assets/Plugins/Source/Editor/Windows/LogLevelWindow.cs
```diff
@@ -0,0 +1,118 @@
...
+            if (GUILayout.Button("Save All Changes"))
+            {
+                SaveToJSONConfig(true);
+            }
+        }
+
+        private void SaveToJSONConfig(bool prettyPrint)
+        {
+            currentLogLevelConfig.Write(prettyPrint);
+        }
```
paulhazen: It feels like this can be simplified. You can remove the function `SaveToJSONConfig` (functions with that name have been removed from all the other windows that used to utilize it). Subsequent to removing that function, you can replace line `108` with simply `currentLogLevelConfig.Write()` (the `prettyPrint` parameter defaults to `true` if I recall correctly).

Assets/StreamingAssets/EOS/log_level_config.json
paulhazen, line 1: Can we add some documentation that makes it clear what the delineation is between each of these categories? It feels like it may be prone to arbitrary assignment. Is it possible that instead we might want to categorize the logs based on the scene?

Assets/Plugins/Source/Core/LogLevelConfig.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        public LogCategoryLevelPair(string category, string level)
+        {
+            this.category = category;
+            this.level = level;
+        }
+    }
+
+    [Serializable]
+    public class LogLevelConfig : Config
```
paulhazen: Because these two classes are only ever utilized by the `LogLevelHelper` class - I suggest that they be made into classes that are nested with in. Furthermore, if I'm not mistaken the first class can also be made to be nested within the second one.

Assets/Plugins/Source/Core/LogLevelHelper.cs
```diff
@@ -0,0 +1,68 @@
...
+using System;
+using System.IO;
+using System.Collections.Generic;
+using UnityEngine;
+using Epic.OnlineServices.Logging;
+using JsonUtility = PlayEveryWare.EpicOnlineServices.Utility.JsonUtility;
+
+namespace PlayEveryWare.EpicOnlineServices
+{
+    public static class LogLevelHelper
```
paulhazen: For most classes of this ilk, we have been using the nomenclature of "Utility," in favor of "Helper." To maintain consistency within the project I recommend this particular class be renamed `LogLevelUtility`, and be moved into the `Source/Core/Utility` directory.

paulhazen (changes requested):
Please see comments for details of my requested changes. Additionally, I believe that we need to put this PR through the prospectus process.

Assets/Plugins/Source/Editor/Windows/LogLevelWindow.cs
arthur740212: This is under the if statement when the list of categories has changed (for scenarios like a new interface from SDK),
and it regenerates a new config file in such cases.

Assets/StreamingAssets/EOS/log_level_config.json
arthur740212: These categories are all grabbed directly from EOSSDK log category enums, I'm not sure what you mean.

paulhazen: Ah - I'm sorry I misunderstood - I thought that _we_ were defining these categories. I'll mark this as resolved.

Assets/Plugins/Source/Editor/Windows/LogLevelWindow.cs
arthur740212: It is used a couple of times in the script.
I think it is cleaner to shorthand these.

paulhazen: Fair enough :)

Assets/Plugins/Source/Editor/Windows/LogLevelWindow.cs
paulhazen: I see - so basically - when the Window is opened, it loads the current log level configuration - then checks to see if there are any log levels that aren't included in it - and if there are, it updates them accordingly and saves the JSON. 

Am I understanding that correctly?

Assets/Plugins/Source/Core/EOSManager.cs
```diff
@@ -584,19 +586,22 @@ public void Init(IEOSCoroutineOwner coroutineOwner, string configFileName)
                         LoggingInterface.SetCallback(SimplePrintCallback);
                         hasSetLoggingCallback = true;
                     }
-#if UNITY_EDITOR
-                    SetLogLevel(LogCategory.AllCategories, LogLevel.VeryVerbose);
-#else
-                    SetLogLevel(LogCategory.AllCategories, LogLevel.Warning);
-#endif
+
+                    for (int logCategoryIndex = 0; logCategoryIndex < logLevelList.Count; logCategoryIndex++)
```
WispyMouse: I feel like since this shows up in three places, we should make a private function that its purpose is to do this code.

Assets/Plugins/Source/Core/EOSManager.cs
```diff
@@ -584,19 +586,22 @@ public void Init(IEOSCoroutineOwner coroutineOwner, string configFileName)
...
+                    {
+                        SetLogLevel((LogCategory)logCategoryIndex, logLevelList[logCategoryIndex]);
+                    }
 
                     InitializeOverlay(coroutineOwner);
                     return;
                 }
 
 #if !UNITY_EDITOR && !UNITY_SWITCH
                 // Set logging to VeryVerbose on EOS SDK bootstrap so we get the most logging information
```
WispyMouse: Comment can be nixed, no longer true

Assets/Plugins/Source/Core/LogLevelConfig.cs
WispyMouse: This suggestion makes sense to me. It's really limited in its use cases, and would add fewer things to the namespace for people to be concerned with.

Assets/Plugins/Source/Core/LogLevelConfig.cs
```diff
@@ -0,0 +1,48 @@
...
+
+namespace PlayEveryWare.EpicOnlineServices
+{
+    [Serializable]
+    public class LogCategoryLevelPair
+    {
+        public string category;
+        public string level;
+
+        public LogCategoryLevelPair(string category, string level)
```
WispyMouse, line 34: My instincts tell me that if a class is Serializable, it needs to have a parameterless constructor if it's going to be serialized/deserialized. That may or may not be true here, but I'd like to just check in that this is deserializing properly.

Assets/Plugins/Source/Core/LogLevelConfig.cs
WispyMouse, line 42: This needs to have a static constructor that registers it to the factory. When trying to use this I get;

> No factory method has been registered for type "PlayEveryWare.EpicOnlineServices.LogLevelConfig". Please make sure that "PlayEveryWare.EpicOnlineServices.LogLevelConfig" registers its constructor with the base Config class via a static constructor.

Assets/Plugins/Source/Editor/Windows/LogLevelWindow.cs
arthur740212: Partially. It creates a new config and set them all to LogLevel.Info.
The usage of this is pretty rare, since EOSSDK wouldn't dish out new categories that often.

Assets/Plugins/Source/Core/LogLevelConfig.cs
arthur740212: The reason why they were separated is because the Helper class interacts with the SDK, while the Config only handles pure text/json without knowing the SDK exists

Assets/Plugins/Source/Core/LogLevelConfig.cs
arthur740212: The serialize attribute is for the two member variables to work with json. 
The constructor is only called from the window and probably has nothing to do with json.

Update : I think you are worried if this will break in the inspector?
We don't use built in inspector features for this config, so we don't need to worry about that part.

Assets/Plugins/Source/Editor/Windows/LogLevelWindow.cs
```diff
@@ -0,0 +1,114 @@
...
+        [MenuItem("Tools/EOS Plugin/Log Level")]
+        public static void ShowWindow()
+        {
+            GetWindow<LogLevelWindow>("Log Level");
+        }
+        protected override void Setup()
+        {
+            currentLogLevelConfig = EpicOnlineServices.Config.Get<LogLevelConfig>();
+
+            // Config.Get<T> only guarantees T to be not null, but not for its members
```
paulhazen, line 51: Excellent observation :thinking: might have to consider this fact if ever there are future adjustments.

Assets/Plugins/Source/Core/LogLevelConfig.cs
paulhazen: Your observation that the Utility class interacts with the SDK while the Config only handles pure text/json is an astute one. I think your reasoning is salient. However - I would argue that making these classes nested does not actually make them interact with the SDK at all - it just limits the scope of where those classes can be utilized.

This is _very_, _very_ pedantic of me (and I take no issue with your reasoning here) but thought I would give voice to my thought-process in case it facilitates more thought on the topic in future circumstances that may be different.

Assets/Plugins/Source/Core/EOSManager.cs
```diff
@@ -769,6 +759,20 @@ public void SetLogLevel(LogCategory Category, LogLevel Level)
                 }
             }
 
+            //-------------------------------------------------------------------------
+            /// <summary>
+            /// Initialize log levels loaded from <see cref="LogLevelConfig" />
```
paulhazen: +100 for using cool `cref` comment linkage. Makes me want to up my own commenting game 😃

paulhazen (changes requested):
The most recent updates to this branch are baller! Kudos for continuing to grind on this, it's great work, and I hope you're okay with us all putting it through the proverbial wringer :)

My only remaining concern is that I do not see entries in `eos_package_description.json` for the files that have been added. If I'm not mistaken - this functionality would be good to have included in the package when imported from UPM.

Assets/Plugins/Source/Core/EOSManager.cs
```diff
@@ -673,13 +669,7 @@ public void Init(IEOSCoroutineOwner coroutineOwner, string configFileName)
 
 
                 InitializeOverlay(coroutineOwner);
-
-                // Default back to quiet logs
-#if UNITY_EDITOR
-                SetLogLevel(LogCategory.AllCategories, LogLevel.VeryVerbose);
```
andrew-hirata-playeveryware, line 679: I think in editor we should still use 'very verbose'

andrew-hirata-playeveryware (changes requested):

Assets/Plugins/Source/Core/EOSManager.cs
```diff
@@ -1772,8 +1771,8 @@ private static void UpdateNetworkStatus()
         }
 #endif
 
-        /// <value>Private static instance of <c>EOSSingleton</c></value>
-        static EOSSingleton s_instance;
+                /// <value>Private static instance of <c>EOSSingleton</c></value>
+                static EOSSingleton s_instance;
```
andrew-hirata-playeveryware: Remove the extra tabs, if you could.

Assets/Plugins/Source/Core/EOSManager.cs
```diff
@@ -584,23 +584,23 @@ public void Init(IEOSCoroutineOwner coroutineOwner, string configFileName)
...
-#if !UNITY_EDITOR && !UNITY_SWITCH
-                InitializeLogLevels();
-#endif
                 s_state = EOSState.Starting;
 
                 LoadEOSLibraries();
 
+                // Set log level prior to platform interface initialization
+                // VeryVerbose for dynamic linking platforms, otherwise set levels from configs 
+#if UNITY_EDITOR || UNITY_SWITCH
```
andrew-hirata-playeveryware: I believe you can remove the extra UNITY_SWITCH define.

Assets/Plugins/Source/Editor/Windows/LogLevelWindow.cs
```diff
@@ -0,0 +1,114 @@
...
+
+        int selectedCategoryIndex = 0;
+        bool showAllCategories = false;
+        string previousAllCategoryLogLevel;
+
+        [MenuItem("Tools/EOS Plugin/Log Level")]
+        public static void ShowWindow()
+        {
+            GetWindow<LogLevelWindow>("Log Level");
+        }
```
paulhazen: For consistency, I think we should rename the window and the corresponding MenuItem `Tools/EOS Plugin/Log Level Configuration`.

Assets/Plugins/Source/Editor/Windows/LogLevelWindow.cs
```diff
@@ -0,0 +1,114 @@
...
+            currentLogLevelConfig = EpicOnlineServices.Config.Get<LogLevelConfig>();
+
+            // Config.Get<T> only guarantees T to be not null, but not for its members
+            if (currentLogLevelConfig.logCategoryLevelPairs == null) 
+            {
+                currentLogLevelConfig.logCategoryLevelPairs = new List<LogCategoryLevelPair>();
+            }
+
+            // Initialize the list if config categories does not match the SDK categories, which might happen when using a different SDK version
+            if (currentLogLevelConfig.logCategoryLevelPairs.Count != categories.Length)
```
paulhazen: This will not catch circumstances where a category is renamed, or if between one SDK and another there is one log category removed and one log category added.

Assets/Plugins/Source/Editor/Windows/LogLevelWindow.cs
```diff
@@ -0,0 +1,114 @@
...
+                    currentLogLevelConfig.logCategoryLevelPairs.Add(new LogCategoryLevelPair(category, "Info"));
+                }
+
+                currentLogLevelConfig.Write(true);
+            }
+        }
+        protected override void RenderWindow()
+        {
+            // Last element in the list is All Categories
+            previousAllCategoryLogLevel = currentLogLevelConfig.logCategoryLevelPairs[categories.Length - 1].level;
```
paulhazen: I think this line should be moved to be between line 97 and line 98. More importantly, it should be a local variable - there is no need to store it as a member of the `LogLevelWindow` class, as it is only ever used in this circumstance - and its value does not need to persist between calls of `RenderWindow`.

Assets/Plugins/Source/Core/LogLevelConfig.cs
```diff
@@ -0,0 +1,52 @@
...
+using System;
+using System.Collections.Generic;
+
+namespace PlayEveryWare.EpicOnlineServices
+{
+    [Serializable]
+    public class LogCategoryLevelPair
+    {
+        public string category;
+        public string level;
```
paulhazen: These should be capitalized.

Assets/Plugins/Source/Core/LogLevelConfig.cs
```diff
@@ -0,0 +1,52 @@
...
+    [Serializable]
+    public class LogLevelConfig : Config
+    {
+        static LogLevelConfig()
+        {
+            RegisterFactory(() => new LogLevelConfig());
+        }
+        public LogLevelConfig() : base("log_level_config.json") { }
+
+        public List<LogCategoryLevelPair> logCategoryLevelPairs;
```
paulhazen: This should also be capitalized.

Assets/Plugins/Source/Core/Utility/LogLevelUtility.cs
```diff
@@ -0,0 +1,68 @@
...
+        public static string[] LogLevelStringArray
+        {
+            get { return Enum.GetNames(typeof(LogLevel)); }
+        }
+
+        public static List<LogLevel> LogLevelList
+        {
+            get
+            {
+                LogLevelConfig logLevelConfig = JsonUtility.FromJsonFile<LogLevelConfig>(Path.Combine(Application.streamingAssetsPath, "EOS", "log_level_config.json"));
```
paulhazen: This should be changed to just: `Config.Get<LogLevelConfig>()`

Assets/Plugins/Source/Core/Utility/LogLevelUtility.cs
```diff
@@ -0,0 +1,68 @@
...
+
+                var logLevels = new List<LogLevel>();
```
paulhazen: Instead of having the ambiguity at the beginning of the variable declaration in the form of the `var` keyword, a better approach would be to utilize the `new()` constructor like this:

```csharp
List<LogLevel> logLevels = new()
```

Assets/Plugins/Source/Editor/Windows/LogLevelWindow.cs
```diff
@@ -0,0 +1,114 @@
...
+using UnityEditor;
+using System;
+using System.Collections.Generic;
+
+#if !EOS_DISABLE
+namespace PlayEveryWare.EpicOnlineServices.Editor.Windows
+{
+    public class LogLevelWindow : EOSEditorWindow
+    {
+        LogLevelConfig currentLogLevelConfig;
```
paulhazen, line 34: There should be no need to store a copy of the log level config. If the log levels change while this is running, there is a chance that the wrong values could be subsequently saved. 

It is safer to use: 
```csharp
Config.Get<LogLevelConfig>();
```

paulhazen (changes requested):

andrew-hirata-playeveryware (approved):

Assets/Plugins/Source/Editor/Windows/LogLevelWindow.cs
arthur740212: That is true, but will be ultra rare, since it only happens with a change from the SDK.
The category names are only used when creating the json, and does not check the names when reading,
so rename won't be a problem.
removing and adding a category is even more rare, but in those case when it happens, the categories and levels are still still represented correctly in the UI.

The purpose of this is to make sure there are not any out of range exceptions

paulhazen: That's a good point regarding the intent of the check. It does however reveal that the aforementioned circumstance can happen and is not accounted for by the current implementation. If this circumstance _did_ happen (and I completely agree it would be extraordinarily rare) I can see it causing a fair amount of confusion and subsequent mayhem.

Assets/Plugins/Source/Editor/Windows/LogLevelWindow.cs
arthur740212: You are right, it should be local.
But it cannot be moved, because the level change should be between previous and current to check if it is dirty.

Assets/Plugins/Source/Core/Utility/LogLevelUtility.cs
arthur740212: Is the Android IO removed?
This was made like this because of that

arthur740212: The IO fix will be on another branch anyways, so yeah, will change this.

Assets/Plugins/Source/Editor/Windows/LogLevelWindow.cs
arthur740212: If it is not stored somewhere, then it will need to reload with `Config.Get` every frame the window renders, which will make the config immutable.

paulhazen (approved):

PR #672, opened 2024-05-14
fix: Assign value to `out` parameter when not in editor mode.

paulhazen:
This fixes an as-yet unreported error that takes place when not in editor mode and the `TryFromJson` from `JsonUtility` function is called. 

The reason is that when not in editor mode, no exception is thrown and `false` is returned - but no value is assigned to the `out` parameter.


arthur740212 (approved):

WispyMouse (approved):
Using a dotnetfiddle, I can see that this would solve that problem. The code compiles now and Unity seems happy. I learned something today about out variables!

andrew-hirata-playeveryware (approved):
Ah! that's a good find!

PR #673, opened 2024-05-14
Fix: Cross-Platform Building (Currently broken)

### LINKED ISSUE #651: Fail to build for Linux Server
We use Linux builds on the server. No EOS functionality is used there, all calls are hidden under #if. However, having your plugin in itself does not allow us to build - we get the following error:

```
2024-05-06T19:42:46.0393390Z ##### Output
2024-05-06T19:42:46.0394337Z ld.lld: error: undefined symbol: DLLH_create_context
2024-05-06T19:42:46.0396707Z >>> referenced by com.playeveryware.eos.core.cpp:3064 (Library/Bee/artifacts/LinuxPlayerBuildProgram/il2cppOutput/cpp/com.playeveryware.eos.core.cpp:3064)
2024-05-06T19:42:46.0399972Z >>>               Library/Bee/artifacts/LinuxPlayerBuildProgram/zbz95/u8xxrn7ehl7g.o:(DLLHandle_LoadDynamicLibrary_m4D283A506980785C68E30AE7A6117C271816B489)
2024-05-06T19:42:46.0403107Z >>> referenced by com.playeveryware.eos.core.cpp:3064 (Library/Bee/artifacts/LinuxPlayerBuildProgram/il2cppOutput/cpp/com.playeveryware.eos.core.cpp:3064)
2024-05-06T19:42:46.0406149Z >>>               Library/Bee/artifacts/LinuxPlayerBuildProgram/zbz95/u8xxrn7ehl7g.o:(SystemDynamicLibrary_get_Instance_mE5272AED736BB77ABE3F7BA88CC63674A5BAC28D)
```

The workaround that helped us was to turn off the dlls from compilation like this, but to do this we had to disable plugin sync via Git.
image

### LINKED ISSUE #652: Mac Unity Editor - compiler error due to symbol duplication (due to compiler directives)
**Describe the bug**
Error `The namespace 'PlayEveryWare.EpicOnlineServices' already contains a definition for 'EOSCreateOptions'`
due to incorrect compiler directives in
`Runtime/macOS/Core/MacOSPlatformSpecifics.cs`
and
`Runtime/Windows/Core/WindowsPlatformSpecifics.cs`

**To Reproduce**
1. Create empty project on Mac
2. Select Windows Intel 64 bit as target platform
3. Add EOS package of version 3.1.3

**Desktop (please complete the following information):**
 - mac OS Intel 14.1.1 (23B81)
 - Editor
 - Unity Version 2021.3.14f1
 - Plugin Version 3.1.3

**Additional context**
Runtime/macOS/Core/MacOSPlatformSpecifics.cs has condition:
`#if (UNITY_STANDALONE_OSX || UNITY_EDITOR_OSX) && !UNITY_EDITOR_WIN`
Runtime/Windows/Core/WindowsPlatformSpecifics.cs has condition:
`#if (UNITY_STANDALONE_WIN || UNITY_EDITOR_WIN || UNITY_WSA_10_0)`
Both conditions are true in Mac editor, and duplicate symbol error occurs
(On Windows only the condition in WindowsPlatformSpecifics.cs is true)

Temp fix - change condition to
`#if (UNITY_STANDALONE_OSX) && !UNITY_EDITOR_WIN`
in MacOSPlatformSpecifics.cs

paulhazen:
# Cross-Platform Build Fixes

Contrary to how it may appear upon first glance - this is not a refactor but a certified fix. In the process of hunting down a bug that a user discovered with the failure to build on the Mac I discovered that none of the cross-platform compilations were working correctly. Mac works if you build on mac, Windows works if you build on Windows, Linux with Linux.

There are two main areas I discovered that needed to be addressed to fix this - and I'm being as verbose both in this description and by adding comments to some of the files to illustrate clearly the changes that were made.
 
## Scripting Defines

There are currently a variety of bugs that exist because of the complicated nature of the scripting define conditionals at the top of some of the classes that derive from `PlatformSpecifics`. 

This PR addresses those issues and resolves a variety of open issues.

Most of these circumstances are pertaining to building on one platform to another platform. Because most developers who work on this plugin do so from the Windows platform, some of the configurations are currently in a broken state because of how the define conditionals work.

Below is a table of what is now supported (which is what documentation already _says_ is supported)

|Host Platform|Target Platform|
|-:|:-|
|Mac|Windows, Linux, Android, iOS|
|Windows|Mac, Linux, Android, iOS|
|Linux|Windows, Mac, Android, iOS|

_To be clear: you can't actually fully deploy an iOS app from anything but Mac, same as with targeting Mac - but "build" in this context means that Unity will do all the building it is capable of doing for that platform given the platform you are building from._

## `BuildRunner` Builder Selection

Previously, build runners were determined using the scripting defines for each platform. The problem with this approach is that it inherently _does not correctly build from one platform to another_.

The reason that this has gone unnoticed for a while is that many of the platform-specific tasks that are unique to each platform are minor - so the build runner for OSX is _close enough_ to the build runner for Windows that it doesn't cause any build errors. 

There are some circumstances where errors are reported - but more importantly these changes both avoid those errors - _and correctly implement cross-platform compilation_.

### DIFF
--- a/Assets/Plugins/Android/Core/AndroidPlatformSpecifics.cs
+++ b/Assets/Plugins/Android/Core/AndroidPlatformSpecifics.cs
@@ -30,8 +30,7 @@ using jsize = System.Int32;
 using JavaVM = System.IntPtr;
 using System.Diagnostics;
 
-#if UNITY_ANDROID && !UNITY_EDITOR
-[assembly: AlwaysLinkAssembly]
+#if UNITY_ANDROID
 namespace PlayEveryWare.EpicOnlineServices
 {
     using Epic.OnlineServices.Platform;
--- a/Assets/Plugins/Android/Core/AndroidPlatformSpecifics.cs
+++ b/Assets/Plugins/Android/Core/AndroidPlatformSpecifics.cs
@@ -20,6 +20,8 @@
 * SOFTWARE.
 */
 
+#if !EOS_DISABLE
+
 using UnityEngine;
 using UnityEngine.Scripting;
 using System.Runtime.InteropServices;
@@ -113,4 +115,5 @@ namespace PlayEveryWare.EpicOnlineServices
         }
     }
 }
+#endif
 #endif
--- a/Assets/Plugins/Android/Core/AndroidPlatformSpecifics.cs
+++ b/Assets/Plugins/Android/Core/AndroidPlatformSpecifics.cs
@@ -30,7 +30,8 @@ using jsize = System.Int32;
 using JavaVM = System.IntPtr;
 using System.Diagnostics;
 
-#if UNITY_ANDROID
+#if UNITY_ANDROID && !UNITY_EDITOR
+
 namespace PlayEveryWare.EpicOnlineServices
 {
     using Epic.OnlineServices.Platform;
--- a/Assets/Plugins/Android/Editor/AndroidBuilder.cs
+++ b/Assets/Plugins/Android/Editor/AndroidBuilder.cs
@@ -36,7 +36,7 @@ namespace PlayEveryWare.EpicOnlineServices.Build
 
     public class AndroidBuilder : PlatformSpecificBuilder
     {
-        public AndroidBuilder() : base("Plugins/Android") { }
+        public AndroidBuilder() : base("Plugins/Android", BuildTarget.Android) { }
 
         public override void PreBuild(BuildReport report)
         {
--- a/Assets/Plugins/Linux/Core/LinuxPlatformSpecifics.cs
+++ b/Assets/Plugins/Linux/Core/LinuxPlatformSpecifics.cs
@@ -20,9 +20,7 @@
 * SOFTWARE.
 */
 
-#if UNITY_EDITOR
-#define EOS_DYNAMIC_BINDINGS
-#endif
+#if !EOS_DISABLE
 
 //#define ENABLE_CONFIGURE_STEAM_FROM_MANAGED
 using PlayEveryWare.EpicOnlineServices.Utility;
@@ -64,7 +62,7 @@ namespace PlayEveryWare.EpicOnlineServices
         //-------------------------------------------------------------------------
         public override void LoadDelegatesWithEOSBindingAPI()
         {
-#if EOS_DYNAMIC_BINDINGS
+#if UNITY_EDITOR
             // TODO: This code does not appear to do anything...
             const string EOSBinaryName = Epic.OnlineServices.Config.LibraryName;
             var eosLibraryHandle = EOSManager.EOSSingleton.LoadDynamicLibrary(EOSBinaryName);
@@ -128,4 +126,4 @@ namespace PlayEveryWare.EpicOnlineServices
     }
 }
 #endif
-
+#endif
--- a/Assets/Plugins/Linux/Core/LinuxPlatformSpecifics.cs
+++ b/Assets/Plugins/Linux/Core/LinuxPlatformSpecifics.cs
@@ -28,7 +28,7 @@ using UnityEngine;
 using UnityEngine.Scripting;
 using System.Runtime.InteropServices;
 
-#if !UNITY_EDITOR_WIN && (UNITY_STANDALONE_LINUX || UNITY_EDITOR_LINUX)
+#if (UNITY_STANDALONE_LINUX || UNITY_EDITOR_LINUX) && (!UNITY_EDITOR_WIN && !UNITY_EDITOR_OSX)
 
 namespace PlayEveryWare.EpicOnlineServices
 {
--- a/Assets/Plugins/Linux/Core/LinuxPlatformSpecifics.cs
+++ b/Assets/Plugins/Linux/Core/LinuxPlatformSpecifics.cs
@@ -28,7 +28,8 @@ using UnityEngine;
 using UnityEngine.Scripting;
 using System.Runtime.InteropServices;
 
-#if (UNITY_STANDALONE_LINUX || UNITY_EDITOR_LINUX) && (!UNITY_EDITOR_WIN && !UNITY_EDITOR_OSX)
+// If standalone linux and not editor, or the linux editor
+#if (UNITY_STANDALONE_LINUX && !UNITY_EDITOR) || UNITY_EDITOR_LINUX
 
 namespace PlayEveryWare.EpicOnlineServices
 {
--- a/Assets/Plugins/Linux/Core/LinuxPlatformSpecifics.cs
+++ b/Assets/Plugins/Linux/Core/LinuxPlatformSpecifics.cs
@@ -63,7 +63,7 @@ namespace PlayEveryWare.EpicOnlineServices
         //-------------------------------------------------------------------------
         public override void LoadDelegatesWithEOSBindingAPI()
         {
-#if UNITY_EDITOR
+#if EOS_DYNAMIC_BINDINGS || UNITY_EDITOR
             // TODO: This code does not appear to do anything...
             const string EOSBinaryName = Epic.OnlineServices.Config.LibraryName;
             var eosLibraryHandle = EOSManager.EOSSingleton.LoadDynamicLibrary(EOSBinaryName);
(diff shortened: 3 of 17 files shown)


Assets/Plugins/Android/Core/AndroidPlatformSpecifics.cs
```diff
@@ -31,7 +33,7 @@
 using System.Diagnostics;
 
 #if UNITY_ANDROID && !UNITY_EDITOR
-[assembly: AlwaysLinkAssembly]
+
```
paulhazen, line 36: Instead of applying this assembly attribute within a conditional within every specific file for every platform - its implementation was moved into the baser `PlatformSpecifics` abstract class where it only needs to be done once.

Assets/Plugins/Linux/Core/LinuxPlatformSpecifics.cs
```diff
@@ -20,27 +20,16 @@
...
 
-#if UNITY_64 || UNITY_EDITOR_64
-#define PLATFORM_64BITS
-#elif UNITY_EDITOR_LINUX || UNITY_STANDALONE_LINUX
-#define PLATFORM_32BITS
-#endif
-
-#if UNITY_EDITOR
-#define EOS_DYNAMIC_BINDINGS
-#endif
```
paulhazen, line 31: Because nearly identical changes to the scripting define conditionals were made in each file where they were altered, I will explain some of the detail here in the first file GitHub indicates is changed by whatever metric they use to determine such things.

- `#define` statements in a Unity C# project are scoped to the file.
- `UNITY_64` is implied by `UNITY_EDITOR_64`, so it is only necessary to check the former.
- `EOS_DYNAMIC_BINDINGS` is in effect an _alias_ for `UNITY_EDITOR`

Care was taken to ensure that no branch of any scripting define conditional is unreachable (no orphaned usages of removed defines).

Assets/Plugins/Linux/Core/LinuxPlatformSpecifics.cs
```diff
@@ -20,27 +20,16 @@
...
 using UnityEngine.Scripting;
 using System.Runtime.InteropServices;
 
-#if !UNITY_EDITOR_WIN && (UNITY_STANDALONE_LINUX || UNITY_EDITOR_LINUX)
-
-#if !UNITY_EDITOR_LINUX
-[assembly: AlwaysLinkAssembly]
-#endif
+// If standalone linux and not editor, or the linux editor
+#if (UNITY_STANDALONE_LINUX && !UNITY_EDITOR) || UNITY_EDITOR_LINUX
```
paulhazen, line 32: For every platform _for which there is a Unity editor_ we need to have that platform's `PlatformSpecifics`-derived class be defined. That means the specifics class must be defined if either the editor for that platform is active, or if the target is the platform without any editor.

Assets/Plugins/Android/Editor/AndroidBuilder.cs
```diff
@@ -36,7 +36,7 @@ namespace PlayEveryWare.EpicOnlineServices.Build
 
     public class AndroidBuilder : PlatformSpecificBuilder
     {
-        public AndroidBuilder() : base("Plugins/Android") { }
+        public AndroidBuilder() : base("Plugins/Android", BuildTarget.Android) { }
```
paulhazen, line 39: In order for the proper Builder to be determined based on the platform _being built to_, the various builder classes need to be able to indicate for which BuildTargets they work for.

*The decentralized nature of this implementation allows for the addition of other platforms __without__ altering anything in the existing build pipeline.*

Assets/Plugins/Android/Core/AndroidPlatformSpecifics.cs
andrew-hirata-playeveryware: That actually works? I was under the belief that it didn't work and or we ran into some kind of issue with it in the past.

Assets/Plugins/Linux/Editor/LinuxBuilder.cs
```diff
@@ -22,9 +22,13 @@
 
 namespace PlayEveryWare.EpicOnlineServices.Build
 {
+    using UnityEditor;
+
     public class LinuxBuilder : PlatformSpecificBuilder
     {
-        public LinuxBuilder() : base("Plugins/Linux")
+        public LinuxBuilder() : base(
```
andrew-hirata-playeveryware: Might I suggest moving these onto the same line as base?

Assets/Plugins/Source/Editor/Build/BuildRunner.cs
```diff
@@ -22,114 +22,55 @@
...
-    using System.Threading.Tasks;
-    using UnityEditor;
     using UnityEditor.Build;
     using UnityEditor.Build.Reporting;
-    using UnityEngine;
-    
-    public class BuildRunner : IPreprocessBuildWithReport, IPostprocessBuildWithReport
+
+    public class BuildRunner : IPreprocessBuildWithReport, 
+        IPostprocessBuildWithReport
```
andrew-hirata-playeveryware: This should also be on the same line as the `:`

Assets/Plugins/Windows/Core/WindowsPlatformSpecifics.cs
```diff
@@ -41,14 +32,12 @@
...
-
-#if !UNITY_EDITOR_WIN
-[assembly: AlwaysLinkAssembly]
-#endif
+// If standalone windows and not editor, or the windows editor.
+#if (UNITY_STANDALONE_WIN && !UNITY_EDITOR) || UNITY_EDITOR_WIN
 
 namespace PlayEveryWare.EpicOnlineServices
 {
+    using System.Runtime.CompilerServices;
```
andrew-hirata-playeveryware: I don't imagine this using is needed here.

andrew-hirata-playeveryware (changes requested):

Assets/Plugins/Linux/Editor/LinuxBuilder.cs
paulhazen: Donzo.

Assets/Plugins/Android/Core/AndroidPlatformSpecifics.cs
paulhazen: It does indeed work properly - I had to test the different build configurations running around my parents dining room between my windows laptop, my mac mini, and the linux vm I had running on one of them to make sure all the build configurations worked properly. 

I think typically this kind of thing is usually resolved with an `AssemblyInfo.cs` file, but Unity doesn't typically feature those.

Assets/Plugins/Source/Core/PlatformSpecifics.cs
```diff
@@ -21,15 +21,21 @@
...
+using UnityEngine.Scripting;
+[assembly: AlwaysLinkAssembly]
+#endif
 namespace PlayEveryWare.EpicOnlineServices
 {
     using System;
     using System.Collections.Generic;
     using System.IO;
     using UnityEngine;
+    using UnityEngine.Scripting;
```
WispyMouse, line 34: Is the twice-included UnityEngine.Scripting necessary?

Assets/Plugins/Android/Core/AndroidPlatformSpecifics.cs
arthur740212: Just to double check, these are tested with IL2CPP code stripping enabled?
That was the conditions when it was causing trouble

andrew-hirata-playeveryware (approved):

WispyMouse (approved):

Assets/Plugins/Android/Core/AndroidPlatformSpecifics.cs
paulhazen: Yes, it appears to be working in that circumstance as well.

Assets/Plugins/Source/Core/PlatformSpecifics.cs
paulhazen: It's not - but I'm not seeing the duplicate, is it out of scope from the preview here?

PR #678, opened 2024-05-16
fix: Correctly determine output buffer size in `ContinuanceToken`'s `ToString` method.

### LINKED ISSUE #674: ContinuanceToken ToString always fails, so they cannot be logged out or seen
The `ContinuanceToken` class, which is sometimes returned from login API calls, cannot be debug-logged out since its `ToString` functionality always fails with result `LimitExceeded`.

```c#
public Result ToString(out Utf8String outBuffer)
{
    int inOutBufferLength = 1024;
    System.IntPtr outBufferAddress = Helper.AddAllocation(inOutBufferLength);

    var funcResult = Bindings.EOS_ContinuanceToken_ToString(InnerHandle, outBufferAddress, ref inOutBufferLength);

    Helper.Get(outBufferAddress, out outBuffer);
    Helper.Dispose(ref outBufferAddress);

    return funcResult;
}
```

It seems this code only attempts `EOS_ContinuanceToken_ToString` with a buffer size of **1024** (which must be too small) and reports the result. Unfortunately, doesn't seem the innards are exposed in any way that would allow me to make the call myself to work around this.

According to a comment in the source, Epic suggest you can call `EOS_ContinuanceToken_ToString` and pass a null buffer and it will report the size required.

It's very useful to be able to log out and inspect the various ids and tokens coming and going, and in my case this discovery has come after a while spent assuming I was getting back 'no continuance token' from calls, since they were logging out as empty strings, when actually I _am_ getting valid continuance tokens.

paulhazen:
For more context, see issue #674 (which this PR closes).


Assets/Plugins/Source/EOS_SDK/Generated/ContinuanceToken.cs
```diff
@@ -38,6 +38,18 @@ public Result ToString(out Utf8String outBuffer)
...
 
+			// NOTES:
+			// 
+			// The changes below are implemented temporarily by PEW to correct
+			// an issue where the inOutBufferLength is not properly determined.
+			// 
+			// Until this is resolved from Epic - these changes will need to
+			// be maintained after the import of a new version of the EOS SDK.
+
+			// Call the EOS_ContinuanceToken_ToString function will a null buffer address to determine the correct buffer length.
```
WispyMouse, line 50: nit: "Call the EOS_ContinuanceToken_ToString function _with_ a null buffer address..." instead of "_will_".

Assets/Plugins/Source/EOS_SDK/Generated/ContinuanceToken.cs
```diff
@@ -38,6 +38,18 @@ public Result ToString(out Utf8String outBuffer)
 			int inOutBufferLength = 1024;
 			System.IntPtr outBufferAddress = Helper.AddAllocation(inOutBufferLength);
 
+			// NOTES:
```
WispyMouse, line 42: Is it within our policies to be allowed to link to GitHub issues within the code? This feels like a place where that context would be helpful. Maybe a date on when this was introduced?

paulhazen: That's a good idea - I'll include that when I re-open the PR.

paulhazen:
I'm converting to a draft until I am actually able to verify that the test works myself.

Assets/Plugins/Source/EOS_SDK/Generated/ContinuanceToken.cs
andrew-hirata-playeveryware: We have a notation for marking a 'PEW' specific modifications to the code:
```
//PEW: start modify
.....
//PEW: end modify
```

This is done to make it easy to kind of `grep` though the code to find PEW edits to the SDK without having to dig though git too much. I will lament that this notation isn't documented anywhere.

andrew-hirata-playeveryware (changes requested):
Tiny change to add a `//PEW: start modify` block around the changes to the SDK.

arthur740212 (approved):
Looks like there is only comments?

PR #683, opened 2024-05-20
Release 3.2.0

paulhazen:


Assets/Plugins/Source/EOS_SDK/Core/Config.cs
```diff
@@ -46,9 +46,8 @@
 
 namespace Epic.OnlineServices
 {
-	public static partial class Config
+	public static class Config
```
paulhazen: To theoretically support future platforms, we need to have this be `partial`.

Assets/Plugins/Source/EOS_SDK/Core/Config.cs
```diff
@@ -46,9 +46,8 @@
...
 	{
-#if EOS_PLATFORM_WINDOWS_32 || EOS_PLATFORM_WINDOWS_64 || EOS_PLATFORM_OSX || EOS_PLATFORM_LINUX || EOS_PLATFORM_IOS || EOS_PLATFORM_ANDROID
```
paulhazen, line 51: This should probably stay put.

etc/PlatformSpecificAssets/EOS/Linux/EasyAntiCheat/SplashScreen.png
paulhazen, line 1: We should switch to using the EOS SDK Unity Plugin splash screen instead of the one provided by default.

lib/NativeCode/DynamicLibraryLoaderHelper/DynamicLibraryLoaderHelper/DynamicLibraryLoaderHelper.vcxproj
paulhazen, line 1: These should be returned to version `17763`, right? Find what the differences are between the target platform versions and be certain that doing so does not drop support for any platform.

lib/NativeCode/DynamicLibraryLoaderHelper/DynamicLibraryLoaderHelper/DynamicLibraryLoaderHelper.vcxproj
```diff
@@ -23,32 +23,32 @@
...
     <RootNamespace>DynamicLibraryLoaderHelper</RootNamespace>
-    <WindowsTargetPlatformVersion>10.0.17763.0</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformVersion>10.0.22621.0</WindowsTargetPlatformVersion>
   </PropertyGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'" Label="Configuration">
     <ConfigurationType>StaticLibrary</ConfigurationType>
     <UseDebugLibraries>true</UseDebugLibraries>
-    <PlatformToolset>v141</PlatformToolset>
+    <PlatformToolset>v143</PlatformToolset>
```
paulhazen: Similar question to above - do we really want to upgrade / update this?

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/NativeRender.vcxproj
paulhazen, line 1: Same concerns about this vcxproj changes as the previous one.

Assets/Plugins/Source/EOS_SDK/Generated/Types/IntegratedPlatform.cs
paulhazen, line 1: Make sure that this directory is included in the package description json file.

.gitattributes
```diff
@@ -35,3 +31,4 @@
 *.asset             merge=unityyamlmerge eol=lf
 *.meta              merge=unityyamlmerge eol=lf
 *.controller        merge=unityyamlmerge eol=lf
+Assets/Plugins/Linux/libDynamicLibraryLoaderHelper.so filter=lfs diff=lfs merge=lfs -text
```
WispyMouse: This looks like a binary file. Should it be -text?

Assets/Plugins/Source/EOS_SDK/Core/ConvertHelper.cs
```diff
@@ -71,7 +71,9 @@ private static void Convert(string from, out byte[] to, int fromLength)
 				from = "";
 			}
 
-			to = Encoding.ASCII.GetBytes(new string(from.Take(fromLength).ToArray()).PadRight(fromLength, '\0'));
+			to = new byte[fromLength];
+			ASCIIEncoding.ASCII.GetBytes(from, 0, from.Length, to, 0);
+			to[from.Length] = 0;  //  Null terminator at the end'\0'
```
WispyMouse, line 76: I'm somewhat surprised reading this.

We have a length `fromLength`. We make a byte array that is of size `fromLength`. Then, we set a variable at the index of `from.Length`. `from` is a string. If `from` has 10 characters in it, and `fromLength` is 10, this'd result in trying to put a 0 at index 10, which would be an out of bounds array.

Is this somehow set up so that every provided `from` is smaller than `fromLength`?

This looks like it would be different from `PadRight`, which would (I guess) put 0 in all empty array indexes to the right.

Assets/Plugins/Source/EOS_SDK/Core/Utf8String.cs
```diff
@@ -11,6 +11,8 @@ namespace Epic.OnlineServices
 	[System.Diagnostics.DebuggerDisplay("{ToString()}")]
 	public sealed class Utf8String
 	{
+		static public Utf8String EmptyString = new Utf8String();
```
WispyMouse, line 14: Can we lock this down to be readonly, or have only a get-accessor?

I've had a project before where "Vector3.zero" was set in one class and re-used, and it took a while to debug a problem when it was realized that the reference of the returned value was being modified rather than using it by value.

paulhazen (commented):
Please see attached notes.

Assets/Plugins/Source/EOS_SDK/Core/Utf8String.cs
paulhazen: I'm going to recommend against us doing this - not because I disagree with the strategy - I _do_ think it is a good one: but because this is code within the EOS SDK itself. 

I don't want to put us in a position where we are having to maintain changes that need reimplementing when we change SDK versions and import an updated copy of this file from Epic.

.gitattributes
paulhazen: No, good catch @WispyMouse.

lib/NativeCode/DynamicLibraryLoaderHelper/DynamicLibraryLoaderHelper/DynamicLibraryLoaderHelper.vcxproj
paulhazen: @andrew-hirata-playeveryware to close the loop on this conversation: The SDK version `22621` supports the same set of platforms as `17763`, so there is not a compatibility issue here - and keeping this as up to date as possible is a good approach, so I am going to leave these updated.

paulhazen: Actually - for the purposes of this release - I am going to change them back to what they were - we can address changing the supported version at a future point in time.

lib/NativeCode/DynamicLibraryLoaderHelper/DynamicLibraryLoaderHelper/DynamicLibraryLoaderHelper.vcxproj
paulhazen: Values were restored to `v141`. Updates to the value will be considered in a change made subsequent to this release.

Assets/Plugins/Source/EOS_SDK/Core/ConvertHelper.cs
paulhazen: According to the [documentation](https://learn.microsoft.com/en-us/dotnet/api/system.text.asciiencoding.getbytes?view=net-8.0#system-text-asciiencoding-getbytes(system-char()-system-int32-system-int32-system-byte()-system-int32)) for `ASCIIEncoding.ASCII.GetBytes`, the parameter for which the value provided is `from.Length` is not an index, but a character count - so it's not out of bounds. 

Good attention to detail though - I had to look this up to confirm and learned some things along the way.

PR #684, opened 2024-05-20
fix(leaderboard sample): Refresh leaderboards after stat ingestion and after wait

WispyMouse:
EOS-1091; The leaderboards sample had a usability issue where the "Refresh" button unintuitively doesn't refresh the leaderboard rankings. This now refreshes those on press.

After using the UI to "ingest a stat", there is a configurable wait before requerying the leaderboard. This will help with reflecting changes to the leaderboard in a way that feels more responsive. Additionally there is a variable for waiting a number of seconds and requerying after the wait.

### DIFF
--- a/Assets/Scripts/UI/Leaderboard/UILeaderboardMenu.cs
+++ b/Assets/Scripts/UI/Leaderboard/UILeaderboardMenu.cs
@@ -127,7 +127,10 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
                 Debug.LogErrorFormat("UILeaderboardMenu (QueryRanksCompleted): returned result error: {0}", result);
 
                 // Even if we failed, try to refresh the leaderboard again later
-                if (refreshLeaderboardCoroutine != null) this.StopCoroutine(refreshLeaderboardCoroutine);
+                if (refreshLeaderboardCoroutine != null)
+                {
+                    this.StopCoroutine(refreshLeaderboardCoroutine);
+                }
                 refreshLeaderboardCoroutine = this.StartCoroutine(RefreshCurrentLeaderboardAfterWait(this.SecondsBetweenLeaderboardRefreshes));
 
                 return;
@@ -162,7 +165,10 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
                     }
                 }
 
-                if (refreshLeaderboardCoroutine != null) this.StopCoroutine(refreshLeaderboardCoroutine);
+                if (refreshLeaderboardCoroutine != null)
+                {
+                    this.StopCoroutine(refreshLeaderboardCoroutine);
+                }
                 refreshLeaderboardCoroutine = this.StartCoroutine(RefreshCurrentLeaderboardAfterWait(this.SecondsBetweenLeaderboardRefreshes));
             }
         }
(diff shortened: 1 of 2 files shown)


Assets/Scripts/UI/Leaderboard/UILeaderboardMenu.cs
```diff
@@ -102,6 +125,11 @@ private void QueryRanksCompleted(Result result)
             if(result != Result.Success)
             {
                 Debug.LogErrorFormat("UILeaderboardMenu (QueryRanksCompleted): returned result error: {0}", result);
+
+                // Even if we failed, try to refresh the leaderboard again later
+                if (refreshLeaderboardCoroutine != null) this.StopCoroutine(refreshLeaderboardCoroutine);
```
paulhazen: With the exception of `continue` statements within loops, I believe our convention for conditionals is to always put them on new-lines with curly brackets on lines of their own.

Assets/Scripts/UI/Leaderboard/UILeaderboardMenu.cs
```diff
@@ -335,5 +378,47 @@ private void SetCurrentLeaderboardDescription()
             CurrentSelectedLeaderboardTxt.text = $"{currentSelectedDefinitionLeaderboardId} - " +
                 $"{currentSelectedDefinitionStatName} - {Enum.GetName(typeof(LeaderboardGroup), currentGroup)}";
         }
+
+        /// <summary>
+        /// If a leaderboard is selected, re-query and re-display its values.
+        /// </summary>
+        public void RefreshCurrentSelectedLeaderboard()
```
paulhazen, line 400: I tend to think that a clearer way to organize this would be to have a `static` method that accepts a leader board as a parameter. This works perfectly fine, but the flexibility of the `static` implementation expands the ways it can be leveraged. Not sure whether that is a correct assessment or whether it is more subjective. Not something I think needs changing perse - just musing.

Assets/Scripts/UI/Leaderboard/UILeaderboardMenu.cs
```diff
@@ -63,6 +64,28 @@ private enum LeaderboardGroup
         private EOSLeaderboardManager LeaderboardManager;
         private EOSFriendsManager PlayerManager;
 
+        /// <summary>
+        /// If this value is greater than 0, then every time this number of seconds pass, the current viewed leaderboard will refresh.
+        /// The internal timer for this resets whenever a leaderboard is queried.
+        /// There is no way to determine when a successfully ingested stat has been fully processed, and that stat proliferated to the leaderboards in the back end.
+        /// Periodic refreshing of the leaderboard is one way to get up to date information.
+        /// Note that per user, you should never request stats more than 100 times a minute.
```
paulhazen, line 72: Great comments, I appreciate the attention to detail. Good work!

paulhazen (commented):
Commenting for now until I have time to evaluate the changes more than just to review the code.

Assets/Scripts/UI/Leaderboard/UILeaderboardMenu.cs
WispyMouse: That's a really agreeable standard. I'll change this over now.

Assets/Scripts/UI/Leaderboard/UILeaderboardMenu.cs
WispyMouse: Possibly! Let's consider what flexibility that gives. If a user has leaderboard `FOO` open, and calls `RefreshCurrentSelectedLeaderboard("bar")`, what should happen?

We can at least grab the updated information from the leaderboard api, but what else? In this current case, this can only be used to gather updated information and refresh the currently displayed leaderboard. If we allowed a leaderboard id as a parameter, what should happen? Should it tell the viewer to switch over to that leaderboard and display it? Should it refresh the data, but then not display it?

andrew-hirata-playeveryware (approved):

PR #694, opened 2024-05-28
Integrate EOS SDK version `1.16.3`

paulhazen:
This PR introduces changes that affect the integration of EOS SDK version `1.16.3` into the project.

The delta between `1.16.2` and `1.16.3` is fairly minimal, so this PR is correspondingly minimal. 

It has been deemed that re-compilation of the helper libraries within `lib/NativeCode` do _not_ need to be updated with this integration.


andrew-hirata-playeveryware (approved):

WispyMouse (approved):
No errors, builds, runs correctly.

I didn't find any obvious version numbers left over. I see that there's some updates to the dlls. Did any of the APIs we're calling increment their API version?

PR #698, opened 2024-05-29
Handle Repeat RequestAppTicket Calls in SteamManager and Invalidate Bad Token

### LINKED ISSUE #686: SteamManager callback re-use edge case
This issue is pretty unlikely to hit others, but it hit me. Since the fix is a simple change thought I'd raise it.

Every call to `SteamManager.RequestAppTicket` shares the same 'CallResult' member object (appTicketCallResult).  If a callback from this function were to result in a new call to this function, then it will hang forever.

E.g. (not real code):

```c#
RequestAppTicket(OnAppTicketRequestComplete);

void OnAppTicketRequestComplete()
{
   if(failed)
   {
       // -> Instigate a single retry of the process, including a new call to RequestAppTicket
   }
}
```

Because the `CallResult` object gets populated during its own callback result, it ends up ill-configured and never triggers for the second call.

A simple fix is to add:

`appTicketCallResult = new CallResult<EncryptedAppTicketResponse_t>();`

Right before this existing line:

`appTicketCallResult.Set(call, RequestAppTicketResponse);`

Which means the new call is using a brand new CallResult object, and the old one will get GC'd. Now the second request comes back as expected.

----

**Side note** the event pattern used in SteamManger also requires small adjustment to not fail in the above use case (where request could be initiated by its own callback):

Here the event invokes and then nulls the line after, but if the invoke were to cause anotehr call, it would register events and then the line after they'd get cancelled.

```
appTicketEvent?.Invoke(null);
appTicketEvent = null;
```

So I adjusted the functions that do this pattern to instead do the clear earlier and invoke a local cached version:

Start of function:
```
var resultEventToTrigger = appTicketEvent;
appTicketEvent = null;
```

Invoke point:
```
resultEventToTrigger?.Invoke(null);
```

WispyMouse:
Before this PR, when the submits a `StartConnectLoginWithSteamAppTicket` call that results in a failed app token retrieval or a failed connect login, the plugin could get in to a state where subsequent calls to this method would not be resolved properly.

There are three areas being addressed:

- `appTicketCallResult` is set to a new instance of `CallResult<EncryptedAppTicketResponse_t>`. This helps to avoid garbage collection that stomps over previous failed calls.
- When a `StartConnectLoginWithSteamAppTicket` fails, always invalidate the `encryptedAppTicket` cache. This allows the code to resolve a situation where a `RequestAppTicket` call successfully returns a token, but for some reason it is not valid for logging in. In this changeset if a repeat call to `RequestAppTicket` may get a new token, it can now possibly use that new value instead of only using the previously successful cached token.
- When preparing to run the `appTicketEvent` event inside `RequestAppTicketResponse`, pull out the current value of `appTicketEvent` in to a new variable, set the `appTicketEvent` to null, then invoke the new variable. This is to help with the scenario where a user has a callback such that it would add an event to `appTicketEvent` in a way that would have been ignored because the delegate was being set to null after invoking.

This PR resolves issue #686.

### DIFF
--- a/Assets/Scripts/Steam/SteamManager.cs
+++ b/Assets/Scripts/Steam/SteamManager.cs
@@ -73,9 +73,9 @@ namespace PlayEveryWare.EpicOnlineServices.Samples.Steam
             {
                 Debug.LogError("Error occured when requesting Encrypted App Ticket");
 
-                // First set the appTicketEvent we're calling to its own variable, set the appTicketEvent to null, then invoke the variable we set aside
-                // This way if the callback for the function would add to the appTicketEvent, it isn't immediately set to null
-                // This pattern is applied to the other calls to raise appTicketEvent
+                // First set the appTicketEvent we're calling to its own variable, set the appTicketEvent to null, then invoke the variable we set aside.
+                // This way if the callback for the function would add to the appTicketEvent, it isn't immediately set to null.
+                // This pattern is applied to the other calls that raise appTicketEvent inside this function.
                 Action<string> appTicketEventLetIOFailureOrNotOkay = appTicketEvent;
                 appTicketEvent = null;
                 appTicketEventLetIOFailureOrNotOkay?.Invoke(null);
--- a/Assets/Scripts/Steam/SteamManager.cs
+++ b/Assets/Scripts/Steam/SteamManager.cs
@@ -72,8 +72,13 @@ namespace PlayEveryWare.EpicOnlineServices.Samples.Steam
             if (ioFailure || response.m_eResult != EResult.k_EResultOK)
             {
                 Debug.LogError("Error occured when requesting Encrypted App Ticket");
-                appTicketEvent?.Invoke(null);
+
+                // First set the appTicketEvent we're calling to its own variable, set the appTicketEvent to null, then invoke the variable we set aside
+                // This way if the callback for the function would add to the appTicketEvent, it isn't immediately set to null
+                // This pattern is applied to the other calls to raise appTicketEvent
+                Action<string> appTicketEventLetIOFailureOrNotOkay = appTicketEvent;
                 appTicketEvent = null;
+                appTicketEventLetIOFailureOrNotOkay?.Invoke(null);
                 return;
             }
 
@@ -91,8 +96,12 @@ namespace PlayEveryWare.EpicOnlineServices.Samples.Steam
             if (!success)
             {
                 Debug.LogError("Failed to retrieve Encrypted App Ticket");
-                appTicketEvent?.Invoke(null);
+
+                // See above in regards to raising appTicketEvent
+                Action<string> appTicketEventLetNotSuccess = appTicketEvent;
                 appTicketEvent = null;
+                appTicketEventLetNotSuccess?.Invoke(null);
+
                 return;
             }
 
@@ -101,8 +110,10 @@ namespace PlayEveryWare.EpicOnlineServices.Samples.Steam
             //convert to hex string
             encryptedAppTicket = System.BitConverter.ToString(buffer).Replace("-", "");
 
-            appTicketEvent?.Invoke(encryptedAppTicket);
+            // See above in regards to raising appTicketEvent
+            Action<string> appTicketEventLetSuccess = appTicketEvent;
             appTicketEvent = null;
+            appTicketEventLetSuccess?.Invoke(encryptedAppTicket);
         }
 
         private void OnApplicationQuit()
--- a/Assets/Scripts/Steam/SteamManager.cs
+++ b/Assets/Scripts/Steam/SteamManager.cs
@@ -423,28 +423,28 @@ namespace PlayEveryWare.EpicOnlineServices.Samples.Steam
 #if DISABLESTEAMWORKS
             onLoginCallback?.Invoke(new Epic.OnlineServices.Connect.LoginCallbackInfo() { ResultCode = Epic.OnlineServices.Result.UnexpectedError });
 #else
-            // If this method is called and the login fails, we need to clean out the cached encryptedAppTicket.
-            EOSManager.OnConnectLoginCallback cacheInvalidationOnFailureCallback = (Epic.OnlineServices.Connect.LoginCallbackInfo callbackInfo) => 
-            {
-                // First, determine if this is a failure. If it is, then clear out the cache.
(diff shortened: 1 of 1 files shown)


Assets/Scripts/Steam/SteamManager.cs
```diff
@@ -411,15 +423,29 @@ public void StartConnectLoginWithSteamAppTicket(EOSManager.OnConnectLoginCallbac
 #if DISABLESTEAMWORKS
             onLoginCallback?.Invoke(new Epic.OnlineServices.Connect.LoginCallbackInfo() { ResultCode = Epic.OnlineServices.Result.UnexpectedError });
 #else
+            // If this method is called and the login fails, we need to clean out the cached encryptedAppTicket.
+            EOSManager.OnConnectLoginCallback cacheInvalidationOnFailureCallback = (Epic.OnlineServices.Connect.LoginCallbackInfo callbackInfo) => 
+            {
+                // First, determine if this is a failure. If it is, then clear out the cache.
+                if (callbackInfo.ResultCode != Epic.OnlineServices.Result.Success)
+                {
+                    Debug.LogWarning($"SteamManager will clear the cached {nameof(this.encryptedAppTicket)} after failed Connect login attempt using this Steam App Ticket.");
```
paulhazen: If I was a developer who read this comment in the log, I'm not certain I would know what action I should take. Can you either convert this to logging an "information," thing - or add some language that describes what actions the developer might want to take if they see this log message?

Assets/Scripts/Steam/SteamManager.cs
```diff
@@ -411,15 +423,29 @@ public void StartConnectLoginWithSteamAppTicket(EOSManager.OnConnectLoginCallbac
...
+                // Then call the original callback we were provided.
+                onLoginCallback?.Invoke(callbackInfo);
+            };
+
             RequestAppTicket((string token) => {
                 if (token == null)
                 {
                     Debug.LogError("Connect Login failed: Unable to get Steam app ticket");
-                    onLoginCallback?.Invoke(new Epic.OnlineServices.Connect.LoginCallbackInfo() { ResultCode = Epic.OnlineServices.Result.UnexpectedError });
+                    cacheInvalidationOnFailureCallback?.Invoke(new Epic.OnlineServices.Connect.LoginCallbackInfo() { ResultCode = Epic.OnlineServices.Result.UnexpectedError });
```
paulhazen: +1 for making the callback action variable name 1000% more descriptive.

Assets/Scripts/Steam/SteamManager.cs
```diff
@@ -72,8 +72,13 @@ private void RequestAppTicketResponse(EncryptedAppTicketResponse_t response, boo
             if (ioFailure || response.m_eResult != EResult.k_EResultOK)
             {
                 Debug.LogError("Error occured when requesting Encrypted App Ticket");
-                appTicketEvent?.Invoke(null);
+
+                // First set the appTicketEvent we're calling to its own variable, set the appTicketEvent to null, then invoke the variable we set aside
+                // This way if the callback for the function would add to the appTicketEvent, it isn't immediately set to null
+                // This pattern is applied to the other calls to raise appTicketEvent
```
paulhazen: Tiny nit (very - very tiny): Either make the line breaks uniform or introduce punctuation to the sentences to delineate things.

Assets/Scripts/Steam/SteamManager.cs
```diff
@@ -411,15 +423,29 @@ public void StartConnectLoginWithSteamAppTicket(EOSManager.OnConnectLoginCallbac
 #if DISABLESTEAMWORKS
             onLoginCallback?.Invoke(new Epic.OnlineServices.Connect.LoginCallbackInfo() { ResultCode = Epic.OnlineServices.Result.UnexpectedError });
 #else
+            // If this method is called and the login fails, we need to clean out the cached encryptedAppTicket.
+            EOSManager.OnConnectLoginCallback cacheInvalidationOnFailureCallback = (Epic.OnlineServices.Connect.LoginCallbackInfo callbackInfo) =>
```
paulhazen: I bring this up as a matter of personal preference - a preference which I am _fairly certain_ is counter to the team's modus operandi - in cases like this where inline actions are being defined, I find it particularly helpful to make use of the `var` keyword in-order to make the code easier to immediately comprehend.

While that is _my_ preference, I think it differs from what the team prefers - but thought I would mention it here regardless as a point of conversation to either confirm that this is the "correct" way to do things as defined within the team, or explore whether adjustments can / should be made.

paulhazen (changes requested):
Very small comments - might not even warrant a change. Once addressed @WispyMouse please ping me and I will prompt re-submit my review.

andrew-hirata-playeveryware (approved):

Assets/Scripts/Steam/SteamManager.cs
WispyMouse: This has been adjusted, and adjusted a bit further.

Now the callback is only modified in the event where RequestAppTicket succeeds, rather than the modification running in the case where RequestAppTicket fails to get a token.

The log has been demoted to informational. I think your insight here that being a 'Warning' makes it sound like we're demanding action from the user, but really we just want to inform them of something.

No action should be required on the user's part. Does the new comment suit our needs?

Assets/Scripts/Steam/SteamManager.cs
WispyMouse: Added punctuation to the end of the sentences. I've always been really inconsistent about this kind of stylistic thing, and will take more care in the future!

Assets/Scripts/Steam/SteamManager.cs
WispyMouse: I'll make an informational post about this in the channel and ask for engineer's opinion on this.

The following is my own personal opinion: I never want to use "var". I find it difficult to understand what's being returned, and whenever var is used I have to rely on Visual Studio to guess at what it is. It makes code review harder and glancing over the code is more difficult, because I have to re-contextualize what the variable each time I look at it.

arthur740212 (approved):

Assets/Scripts/Steam/SteamManager.cs
paulhazen: I understand your perspective. Good call to poll the engineering team as a whole.

paulhazen (approved):

PR #703, opened 2024-06-04
fix (EAC): Replace text replacement strings

### LINKED ISSUE #701: 3.2.0 - wrong placeholders in PlatformSpecificAssets/EOS/[Windows|Mac|Linux]/EasyAntiCheat/Settings.json
**Describe the bug**
In plugin version 3.2.0 `PlatformSpecificAssets/EOS/[Windows|Mac|Linux]/EasyAntiCheat/Settings.json` contains wrong placeholders
Should be
```	"title"											: "<ProductName>",
	"executable"									: "<ExeName>",
	"productid"										: "<ProductID>",
	"sandboxid"										: "<SandboxID>",
	"deploymentid"									: "<DeploymentID>",
```
But in fact:
```	"title"											: "Example Game",
	"executable"									: "<gamebinary.exe>",
	"productid"										: "<your_eos_productid>",
	"sandboxid"										: "<your_eos_sandboxid>",
	"deploymentid"									: "<your_eos_deploymentid>",
```
And code in
`PlayEveryWare.EpicOnlineServices.Build.EACUtility.ReplaceFileContentVars`
cannot find and properly replace them durng build

 - Plugin Version 3.2.0

WispyMouse:
When Easy Anti Cheat enabled builds are made, these settings files are copied over to the build and have their strings replaced. Recently these setting json base files have been modified to have different text.

There are essentially two simple ways to address this:

- This changelist, which changes the string targets that would be replaced during build back to their original
- Alternatively we can change the strings that are being targeted for replacement to the new text strings that are present in these settings

I think this changelist is what we should do, but I want to leave open the possibility that this isn't desirable for some reason.


arthur740212 (approved):

andrew-hirata-playeveryware (approved):
This seems like a good solution to me!

PR #710, opened 2024-06-06
Fixes post 3.2.0 release

arthur740212:
The input login was inverted during the refactor and was causing breakages when interacting with the overlay.
This fix applies the working logic.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -31,5 +31,3 @@
 *.asset             merge=unityyamlmerge eol=lf
 *.meta              merge=unityyamlmerge eol=lf
 *.controller        merge=unityyamlmerge eol=lf
-
-Assets/Plugins/Linux/libDynamicLibraryLoaderHelper.so filter=lfs diff=lfs merge=lfs binary
--- a/Assets/Scripts/UI/Login/UILoginMenu.cs
+++ b/Assets/Scripts/UI/Login/UILoginMenu.cs
@@ -319,6 +319,12 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
         /// <returns>True if input should be handled, false if not.</returns>
         private static bool ShouldInputBeHandled()
         {
+            if (null == EventSystem.current)
+            {
+                Debug.Log("EventSystem is null");
+                return false;
+            }
+
             bool shouldHandle = !EOSManager.Instance.IsOverlayOpenWithExclusiveInput();
 
 #if ENABLE_INPUT_SYSTEM
--- a/Assets/Scripts/UI/Login/UILoginMenu.cs
+++ b/Assets/Scripts/UI/Login/UILoginMenu.cs
@@ -339,15 +339,14 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
         }
 
 #if ENABLE_DEBUG_INPUT
-        static bool previousShouldHandle = false;
-        static void LogInputChanged(bool shouldHandle)
+        private static bool previousShouldHandle = false;
+        private static void LogInputChanged(bool shouldHandle)
         {
-            bool result = previousShouldHandle != shouldHandle ? true : false;
-            if (result)
+            if (previousShouldHandle != shouldHandle)
             {
                 Debug.LogWarning($"Input {(shouldHandle ? "enabled" : "disabled")} for main app");
+                previousShouldHandle = shouldHandle;
             }
-            previousShouldHandle = shouldHandle;
         }
 #endif
         /// <summary>
--- a/Assets/Scripts/UI/Login/UILoginMenu.cs
+++ b/Assets/Scripts/UI/Login/UILoginMenu.cs
@@ -319,12 +319,14 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
         /// <returns>True if input should be handled, false if not.</returns>
         private static bool ShouldInputBeHandled()
         {
+            // Event System isn't found, so main app cannot handle input
             if (null == EventSystem.current)
             {
                 Debug.Log("EventSystem is null");
                 return false;
             }
 
+            // Main app handles input if overlay isn't open
             bool shouldHandle = !EOSManager.Instance.IsOverlayOpenWithExclusiveInput();
 
 #if ENABLE_INPUT_SYSTEM
--- a/Assets/Scripts/UI/Login/UILoginMenu.cs
+++ b/Assets/Scripts/UI/Login/UILoginMenu.cs
@@ -330,11 +330,26 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
             EventSystem.current.sendNavigationEvents = shouldHandle;
 #endif
 #endif
-            Debug.Log($"Input {(shouldHandle ? "enabled" : "disabled")} due to EOS Overlay.");
 
+#if ENABLE_DEBUG_INPUT
+            LogInputChanged(shouldHandle);
+#endif
             return shouldHandle;
+
         }
 
+#if ENABLE_DEBUG_INPUT
+        static bool previousShouldHandle = false;
+        static void LogInputChanged(bool shouldHandle)
+        {
+            bool result = previousShouldHandle != shouldHandle ? true : false;
+            if (result)
+            {
+                Debug.LogWarning($"Input {(shouldHandle ? "enabled" : "disabled")} for main app");
+            }
+            previousShouldHandle = shouldHandle;
+        }
+#endif
         /// <summary>
         /// Determines whether a GameObject needs to be set as selected.
         /// </summary>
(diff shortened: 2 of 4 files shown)


Assets/Scripts/UI/Login/UILoginMenu.cs
```diff
@@ -319,29 +319,37 @@ private static void PreventDeselection(ref GameObject previouslySelectedGameObje
...
-            return true;
+#if ENABLE_DEBUG_INPUT
+            LogInputChanged(shouldHandle);
 #endif
+            return shouldHandle;
+
         }
 
+#if ENABLE_DEBUG_INPUT
+        static bool previousShouldHandle = false;
```
paulhazen: I believe our practice for this circumstance would be for this to be named `s_previousShouldHandle`. Either way, I think it's probably appropriate for it to be made `private`.

Separately, is it guaranteed that the previous state of whether input should be handled is `false`?

Assets/Scripts/UI/Login/UILoginMenu.cs
```diff
@@ -319,29 +319,37 @@ private static void PreventDeselection(ref GameObject previouslySelectedGameObje
...
+        static void LogInputChanged(bool shouldHandle)
+        {
+            bool result = previousShouldHandle != shouldHandle ? true : false;
+            if (result)
```
paulhazen: I think it would be clearer if you changed this to be

```c#
if (previousShouldHandle != shouldHandle)
```
(including suggested changes from earlier comment)

That way you can entirely remove line `345`.

Assets/Scripts/UI/Login/UILoginMenu.cs
```diff
@@ -319,29 +319,37 @@ private static void PreventDeselection(ref GameObject previouslySelectedGameObje
...
+#if ENABLE_DEBUG_INPUT
+            LogInputChanged(shouldHandle);
 #endif
+            return shouldHandle;
+
         }
 
+#if ENABLE_DEBUG_INPUT
+        static bool previousShouldHandle = false;
+        static void LogInputChanged(bool shouldHandle)
```
paulhazen: As far as I can tell, I think this should be marked as `private`.

Assets/Scripts/UI/Login/UILoginMenu.cs
```diff
@@ -319,29 +319,37 @@ private static void PreventDeselection(ref GameObject previouslySelectedGameObje
...
+        {
+            bool result = previousShouldHandle != shouldHandle ? true : false;
+            if (result)
+            {
+                Debug.LogWarning($"Input {(shouldHandle ? "enabled" : "disabled")} for main app");
+            }
+            previousShouldHandle = shouldHandle;
```
paulhazen: I believe that this assignment should be within the conditional.

Assets/Scripts/UI/Login/UILoginMenu.cs
```diff
@@ -319,29 +319,37 @@ private static void PreventDeselection(ref GameObject previouslySelectedGameObje
...
         }
 
+#if ENABLE_DEBUG_INPUT
+        static bool previousShouldHandle = false;
+        static void LogInputChanged(bool shouldHandle)
+        {
+            bool result = previousShouldHandle != shouldHandle ? true : false;
+            if (result)
+            {
+                Debug.LogWarning($"Input {(shouldHandle ? "enabled" : "disabled")} for main app");
```
paulhazen, line 355: I've found that for clarity it can be useful to (instead of having an inline ternary statement) rephrase the log so it makes sense for the value `true` or `false` to be used. Something like:
```C#
Debug.LogWarning($"Should input be handled: {shouldHandle}.");
```

This is _not_ a critique relevant exactly for this PR - since you merely relocated the log line - and (if memory serves) I'm the one who originally wrote it 😆

Assets/Scripts/UI/Login/UILoginMenu.cs
```diff
@@ -319,29 +319,37 @@ private static void PreventDeselection(ref GameObject previouslySelectedGameObje
         /// <returns>True if input should be handled, false if not.</returns>
         private static bool ShouldInputBeHandled()
         {
-            // TODO: Clarify why this is only evaluated for PS4 and PS5
-#if UNITY_PS4 || UNITY_PS5
-            // TODO: Simplify this conditional - it's sort of confusing to read despite doing exactly what we want
-            if (null == EventSystem.current || EventSystem.current.sendNavigationEvents == !EOSManager.Instance.IsOverlayOpenWithExclusiveInput())
```
paulhazen, line 325: I think we still need to check if `EventSystem.current` is `null`. While it was far from perfect, this conditional _did_ provide safety against referencing a null pointer.

paulhazen (changes requested):

.gitattributes
```diff
@@ -30,6 +30,4 @@
 *.physicMaterial    merge=unityyamlmerge eol=lf
 *.asset             merge=unityyamlmerge eol=lf
 *.meta              merge=unityyamlmerge eol=lf
-*.controller        merge=unityyamlmerge eol=lf
-
-Assets/Plugins/Linux/libDynamicLibraryLoaderHelper.so filter=lfs diff=lfs merge=lfs binary
```
WispyMouse, line 35: Is this removal intentional?

Assets/Scripts/UI/Login/UILoginMenu.cs
WispyMouse: Agreeable; it seems like we're using EventSystem.current without checking that it exists, later in this function.

Assets/Scripts/UI/Login/UILoginMenu.cs
```diff
@@ -319,29 +319,37 @@ private static void PreventDeselection(ref GameObject previouslySelectedGameObje
...
             bool shouldHandle = !EOSManager.Instance.IsOverlayOpenWithExclusiveInput();
-                
-            Debug.Log($"Input {(shouldHandle ? "enabled" : "disabled")} due to EOS Overlay.");
+
 #if ENABLE_INPUT_SYSTEM
             EventSystem.current.currentInputModule.enabled = shouldHandle;
+            EventSystem.current.sendNavigationEvents = shouldHandle;
 #else
+            // TODO: Clarify why this is only evaluated for PS4 and PS5
+#if UNITY_PS4 || UNITY_PS5
```
WispyMouse, line 337: Is this an appropriate place for UNITY_PS4 and UNITY_PS5 specific code? Would this be appropriate to always set like this, even if ENABLE_INPUT_SYSTEM is not true?

Assets/Scripts/UI/Login/UILoginMenu.cs
arthur740212: It is the input state during startup, so it does not matter whether it is false or true.
It is private already, but is has to be static if used in a static function, I could add a private keyword to be precise

WispyMouse (changes requested):
I think it is probably necessary to nullcheck against EventSystem.current, like it used to. I can be convinced otherwise.

Assets/Scripts/UI/Login/UILoginMenu.cs
arthur740212: yep, there used to be a `return result;`, then the function was rewritten into a void, so it is not needed anymore, nice find.

.gitattributes
arthur740212: Sort of, this PR was supposed to be all fixes post 3.2.0, there was only one thus the name.
I'll rename the PR to reflect the branch

Assets/Scripts/UI/Login/UILoginMenu.cs
arthur740212: It is private, will add the keyword for less confusion.

Assets/Scripts/UI/Login/UILoginMenu.cs
arthur740212: Yep, putting it out doesn't do more because we only need to reassign if the values changed, nice catch.

Assets/Scripts/UI/Login/UILoginMenu.cs
arthur740212: Fixed

.gitattributes
andrew-hirata-playeveryware: Both those removes seems suspicious to me.

arthur740212: Only the linux line is actually removed, the other one is because of line ending changes

andrew-hirata-playeveryware (approved):

.gitattributes
arthur740212: The linux line is moved to the `.gitattribute` under `Linux/`
Since there is no going back from lfs I suppose

WispyMouse (approved):

paulhazen (approved):

PR #711, opened 2024-06-07
Relocate `TryGetTempDirectory` function to `FileUtility` from `PackageFileUtility`

paulhazen:
This PR moves the `TryGetTempDirectory` function from `PackageFileUtility` to `FileUtility`, as its functionality is utilized in contexts other than package creation.

Some minor improvements to the implementation were made to reduce repeated code. Places that called the function were updated accordingly.

### DIFF
--- a/Assets/Plugins/Source/Core/Utility/FileUtility.cs
+++ b/Assets/Plugins/Source/Core/Utility/FileUtility.cs
@@ -33,26 +33,80 @@ namespace PlayEveryWare.EpicOnlineServices.Utility
     public static class FileUtility
     {
         /// <summary>
-        /// Generates a unique and new temporary directory inside the Temporary Cache Path as determined by Unity,
-        /// and returns the fully-qualified path to the newly created directory.
+        /// Generates a unique and new temporary directory inside the Temporary
+        /// Cache Path as determined by Unity, and returns the fully-qualified
+        /// path to the newly created directory. The directory name will be
+        /// "Temp-" appended with a GUID.
         /// </summary>
-        /// <returns>Fully-qualified file path to the newly generated directory.</returns>
-        public static string GenerateTempDirectory()
+        /// <param name="path">
+        /// Fully-qualified path to the newly generated directory if one was
+        /// created, otherwise path is set to default System.String value.
+        /// </param>
+        /// <returns>
+        /// True if the temporary directory was created, false otherwise.
+        /// </returns>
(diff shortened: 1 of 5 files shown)


Assets/Plugins/Source/Core/Utility/FileUtility.cs
```diff
@@ -33,26 +33,80 @@ namespace PlayEveryWare.EpicOnlineServices.Utility
...
+        /// </returns>
+        public static bool TryGetTempDirectory(out string path)
         {
+            path = default;
+            
+            // Nested local function to reduce repetitive code.
+            string GenerateTempPath()
+            {
+                return Path.Combine(
+                    Application.temporaryCachePath,
```
WispyMouse, line 56: For reasons relating to particular platform's reaction to asking for things like Application.temporaryCachePath, can you make a static function or something that this value can be grabbed from? That'd allow easy extension on other platforms for getting this. Otherwise I guess those changes can just be made in-line here.

Assets/Plugins/Source/Core/Utility/FileUtility.cs
```diff
@@ -33,26 +33,80 @@ namespace PlayEveryWare.EpicOnlineServices.Utility
...
-            while (Directory.Exists(tempDirectory))
+            // If (by some crazy miracle) the directory path already exists,
+            // try once more.
+            if (Directory.Exists(tempPath))
             {
-                tempDirectory = Path.Combine(Application.temporaryCachePath, $"/Output-{Guid.NewGuid()}/");
+                Debug.LogWarning(
+                    $"The temporary directory created collided with " +
+                    $"an existing temporary directory of the same name. This " +
+                    $"is very unlikely.");
```
WispyMouse, line 70: not a critique: very, very, very unlikely. Something very strange and wrong must happen for a guid collision, ahah.

WispyMouse (approved):
Simple moving around of stuff. This seems good to me. The idea of a double name collision on two guids is really funny.

My only concern remaining is what happens when a name is too-long. Throwing exceptions, as we do, seems like the right response.

I have some small concern about using Applicatoin.temporaryCachePath because of compatibility with some platforms.

Assets/Plugins/Source/Core/Utility/FileUtility.cs
paulhazen: I'm not certain. That is worth considering, but is outside the scope of these changes.

PR #712, opened 2024-06-07
Relocate file copy operations from `PackageFileUtility` to `FileUtility`

paulhazen:
This PR moves the asynchronous copy file operation functions from `PackageFileUtility.cs` to `FileUtility.cs` and makes clearer how the functions operate by renaming some of the signatures and data structures involved.

Corresponding places in the code where those functions and data structures are referenced have been updated.

### DIFF
--- a/Assets/Plugins/Source/Core/Utility/Extensions/ListExtensions.cs
+++ b/Assets/Plugins/Source/Core/Utility/Extensions/ListExtensions.cs
@@ -1,0 +1,68 @@
+/*
+ * Copyright (c) 2024 PlayEveryWare
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+namespace PlayEveryWare.EpicOnlineServices.Extensions
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Threading;
+
+    /// <summary>
+    /// Threadsafe Random class.
+    /// </summary>
+    internal static class ThreadSafeRandom
+    {
+        private static readonly ThreadLocal<Random> threadLocalRandom = new(() => new Random());
+
+        public static Random Instance => threadLocalRandom.Value;
+    }
+
+    /// <summary>
+    /// Contains extension methods for lists.
+    /// </summary>
+    public static class ListExtensions
+    {
+        /// <summary>
+        /// Uses the Fisher-Yates algorithm to efficiently and uniformly shuffle
+        /// the contents of an object that implements the generic list
+        /// interface.
+        /// </summary>
+        /// <typeparam name="T">The template parameter for the list.</typeparam>
+        /// <param name="list">List to shuffle the contents of.</param>
+        public static void Shuffle<T>(this IList<T> list)
+        {
+            int n = list.Count;
+            for (int i = 0; i < n; i++)
+            {
+                // Lock to ensure things work as expected in threaded
+                // circumstances.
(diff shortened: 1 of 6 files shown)


Assets/Plugins/Source/Core/Utility/FileUtility.cs
```diff
@@ -32,6 +36,62 @@ namespace PlayEveryWare.EpicOnlineServices.Utility
...
+
+        /// <summary>
+        /// Stores information regarding a file copy operation.
+        /// </summary>
+        public struct CopyFileOperation
+        {
+            /// <summary>
+            /// The fully-qualified path of the source file.
+            /// </summary>
+            public string From;
```
andrew-hirata-playeveryware: This might be pedantic, but instead of 'from' and 'to', can we use something like 'SourcePath' and 'DestinationPath'?

Assets/Plugins/Source/Core/Utility/FileUtility.cs
```diff
@@ -22,8 +22,12 @@
 
 namespace PlayEveryWare.EpicOnlineServices.Utility
 {
+    using Extensions;
     using System;
+    using System.Collections.Generic;
     using System.IO;
+    using System.Linq;
```
WispyMouse: Danger, System.Linq spotted in runtime code! Is this an acceptable use of it?

andrew-hirata-playeveryware (dismissed):

Assets/Plugins/Source/Editor/Windows/CreatePackageWindow.cs
```diff
@@ -262,14 +262,14 @@ private async void StartCreatePackageAsync(UPMUtility.PackageType type)
...
-            var progressHandler = new Progress<UnityPackageCreationUtility.CreatePackageProgressInfo>(value =>
+            var progressHandler = new Progress<FileUtility.CopyFileProgressInfo>(value =>
             {
                 var fileCountStrSize = value.TotalFilesToCopy.ToString().Length;
                 string filesCopiedStrFormat = "{0," + fileCountStrSize + "}";
                 var filesCopiedCountStr = String.Format(filesCopiedStrFormat, value.FilesCopied);
                 var filesToCopyCountStr = String.Format(filesCopiedStrFormat, value.TotalFilesToCopy);
 
-                _progress = value.SizeOfFilesCopied / (float)value.TotalSizeOfFilesToCopy;
+                _progress = value.BytesCopied / (float)value.TotalBytesToCopy;
```
WispyMouse: Is there ever a cursed world where TotalBytesToCopy is somehow 0? Like maybe a file with no contents and just a name?

Assets/Plugins/Source/Core/Utility/Extensions/ListExtensions.cs
```diff
@@ -0,0 +1,68 @@
...
+            for (int i = 0; i < n; i++)
+            {
+                // Lock to ensure things work as expected in threaded
+                // circumstances.
+                lock (ThreadSafeRandom.Instance)
+                {
+                    // Get the next random index to shuffle
+                    int j = ThreadSafeRandom.Instance.Next(i, n);
+                    // Use deconstruction syntax to perform the swap.
+                    (list[i], list[j]) = (list[j], list[i]);
```
WispyMouse, line 63: not a critique; Wow I had never heard of this syntax form! Solves the classical "how do you swap two things without a third variable". I'm supposing this works just fine if `j == i` by circumstance of random.

Assets/Plugins/Source/Core/Utility/FileUtility.cs
```diff
@@ -55,6 +115,156 @@ public static string GenerateTempDirectory()
...
+        /// </summary>
+        /// <param name="filepaths">
+        /// The filepaths to get the directories for.
+        /// </param>
+        /// <param name="inOrder">
+        /// If true, return the list of directories by order shortest path
+        /// first.
+        /// </param>
+        /// <returns></returns>
+        private static IEnumerable<string> GetDirectories(IEnumerable<string> filepaths, bool inOrder = true)
```
WispyMouse: I almost feel like `inOrder` needs renaming. It definitely does make things go in _an_ order, but isn't clear what kind. That's probably what your handy param documentation is for, though, so maybe not.

WispyMouse (approved):
Approved! I feel like I learned a lot about some neat tricks from reading this pull request.

I had some naming comments, but everything here seems solid.

Assets/Plugins/Source/Core/Utility/FileUtility.cs
paulhazen: It is a reasonable question to ask, and to be more clear it should be guarded against, but the places where the `System.Linq` namespace is actually utilized are within functions that should only ever be called when in runtime. I'll adjust things to guarantee that is the case and ping you back once it's fixed.

Assets/Plugins/Source/Core/Utility/FileUtility.cs
paulhazen: Yeah, I think that's a reasonable thing to do. I _do_ prefer `From` and `To` to make things easier to read, but I think you're right that a more descriptive name is more appropriate.

Assets/Plugins/Source/Editor/Windows/CreatePackageWindow.cs
paulhazen: I don't believe so - but the change does not introduce a situation where that could happen - that possibility already exists. So worth pointing out, and I may make that change here and now - but technically it's outside the scope of the PR.

Assets/Plugins/Source/Core/Utility/Extensions/ListExtensions.cs
paulhazen: Right?!?!? I thought it was super cool.

Assets/Plugins/Source/Core/Utility/FileUtility.cs
paulhazen: I think a rename is appropriate. I think `creationOrder` would make sense, with notes indicating that it orders the directories in such a way that one can be created after the next. It just so _happens_ that order is also sorting the path strings by length in ascending order.

PR #713, opened 2024-06-10
QOL: Smooth progress bar

paulhazen:
Quality of life change to the implementation of the progress bar used in `CreatePackageWindow`. 

Makes the progress bar update smoothly, instead of being choppy.

### DIFF
--- a/Assets/Plugins/Source/Editor/Windows/CreatePackageWindow.cs
+++ b/Assets/Plugins/Source/Editor/Windows/CreatePackageWindow.cs
@@ -36,6 +36,7 @@ namespace PlayEveryWare.EpicOnlineServices.Editor.Windows
     using System.IO;
     using System.Threading;
     using System.Threading.Tasks;
+    using UnityEditorInternal;
     using Utility;
     using Config = EpicOnlineServices.Config;
 
@@ -58,8 +59,16 @@ namespace PlayEveryWare.EpicOnlineServices.Editor.Windows
         private CancellationTokenSource _createPackageCancellationTokenSource;
         
         private bool _operationInProgress;
-        private float _progress;
+
+        #region Progress Bar Stuff
+        
+        private float _actualProgress;
+        private float _displayedProgress;
         private string _progressText;
+        private object _progressLock = new object();
+        private Thread _progressUpdateThread;
+
+        #endregion
 
         [MenuItem("Tools/EOS Plugin/Create Package")]
         public static void ShowWindow()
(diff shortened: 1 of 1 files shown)


Assets/Plugins/Source/Editor/Windows/CreatePackageWindow.cs
```diff
@@ -203,22 +212,41 @@ protected override void RenderWindow()
...
+                    _actualProgress = 0.0f;
                     _progressText = "";
                     _createPackageCancellationTokenSource?.Cancel();
+                    FileUtility.CleanDirectory(_packagingConfig.pathToOutput);
                 }
                 GUILayout.EndVertical();
             }
         }
 
+        private void UpdateProgressBar(CancellationToken cancellationToken)
```
WispyMouse, line 236: The name of this function looks like it would actually update the progress bar, but all it seems to do is wait until cancellation is requested.

Does this thread close itself out upon a completion / non-cancellation?

Assets/Plugins/Source/Editor/Windows/CreatePackageWindow.cs
```diff
@@ -262,15 +290,29 @@ private async void StartCreatePackageAsync(UPMUtility.PackageType type)
...
+                {
+                    float newActualProgress = value.SizeOfFilesCopied / (float)value.TotalSizeOfFilesToCopy;
+
+                    // Just to guarantee that the progress is increasing
+                    if (newActualProgress > _actualProgress)
+                    {
+                        _actualProgress = newActualProgress;
+                    }
+
+                    _progressText = $"{filesCopiedCountStr} out of {filesToCopyCountStr} files copied";
```
WispyMouse, line 313: Since progress is measured in terms of size rather than files remaining, would it make sense to also include information about the number of files remaining? It surprised me when the progress bar was half-way done while being one fourth of the files remaining, then it eventually made sense as it took more time.

Assets/Plugins/Source/Editor/Windows/CreatePackageWindow.cs
```diff
@@ -311,7 +353,7 @@ private async void StartCreatePackageAsync(UPMUtility.PackageType type)
             {
                 _operationInProgress = false;
                 _progressText = "";
-                _progress = 0f;
+                _actualProgress = 0f;
```
WispyMouse, line 356: Should we tell _progressUpdateThread to stop here?

Assets/Plugins/Source/Editor/Windows/CreatePackageWindow.cs
```diff
@@ -203,22 +212,41 @@ protected override void RenderWindow()
...
         }
 
+        private void UpdateProgressBar(CancellationToken cancellationToken)
+        {
+            const int smoothingFactor = 500;
+            while (!cancellationToken.IsCancellationRequested)
+            {
+                Thread.Sleep(smoothingFactor);
+
+                if (cancellationToken.IsCancellationRequested)
```
WispyMouse, line 243: This makes me feel like something else was going to go on in here. while cancellation isn't requested, sleep, then check if cancellation is requested, if not then loop.

WispyMouse (approved):
This appears to work as desired! Left some questions to look in to if you'd like.

andrew-hirata-playeveryware (approved):

Assets/Plugins/Source/Editor/Windows/CreatePackageWindow.cs
paulhazen: I'm not certain I understand - the progress bar _is _being updated here to tell you how many files are left. I'm going to assume you're calling out the discrepancy between having the progress of the bar be based on file size, while the text indicates how many files. I chose to make this distinction because it provided the best experience upon experimentation, since the alternative would have it stuck on a single file for a while if it's particularly large.

Assets/Plugins/Source/Editor/Windows/CreatePackageWindow.cs
paulhazen: Nope - what's really nice about the way this is done is that the `_progressUpdateThread` is created using a `CancellationToken`, and all we need to do is cancel the token, and it will be handled for us. 

For the sake of simplicity, you can use the same token for multiple threads. In our case - we have one thread that's copying files, and one thread that's pausing the UI briefly in order to make the progress bar update smoothly - both circumstances can use the same `CancellationToken` so we can cancel them simultaneously.

Assets/Plugins/Source/Editor/Windows/CreatePackageWindow.cs
paulhazen: I think is tied to the fact that the function name it is within is also misleading. But I'll address those concerns in a separate PR where I move all of the ProgressBar stuff into its own class, so that `CreatePackageWindow` does not have to include code for ProgressBar - something it really is out of scope for it to know.

Assets/Plugins/Source/Editor/Windows/CreatePackageWindow.cs
paulhazen: It does indeed. I also agree that the function name is a bit misleading 🤔

PR #714, opened 2024-06-10
Package creation and `.meta` files

paulhazen:
## Summary of Changes

1. **Exclude `.meta` Files in Tilde Directories (when creating package)**

   Unity ignores files in directories containing a tilde (`~`). Therefore, `.meta` files in such directories are unnecessary and can cause confusion. This update ensures `.meta` files are not added to these directories. See [Unity's Documentation](https://docs.unity3d.com/Manual/SpecialFolders.html) for confirmation of this behavior.

   One exception to this is that the files within the `Samples~` directory should have `.meta` files included, because while the contents of that directory are not initially considered by the import process that Unity uses, they _are_ considered once the contents of the samples directory are imported subsequent to the plugin being added.

2. **Auto-Include `.meta` Files in Package Creation**

   For directories without a tilde, all files and folders must have corresponding `.meta` files. This update automates the inclusion of these files, eliminating the need to manually reference them in the `eos_package_description.json` file, thereby reducing its size _considerably_.

3. **Additional testing for file equality**

   In the process of making these changes, I added some debug stuff to the `FileInfoExtensions` class for comparing files - so that we can more precisely see the way in which one file may not be equal to another.

## Testing

Testing involved comparing the package contents created with and without these changes. The new process correctly added missing `.meta` files and removed `.meta` files from tilde directories, ensuring consistency and accuracy.


Assets/Plugins/Source/Editor/Utility/PackageFileUtility.cs
```diff
@@ -208,20 +244,22 @@ private static IEnumerable<FileInfoMatchingResult> FindMatchingFiles(string root
             CancellationToken cancellationToken = default,
             Action<string> postProcessCallback = null)
         {
-            // TODO: A major improvement to this system would be to automatically include .meta files, and gracefully
-            //       handle instances where the meta file *should* exist, but does not - gracefully in this context
-            //       would mean that if the .meta file exists, it should just be copied over; whereas if there is a
-            //       file entry copied over that does not have a corresponding meta file in the source, it indicates
-            //       that a meta file entry is MISSING from the source, and that should stop the creation of a package.
```
paulhazen, line 215: This `TODO` remark is accomplished by this PR, so it is another way to understand it if you would like. The only change from the implementation suggested above is a missing meta file does not result in a failed build but a logged warning.

arthur740212:
This will be a problem if files inside `~folders` hold a reference to another file within `~folders`.
For example, all samples are under `Samples~`

paulhazen:
> This will be a problem if files inside `~folders` hold a reference to another file within `~folders`. For example, all samples are under `Samples~`

Excellent catch. The nuance here is that while it's true that the meta files in `Samples~` will initially be ignored by Unity, they _will_ be considered the moment the samples are imported. So, an exception to the rule must be that files within `Samples~` must include `.meta` files to accommodate that scenario.

I have just updated the implementation and confirmed that the `Samples~` perform as expected.

Assets/Plugins/Source/Editor/Utility/PackageFileUtility.cs
```diff
@@ -180,7 +186,41 @@ private static IEnumerable<FileInfoMatchingResult> FindMatchingFiles(string root
...
+                    // If the destination path does not contain a tilde, and
+                    // the source file does _not_ have a .meta extension, then
+                    // AND if the current file has a sibling that _does_ have a
+                    // .meta extension, and we should copy that file additionally.
+                    if (".meta" != src.Extension && File.Exists($"{src.FullName}.meta"))
+                    {
+                        FileInfoMatchingResult metaFile = file;
+                        metaFile.fileInfo = new FileInfo($"{src.FullName}.meta");
+                        metaFile.originalSrcDestPair = file.originalSrcDestPair;
+                        //metaFile.relativePath = $"{file.relativePath}.meta";
```
paulhazen: remove commented out line.

Assets/Plugins/Source/Editor/Windows/CreatePackageWindow.cs
```diff
@@ -301,7 +301,7 @@ private async void StartCreatePackageAsync(UPMUtility.PackageType type)
                 {
                     // Ternary statement here to prevent a divide by zero problem
                     // ever happening, despite how odd it would be in this case.
-                    float newActualProgress = (0.0f >= value.TotalBytesToCopy)
+                    float newActualProgress = (0.0f <= value.TotalBytesToCopy)
```
paulhazen: This needs to just be a `<` symbol.

andrew-hirata-playeveryware (approved):

PR #715, opened 2024-06-11
Introduction of `EOSStatsManager`

paulhazen:
## Summary of Changes

1. **Decouple Stats Management from `AchievementManager`**

    All functionality pertaining to Stats management that _did_ exist within `AchievementManager` was moved into a new manager (`EOSStatsManager`). This is the first step toward efforts to delineate the different sample scenes approximately by which "Interfaces" (as defined by Epic) they demonstrate.

2. **Move Common Functionality to `EOSConnectLogin`**

    Both `AchievementManager` and `StatsManager` share functionality surrounding their implementation of the `IEOSConnectLogin` interface - so they were updated to implement a new abstract class called `EOSConnectLogin`. 

    Nominally, a project should only really ever have an `interface` _or_ an `abstract` class, however I did not want to refactor code that was unrelated to the work that was immediately at hand - that is: separating the stats functionality from the achievements functionality. Nonetheless, I did not want to continue the practice in the files that this PR needed to touch.

### DIFF
--- a/Assets/Plugins/Source/Core/EOSOnConnectLogin.cs
+++ b/Assets/Plugins/Source/Core/EOSOnConnectLogin.cs
@@ -1,0 +1,58 @@
+/*
+ * Copyright (c) 2021 PlayEveryWare
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+namespace PlayEveryWare.EpicOnlineServices
+{
+    using Epic.OnlineServices;
+    using Epic.OnlineServices.Connect;
+    using System;
+    using UnityEngine;
+
+    /// <summary>
+    /// Contains common implementations for the IEOSOnConnectLogin interface.
+    /// Not all instances where IEOSOnConnectLogin is implemented are replaced
+    /// by this abstract class, because a gradual conversion will have a smaller
+    /// impact.
+    /// </summary>
+    public abstract class EOSOnConnectLogin : IEOSOnConnectLogin
+    {
+        /// <summary>
+        /// The product user id associated with the connect login that called
+        /// the connect login callback for this manager.
+        /// </summary>
+        protected ProductUserId _productUserId;
+
+        public void OnConnectLogin(LoginCallbackInfo loginCallbackInfo)
+        {
+            if (loginCallbackInfo.ResultCode == Result.Success)
+            {
+                _productUserId = loginCallbackInfo.LocalUserId;
+            }
+            else
+            {
+                Debug.LogError($"Connect login failed with result code: {Enum.GetName(typeof(Result), loginCallbackInfo.ResultCode)}");
+            }
+        }
+
+        protected abstract void PostSuccessfulLogin();
+    }
+}
--- a/Assets/Plugins/Source/Editor/Windows/CreatePackageWindow.cs
+++ b/Assets/Plugins/Source/Editor/Windows/CreatePackageWindow.cs
@@ -301,7 +301,7 @@ namespace PlayEveryWare.EpicOnlineServices.Editor.Windows
                 {
                     // Ternary statement here to prevent a divide by zero problem
                     // ever happening, despite how odd it would be in this case.
-                    float newActualProgress = (0.0f <= value.TotalBytesToCopy)
+                    float newActualProgress = (0.0f < value.TotalBytesToCopy)
                         ? value.BytesCopied / (float)value.TotalBytesToCopy
                         : 0;
 
@@ -338,7 +338,7 @@ namespace PlayEveryWare.EpicOnlineServices.Editor.Windows
                     }
                 }
 
-                await UPMUtility.CreatePackage(type, progressHandler, _createPackageCancellationTokenSource.Token);
+                await UPMUtility.CreatePackage(type, false, progressHandler, _createPackageCancellationTokenSource.Token);
 
                 if (EditorUtility.DisplayDialog("Package Created", "Package was successfully created",
                         "Open Output Path", "Close"))
--- a/Assets/Plugins/Source/Editor/Windows/CreatePackageWindow.cs
+++ b/Assets/Plugins/Source/Editor/Windows/CreatePackageWindow.cs
@@ -301,7 +301,7 @@ namespace PlayEveryWare.EpicOnlineServices.Editor.Windows
                 {
                     // Ternary statement here to prevent a divide by zero problem
                     // ever happening, despite how odd it would be in this case.
-                    float newActualProgress = (0.0f >= value.TotalBytesToCopy)
+                    float newActualProgress = (0.0f <= value.TotalBytesToCopy)
                         ? value.BytesCopied / (float)value.TotalBytesToCopy
                         : 0;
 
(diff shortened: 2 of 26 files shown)


Assets/Plugins/Source/Editor/Windows/CreatePackageWindow.cs
```diff
@@ -301,7 +301,7 @@ private async void StartCreatePackageAsync(UPMUtility.PackageType type)
                 {
                     // Ternary statement here to prevent a divide by zero problem
                     // ever happening, despite how odd it would be in this case.
-                    float newActualProgress = (0.0f >= value.TotalBytesToCopy)
+                    float newActualProgress = (0.0f <= value.TotalBytesToCopy)
```
WispyMouse: Oh hey, this is familiar! Was this hit in another PR?

paulhazen: Yep! Although I duplicated the mistake - so good that you noticed this twice :D

Assets/Plugins/Source/Core/EOSOnConnectLogin.cs
```diff
@@ -0,0 +1,58 @@
...
+            {
+                _productUserId = loginCallbackInfo.LocalUserId;
+            }
+            else
+            {
+                Debug.LogError($"Connect login failed with result code: {Enum.GetName(typeof(Result), loginCallbackInfo.ResultCode)}");
+            }
+        }
+
+        protected abstract void PostSuccessfulLogin();
```
WispyMouse, line 56: This *looks* like it's supposed to be called by OnConnetLogin.

Assets/Plugins/Source/Core/EOSOnConnectLogin.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        public void OnConnectLogin(LoginCallbackInfo loginCallbackInfo)
+        {
+            if (loginCallbackInfo.ResultCode == Result.Success)
+            {
+                _productUserId = loginCallbackInfo.LocalUserId;
+            }
+            else
+            {
+                Debug.LogError($"Connect login failed with result code: {Enum.GetName(typeof(Result), loginCallbackInfo.ResultCode)}");
```
WispyMouse, line 52: not a critique: Wild, I've never seen this use of Enum before! I'll think about when to apply this versus just tostringing the enum.

Assets/Plugins/Source/Core/EOSOnConnectLogin.cs
paulhazen: It does indeed look like that - I'll look into it - because I thought that's what I had done 🤔

Assets/Scripts/EOSStatsManager.cs
```diff
@@ -0,0 +1,176 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+//#define ENABLE_DEBUG_EOSSTATSMANAGER;
+
+using System.Collections.Generic;
+using Epic.OnlineServices.Stats;
+using System;
```
WispyMouse, line 27: tiny nit: Can we reorder these using statements?

Assets/Plugins/Source/Core/EOSOnConnectLogin.cs
```diff
@@ -0,0 +1,58 @@
...
+    /// by this abstract class, because a gradual conversion will have a smaller
+    /// impact.
+    /// </summary>
+    public abstract class EOSOnConnectLogin : IEOSOnConnectLogin
+    {
+        /// <summary>
+        /// The product user id associated with the connect login that called
+        /// the connect login callback for this manager.
+        /// </summary>
+        protected ProductUserId _productUserId;
```
WispyMouse, line 42: Hmmm... have we created an additional source of truth for the product user id? Should we pull from EOSManager / EOSSingleton?

WispyMouse: Maybe if all this function is doing is passing this along on a successful login, instead of storing the value here, it's just passed in as an argument to PostSuccessfulLogin?

Assets/Scripts/EOSStatsManager.cs
```diff
@@ -0,0 +1,176 @@
...
+            {
+                TargetUserId = productUserId,
+                StatIndex = 0
+            };
+
+            for (uint i = 0; i < statsCountForProductUserId; ++i)
+            {
+                copyStatsByIndexOptions.StatIndex = i;
+
+                Result copyStatResult = statInterface.CopyStatByIndex(ref copyStatsByIndexOptions, out Stat? stat);
```
WispyMouse, line 80: not a critique: Legitimately didn't know you could use the null coalescing thing on an out. Interesting! I guess it provides another way of knowing if the function got something.

Assets/Scripts/EOSStatsManager.cs
```diff
@@ -0,0 +1,176 @@
...
+
+            _productUserIdToStatsCache[productUserId] = collectedStats;
+        }
+
+        private void QueryStatsForProductUserId(ProductUserId productUserId, OnQueryStatsCompleteCallback callback)
+        {
+            if (!productUserId.IsValid())
+            {
+                print("Invalid product user id sent in!");
+                return;
```
WispyMouse, line 96: I feel like we should run the callback with an OnQueryStatsCompleteCallbackInfo that indicates a failure.

Assets/Scripts/EOSStatsManager.cs
```diff
@@ -0,0 +1,176 @@
...
+                LocalUserId = productUserId,
+                TargetUserId = productUserId
+            };
+
+            statInterface.QueryStats(ref statsOptions, null, (ref OnQueryStatsCompleteCallbackInfo queryStatsCompleteCallbackInfo) =>
+            {
+                if (queryStatsCompleteCallbackInfo.ResultCode != Result.Success)
+                {
+                    // TODO: handle error
+                    print("Failed to query stats: " + queryStatsCompleteCallbackInfo.ResultCode);
```
WispyMouse, line 111: Similar above; can we run the callback function provided passing in a struct indicating failure?

Assets/Scripts/EOSStatsManager.cs
```diff
@@ -0,0 +1,176 @@
...
+                }
+                if (callback != null)
+                {
+                    callback(ref queryStatsCompleteCallbackInfo);
+                }
+            });
+        }
+
+        public void RefreshData()
```
WispyMouse, line 120: Should there be some kind of callback for informing the user of when this operation has completed?

Assets/Scripts/EOSStatsManager.cs
```diff
@@ -0,0 +1,176 @@
...
+        {
+            ProductUserId productUserId = EOSManager.Instance.GetProductUserId();
+            QueryStatsForProductUserId(productUserId, (ref OnQueryStatsCompleteCallbackInfo statsQueryData) =>
+            {
+                CacheStatsForProductUserId(productUserId);
+            });
+        }
+
+        public void IngestStat(string statName, int statDelta)
```
WispyMouse, line 129: `statDelta` makes it sound like you're ingesting a stat by saying, "add 5 to this stat name". I'm not certain that's what this is supposed to be, though. Can you confirm if this is a delta or a direct absolute value to set?

WispyMouse: The reason I think it's not like this is: Imagine a highscore table of "least amount of seconds needed to clear a race course". If I finished it in 100 seconds, and my previous record is 123, do I;
- Submit -23,
- or Submit 100?

Assets/Plugins/Source/Editor/Windows/CreatePackageWindow.cs
paulhazen: Marking this as resolved because updating this PR with `development` put this change out of the PR (and doing so resolved the issue).

Assets/Plugins/Source/Core/EOSOnConnectLogin.cs
paulhazen: This one is a little confusing to me to be honest. 

The way `EOSAchievementsManager` _currently_ caches the data is by using a `ConcurrentDictionary` that maps a `ProductUserId` type key to a `List<Stats>` type value. What's odd (to me) is that implies that more than one `ProductUserId` might be present - in which case storing the `ProductUserId` instance for the _manager_ might make sense. Like - this would store the `ProductUserId` that is _active_ when the manager is "logged in."

I'm a huge fan of reducing the number of sources of truth to be a singularity - almost to a fault. 

So, clarity on this would probably be good to get. @andrew-hirata-playeveryware Do you have insight you can lend?

Assets/Scripts/EOSStatsManager.cs
paulhazen: Yeah - super helpful! :)

Assets/Scripts/EOSStatsManager.cs
paulhazen: While I am a fan of doing something like that - this PR has the goal of just transplanting the functionality that exists within `EOSAchievementsManager` that pertain to the `Stats Interface` as Epic defines it. Which means expanding upon, or altering the implementation should be considered out of scope for this PR - since these lines were copied over.

Assets/Scripts/EOSStatsManager.cs
paulhazen: See [my response](https://github.com/PlayEveryWare/eos_plugin_for_unity/pull/715#discussion_r1635397588) to your other comment.

Assets/Plugins/Source/Core/EOSOnConnectLogin.cs
andrew-hirata-playeveryware: The short answer is that we shouldn't store the _productUserID in this class as the EOS Manager should be the 'owner' of the product user ID.

paulhazen:
Closing PR - restarting on a new branch.

PR #716, opened 2024-06-11
Threaded File Copy (aka ~5x Speed Boost in Package Creation)

paulhazen:
This PR depends upon the approval and subsequent merge of PR #714, since by happen-stance these changes were based off of that branch. (This line of description will be removed when the PR is moved from draft to open following the merge of aforementioned PR).

Package creation for this project boils down to a series of files being copied from one directory to another. Turns out that (including the changes from PR #714), the matter of multi-threading this process is _very_ __very__ simple. Additionally: doing so reduces package creation time from ~2.5 minutes to ~30 seconds - which is _phenomenal_ for both testing and iterating.


Assets/Plugins/Source/Editor/Windows/CreatePackageWindow.cs
```diff
@@ -301,7 +301,7 @@ private async void StartCreatePackageAsync(UPMUtility.PackageType type)
                 {
                     // Ternary statement here to prevent a divide by zero problem
                     // ever happening, despite how odd it would be in this case.
-                    float newActualProgress = (0.0f >= value.TotalBytesToCopy)
+                    float newActualProgress = (0.0f <= value.TotalBytesToCopy)
```
WispyMouse: value.TotalBytesToCopy can be less than or equal to 0. If true, then it's going to try to divide by zero. Was this boolean logic flip intentional?

Assets/Plugins/Source/Editor/Utility/PackageFileUtility.cs
```diff
@@ -208,20 +248,22 @@ private static IEnumerable<FileInfoMatchingResult> FindMatchingFiles(string root
...
             Directory.CreateDirectory(destination);
 
             GetFileSystemOperations(
                 destination, 
                 matchingResults,
                 out List<FileUtility.CopyFileOperation> copyOperations);
 
+            if (0 == copyOperations.Count)
+            {
+                Debug.LogWarning("There were no files that need to be moved to create the package (nothing seems to have changed).");
```
WispyMouse: If there are no things to copy, maybe we can exit out here instead of continuing to the await?

WispyMouse (approved):

Assets/Plugins/Source/Editor/Windows/CreatePackageWindow.cs
paulhazen: This got fixed via the previous PR being merged into `development`, then `development` being subsequently merged back into this branch.

Assets/Plugins/Source/Editor/Utility/PackageFileUtility.cs
paulhazen: Because a function that is defined with the `async` keyword _must_ return a type of `Task`, we could not just stop here, we would need to return an empty `Task`, which I thought would be more confusing that letting the `await` keyword accomplish that behavior for us.

paulhazen: Marking this as resolved because these changes actually got merged in the other PR, and are no longer a part of _this_ PR.

PR #720, opened 2024-06-13
fix: Make `Config.Get` and `Config.GetAsync` compatible with Android

paulhazen:
This PR addresses an issue where (when running on the Android platform) the `Config.Get<T>` and `Config.GetAsync<T>` functions cause a crash. The problem was that ultimately, those code execution paths were not making use of our `FileUtility` class - which accounts for the oddities of the Android platform to deal with this appropriately.

Kudos to @arthur740212 for discovering this issue in the process of his work making our Logging configuration better!

### DIFF
--- a/Assets/Plugins/Source/Core/Config.cs
+++ b/Assets/Plugins/Source/Core/Config.cs
@@ -33,6 +33,7 @@ namespace PlayEveryWare.EpicOnlineServices
     using System.Text;
     using JsonUtility = PlayEveryWare.EpicOnlineServices.Utility.JsonUtility;
     using System.Runtime.CompilerServices;
+    using Utility;
 
     /// <summary>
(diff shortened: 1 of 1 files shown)


Assets/Plugins/Source/Core/Config.cs
```diff
@@ -269,8 +270,7 @@ protected virtual async Task ReadAsync()
 #endif
             }
 
-            using StreamReader reader = new(FilePath);
-            _lastReadJsonString = await reader.ReadToEndAsync();
+            _lastReadJsonString = await FileUtility.ReadAllTextAsync(FilePath);
```
paulhazen, line 273: Make sure to update the `ReadAllTextAsync` to include Android behavior.

andrew-hirata-playeveryware (approved):

arthur740212 (approved):

PR #721, opened 2024-06-13
fix: Enable Recursive File Selection during Package Creation

paulhazen:
## Overview
For a long time, we have had the `recursive` field member on `SrcDestPair`, but it was neither utilized, nor implemented to function properly.

This PR fixes that. When an entry in `eos_package_description.json` is encountered that has `recursive` set to `true`, the package creation process will encounter that, and when ingesting the list of files to copy it will add new `SrcDestPair` objects to the list to accomplish the recursive nature of the file selection.

## Impact
Implementing this means we are able to further reduce the size of our `eos_package_description.json` file, while greatly increasing how easy it is to visually inspect the file to see what is happening when a package is created. 

## Testing Considerations
Tests have been run to ensure that the package created contains the same files as the previous implementation and functions the same way when imported into a fresh Unity project.

One concern that has been brought up in the course of code reviews is that utilizing the recursive feature makes it possible that console specific code - code that cannot be made public - could errantly be introduced. This concern is addressed by introducing entries to `eos_package_description.json` that explicitly ignore any such code - making the changes here no more or less likely to cause such a mistake to take place.


etc/PackageConfigurations/eos_package_description.json
paulhazen, line 1: Check for EOS_SDK/Generated/ being done correctly.

paulhazen: This has been corrected to explicitly ignore any console-specific code.

WispyMouse (approved):
Packaging appears to work. My only concern is order of operations, that it's possible to move a /* folder and then afterwards try to move something specific under that folder.

Neat syntax, had never seen some of what you've done with strings to arrays to strings.

etc/PackageConfigurations/eos_package_description.json
```diff
@@ -1,520 +1,118 @@
...
-        {"src":"Assets/Plugins/Source/EOS_SDK/Generated/CustomInvites.meta", "dest": "Runtime/EOS_SDK/Generated/"},
-        {"src":"Assets/Plugins/Source/EOS_SDK/Generated/CustomInvites/*.cs", "dest": "Runtime/EOS_SDK/Generated/CustomInvites/"},
-        {"src":"Assets/Plugins/Source/EOS_SDK/Generated/CustomInvites/*.cs.meta", "dest": "Runtime/EOS_SDK/Generated/CustomInvites/"},
-
-        {"src":"Assets/Plugins/Source/EOS_SDK/Generated/Ecom.meta", "dest": "Runtime/EOS_SDK/Generated/"},
-        {"src":"Assets/Plugins/Source/EOS_SDK/Generated/Ecom/*.cs", "dest": "Runtime/EOS_SDK/Generated/Ecom/"},
-        {"src":"Assets/Plugins/Source/EOS_SDK/Generated/Ecom/*.cs.meta", "dest": "Runtime/EOS_SDK/Generated/Ecom/"},
-
-        {"src":"Assets/Plugins/Source/EOS_SDK/Generated/Friends.meta", "dest": "Runtime/EOS_SDK/Generated/"},
-        {"src":"Assets/Plugins/Source/EOS_SDK/Generated/Friends/*.cs", "dest": "Runtime/EOS_SDK/Generated/Friends/"},
```
paulhazen, line 284: List each of the entries within the `Generated` directory.

etc/PackageConfigurations/eos_package_description.json
```diff
@@ -1,520 +1,118 @@
...
-
-        {"src":"Assets/Readme/*.txt", "dest": "Samples~/Samples/StandardSamples/Readme/"}
+        {"src":"Assets/Plugins/Standalone/Editor/*",                             "dest": "Editor/Standalone/"},
+        {"src":"Assets/Plugins/Standalone/Editor/anticheat_integritytool.cfg",   "dest": "Editor/Standalone/"},
+
+        {"src":"tools/bin/*",                                                    "dest": "bin~/", "recurse": true},
+
+        {"src":"Assets/link.xml",                                                "dest": "Editor/"},
+
+        {"src":"Assets/Plugins/Source/EOS_SDK/*",                                "dest": "Runtime/EOS_SDK/", "recurse": true },
```
paulhazen, line 55: Remove the `recurse` from this for the time being - instead explicitly ask to include each of the directories inside the `Generated` directory.

etc/PackageConfigurations/eos_package_description.json
```diff
@@ -1,520 +1,118 @@
...
+        
+        {"ignore_regex":"Assets/Plugins/Source/EOS_SDK/Generated/PS4"},
+        {"ignore_regex":"Assets/Plugins/Source/EOS_SDK/Generated/PS5"},
+        {"ignore_regex":"Assets/Plugins/Source/EOS_SDK/Generated/XboxOneGDK"},
+        {"ignore_regex":"Assets/Plugins/Source/EOS_SDK/Generated/XSX"},
+        {"ignore_regex":"Assets/Plugins/Source/EOS_SDK/Generated/Switch"},
+
+        {"ignore_regex":"Assets/Plugins/Source/EOS_SDK/PewExtension/GDK"},
+        {"ignore_regex":"Assets/Plugins/Source/EOS_SDK/PewExtension/PS4"},
```
paulhazen, line 64: Also remove these so we are not explicitly mentioning these platforms.

andrew-hirata-playeveryware (approved):

PR #722, opened 2024-06-13
fix: Correct implementation of `ReadAllTextAsync` for Android

paulhazen:
This PR fixes an issue where the wrong IO method is used when reading a fily asynchronously on the Android platform.


arthur740212 (commented):
Does the return await work if Android reader is not an async function?

paulhazen:
> Does the return await work if Android reader is not an async function?

Yes - when the Android reader is used the method from `AndroidIOHelper` is used, and the resulting string is returned using `await Task.FromResult(text);`

arthur740212 (approved):

andrew-hirata-playeveryware (approved):

arthur740212 (approved):

PR #724, opened 2024-06-14
fix(mac) : Update burst to resolve persistent login crashing issue

arthur740212:
Update Burst version to 1.18.8 which is latest that all platforms support.


andrew-hirata-playeveryware (approved):

paulhazen (approved):
You should add yourself as an assignee to this PR

PR #726, opened 2024-06-15
Steam login with `GetAuthTicketForWebApi `

### LINKED ISSUE #668: Steam Manager does not use recommended ticket
Epic documentation states that Steam Auth login should be performed using a ticket from Steam's **GetAuthTicketForWebApi** call.

See here: https://dev.epicgames.com/docs/dev-portal/identity-provider-management#steam

However the PEW Steam Manager / samples do not use this type of ticket (they instead use SteamUser.GetAuthSessionTicket)

arthur740212:
This will be the only way to login in with steam session tickets in the future, 
according to https://dev.epicgames.com/docs/dev-portal/identity-provider-management#steam

GetAuthTicketForWebApi is async and requires a callback.

Current state : 
First login will attempt to grab the ticket.
Will log a message when the ticket is retrieved.
Press login button again to successfully login.

Todo : 
Support custom network identity with configs

### DIFF
--- a/Assets/Scripts/Steam/SteamManager.cs
+++ b/Assets/Scripts/Steam/SteamManager.cs
@@ -63,7 +63,7 @@ namespace PlayEveryWare.EpicOnlineServices.Samples.Steam
         HAuthTicket sessionTicketHandle = HAuthTicket.Invalid;
         string sessionTicketString = null;
 
-        Callback<GetTicketForWebApiResponse_t> m_AuthTicketForWebApiResponseCallback;
+        Callback<GetTicketForWebApiResponse_t> authTicketForWebApiResponseCallback { get; set; }
 
         CallResult<EncryptedAppTicketResponse_t> appTicketCallResult = new CallResult<EncryptedAppTicketResponse_t>();
         private event Action<string> appTicketEvent;
@@ -151,16 +151,20 @@ namespace PlayEveryWare.EpicOnlineServices.Samples.Steam
                 SteamClient.SetWarningMessageHook(m_SteamAPIWarningMessageHook);
             }
 
-            if (m_AuthTicketForWebApiResponseCallback == null)
+            if (authTicketForWebApiResponseCallback == null)
             {
-                m_AuthTicketForWebApiResponseCallback = Callback<GetTicketForWebApiResponse_t>.Create(
+                authTicketForWebApiResponseCallback = Callback<GetTicketForWebApiResponse_t>.Create(
                     (GetTicketForWebApiResponse_t pCallback) =>
                     {
                         if (pCallback.m_eResult == EResult.k_EResultOK)
                         {
                             sessionTicketHandle = pCallback.m_hAuthTicket;
                             sessionTicketString = System.BitConverter.ToString(pCallback.m_rgubTicket).Replace("-", "");
-                            Debug.LogWarning ("GetAuthTicketForWebApi succeed, try logging in again");
+                            print("GetAuthTicketForWebApi succeed, try logging in again");
+                        }
+                        else 
+                        {
+                            Debug.LogError($"GetAuthTicketForWebApi Result : {pCallback.m_eResult}");
                         }
                     });
             }
@@ -339,7 +343,14 @@ namespace PlayEveryWare.EpicOnlineServices.Samples.Steam
                 return sessionTicketString;
             }
 
-#if GET_STEAM_SESSION_TICKET_LEGACY
+            // GetAuthTicketForWebApi will become the only EOS supported way to grab session tickets in the future
+            // https://dev.epicgames.com/docs/dev-portal/identity-provider-management#steam
+            // For now (2024/06/19) it still used the old way to grab session tickets
+            // Manually define GET_SESSION_TICKET_FOR_WEB_API to use the new version
+#if GET_SESSION_TICKET_PREVIEW
+            SteamUser.GetAuthTicketForWebApi(null);
+            sessionTicketString = "";
+#else
             int bufferSize = 1024;
             byte[] buffer = new byte[bufferSize];
             uint ticketSize = 0;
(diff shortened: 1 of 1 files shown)


Assets/Scripts/Steam/SteamManager.cs
```diff
@@ -323,6 +339,7 @@ public string GetSessionTicket()
                 return sessionTicketString;
             }
 
+#if GET_STEAM_SESSION_TICKET_LEGACY
```
paulhazen: We should add or include documentation making clear what this scripting define does (ie why do we have this conditional here?)

Assets/Scripts/Steam/SteamManager.cs
```diff
@@ -63,6 +63,8 @@ protected static void SteamAPIDebugTextHook(int nSeverity, System.Text.StringBui
         HAuthTicket sessionTicketHandle = HAuthTicket.Invalid;
         string sessionTicketString = null;
 
+        Callback<GetTicketForWebApiResponse_t> m_AuthTicketForWebApiResponseCallback;
```
paulhazen: While the rest of this file makes use in some cases of the prefix `m_`, adding _new_ things within this project we should make sure that new additions match the overall project style guidelines (at least in this case).

paulhazen (changes requested):
Requesting changes on some minor code formatting and documentation things - but I am uncertain about the nature of half-way addressing the linked issue - I'd like to understand better whether or not a complete solution can be accomplished - and why or why not in either case.

Assets/Scripts/Steam/SteamManager.cs
WispyMouse: Because this is a MonoBehaviour, I really want everything to have accessors. Otherwise Unity might do something funny and serialize the value of this, even though it is private. Do you think that applies here?

Assets/Scripts/Steam/SteamManager.cs
```diff
@@ -148,6 +150,20 @@ protected virtual void OnEnable()
...
+            if (m_AuthTicketForWebApiResponseCallback == null)
+            {
+                m_AuthTicketForWebApiResponseCallback = Callback<GetTicketForWebApiResponse_t>.Create(
+                    (GetTicketForWebApiResponse_t pCallback) =>
+                    {
+                        if (pCallback.m_eResult == EResult.k_EResultOK)
+                        {
+                            sessionTicketHandle = pCallback.m_hAuthTicket;
+                            sessionTicketString = System.BitConverter.ToString(pCallback.m_rgubTicket).Replace("-", "");
+                            Debug.LogWarning ("GetAuthTicketForWebApi succeed, try logging in again");
```
WispyMouse: Should this use one of our conditional print functions instead of Unity's logging directly?

Assets/Scripts/Steam/SteamManager.cs
```diff
@@ -148,6 +150,20 @@ protected virtual void OnEnable()
                 m_SteamAPIWarningMessageHook = new SteamAPIWarningMessageHook_t(SteamAPIDebugTextHook);
                 SteamClient.SetWarningMessageHook(m_SteamAPIWarningMessageHook);
             }
+
+            if (m_AuthTicketForWebApiResponseCallback == null)
+            {
+                m_AuthTicketForWebApiResponseCallback = Callback<GetTicketForWebApiResponse_t>.Create(
+                    (GetTicketForWebApiResponse_t pCallback) =>
+                    {
+                        if (pCallback.m_eResult == EResult.k_EResultOK)
```
WispyMouse, line 159: If the result is not a success, should we log out an error message?

WispyMouse (changes requested):
I'd like to see some additional logging changes.

Assets/Scripts/Steam/SteamManager.cs
arthur740212: print() logs a log not a log warning.
I made this warning to stand out (warnings are yellow), so user knows the ticket is ready to try logging in again.
We decided that this two-step login will need to be changed.
So this line will probably disappear.

Assets/Scripts/Steam/SteamManager.cs
arthur740212: Just learned a lot about serialization, after a crash course with Mint.

SteamManager only gets created a singleton-ish instance in runtime, and there are no other references in scenes, so there should be nothing getting serialized here.
Also `Callback` here is an object from Steamworks, which Unity wouldn't be able to serialize.

Assets/Scripts/Steam/SteamManager.cs
```diff
@@ -126,8 +129,7 @@ private void OnApplicationQuit()
             }
         }
 
-
-
+        TaskCompletionSource<string> tcs;
```
andrew-hirata-playeveryware: This should be moved to near the definition of the callback. and tcs should be renamed to something like authTicketForWebAPIResponseTaskCompletionSource or something shorter if you can think of it.

Assets/Scripts/Steam/SteamManager.cs
```diff
@@ -312,7 +332,29 @@ public string GetSteamID()
...
+        {
+#if DISABLESTEAMWORKS
+            return null;
+#else
+            // GetAuthTicketForWebApi will become the only EOS supported way to grab session tickets in the future
+            // https://dev.epicgames.com/docs/dev-portal/identity-provider-management#steam
+            // For now (2024/06/19) it still used the old way to grab session tickets
+            // Manually define GET_SESSION_TICKET_FOR_WEB_API to use the new version
+    #if ASYNC_GET_STEAM_SESSION_TICKET_PREVIEW
+            //if (sessionTicketString != null)
```
andrew-hirata-playeveryware: If this isn't needed, we should delete it.

andrew-hirata-playeveryware (changes requested):

WispyMouse (approved):

andrew-hirata-playeveryware (approved):

paulhazen (approved):

PR #728, opened 2024-06-17
Introduction of `StatsManager`

paulhazen:
## Overview

This PR introduces the `StatsManager`, which handles all the functionality that currently exists across the different sample scenes pertaining to interaction with the Stats Interface of the EOS SDK.

### Abstract `ServiceManager` class

Functionality that is common to both `StatsManager`, `EOSAchievementManager` and `EOSLeaderboardManager` is contained within this class.

### Async

Many of the functions within both `EOSAchievementManager` and `StatsManager` have been moved from being callback-oriented to being wrapped such that they can be more easily called utilizing the async / await pattern. This does little to significantly impact the API of either manager other than to simplify both.


Assets/Scripts/EOSAchievementManager.cs
```diff
@@ -22,29 +22,31 @@
...
-using System.Collections.Concurrent;
-using Epic.OnlineServices.Achievements;
-using Epic.OnlineServices.Stats;
-using System;
-using Epic.OnlineServices;
-using System.Diagnostics;
-
 namespace PlayEveryWare.EpicOnlineServices.Samples
 {
+    using System;
```
WispyMouse, line 27: stylistic non-criticism: I've always liked putting using statements under the namespace declaration, like this. Mint will bring this up as a stylistic thing for the team to consider doing.

Assets/Scripts/EOSAchievementManager.cs
```diff
@@ -205,23 +119,27 @@ public void OnConnectLogin(Epic.OnlineServices.Connect.LoginCallbackInfo loginCa
...
-            });
+
+            // NOTE: Because there is no check in the above code to determine if
+            //       any achievements have actually changed, EVERY call to this
+            //       Refresh function will trigger a data updated callback to
+            //       all listeners, regardless of whether anything has changed.
+            //       Possible improvement would be to evaluate each achievement
+            //       definition, and the achievement for each player and only 
+            //       trigger this callback if a change is detected. However,
+            //       comparing equality of the PlayerAchievement struct (defined
```
WispyMouse, line 148: nit: There is an unclosed parenthesis in this comment

Assets/Scripts/EOSAchievementManager.cs
```diff
@@ -205,23 +119,27 @@ public void OnConnectLogin(Epic.OnlineServices.Connect.LoginCallbackInfo loginCa
         /// to any player for whom that information has been requested
         /// previously.
         /// </summary>
-        public void Refresh()
+        public async override void Refresh()
```
WispyMouse: What is the reason for preferring `async void` signatures instead of `async Task` signatures?

Assets/Scripts/EOSAchievementManager.cs
```diff
@@ -205,23 +119,27 @@ public void OnConnectLogin(Epic.OnlineServices.Connect.LoginCallbackInfo loginCa
...
-            QueryAchievements(productUserId, (ref OnQueryDefinitionsCompleteCallbackInfo defQueryData) =>
-            {
-                _achievements = GetCachedAchievements();
-                foreach (var userId in _playerStats.Keys)
+            _achievements = await QueryAchievementsAsync(productUserId);
+            
+                foreach (var userId in _playerAchievements.Keys)
                 {
-                    RefreshPlayerStatsAndAchievements(userId);
+                    RefreshPlayerAchievementsAsync(userId);
```
WispyMouse: Is there a reason we're not awaiting a Task for each of these? This _looks_ like something that could create race conditions or inconsistent results.

Assets/Scripts/EOSAchievementManager.cs
```diff
@@ -233,19 +151,10 @@ public void Refresh()
...
-                _playerStats[productUserId] = GetCachedPlayerStats(productUserId);
-            });
-
-            QueryPlayerAchievements(productUserId, (ref OnQueryPlayerAchievementsCompleteCallbackInfo achievementsInfo) =>
-            {
-                _playerAchievements[productUserId] = GetCachedPlayerAchievements(productUserId);
-            });
-
-            NotifyAchievementDataUpdated();
+            _playerAchievements[productUserId] = await QueryPlayerAchievementsAsync(productUserId);
```
WispyMouse: It _feels_ like we should use something like `AddOrUpdate`, but it's not clear to me if that actually provides significant benefit over what you're doing here.

Assets/Scripts/EOSAchievementManager.cs
```diff
@@ -275,22 +184,61 @@ private static AchievementsInterface GetEOSAchievementInterface()
         /// Invoked when the request from the server has completed - whether
         /// successful or not.
         /// </param>
-        private static void QueryAchievements(ProductUserId productUserId, OnQueryDefinitionsCompleteCallback callback)
+        private Task<List<DefinitionV2>> QueryAchievementsAsync(ProductUserId productUserId)
```
WispyMouse, line 202: Should this have the "async" decorator? Maybe not, because this is managed through setting a task completion source?

Assets/Scripts/EOSAchievementManager.cs
```diff
@@ -275,22 +184,61 @@ private static AchievementsInterface GetEOSAchievementInterface()
...
+            {
+                AchievementIndex = 0
+            };
+
+            List<DefinitionV2> achievements = new();
+
+            for (uint i = 0; i < achievementDefinitionCount; ++i)
+            {
+                options.AchievementIndex = i;
+                GetEOSAchievementInterface().CopyAchievementDefinitionV2ByIndex(ref options, out DefinitionV2? definition);
```
WispyMouse: Should we observe the `Result` returned from this?

Assets/Scripts/EOSAchievementManager.cs
```diff
@@ -275,22 +184,61 @@ private static AchievementsInterface GetEOSAchievementInterface()
...
+
+            return tcs.Task;
+        }
+
+        /// <summary>
+        /// Retrieves the list of achievement definitions for the game that have
+        /// been cached locally.
+        /// </summary>
+        /// <returns>A list of achievement definitions.</returns>
+        private List<DefinitionV2> GetCachedAchievements()
```
WispyMouse, line 229: not useful criticism: It somewhat surprises me that I could call a method called "GetCachedAchievements", but it has to rebuild the list each time. I really wish we could know if this work needed to be re-done after its first calling. Thing is, I don't think Epic has provided a way of knowing that the provided achievements list is a different version than the past call...

Assets/Scripts/EOSAchievementManager.cs
```diff
@@ -275,22 +184,61 @@ private static AchievementsInterface GetEOSAchievementInterface()
...
+                options.AchievementIndex = i;
+                GetEOSAchievementInterface().CopyAchievementDefinitionV2ByIndex(ref options, out DefinitionV2? definition);
+
+                // Move on if the definition is empty
+                if (!definition.HasValue)
+                    continue;
+
+                achievements.Add(definition.Value);
+
+                UnityEngine.Debug.LogFormat("Achievements (CacheAchievementDef): Id={0}, LockedDisplayName={1}", definition.Value.AchievementId, definition.Value.LockedDisplayName);
```
WispyMouse: Should this use the `print` functions so that it can be conditionally excluded from prod builds?

Assets/Scripts/EOSAchievementManager.cs
```diff
@@ -301,34 +249,35 @@ private static void QueryAchievements(ProductUserId productUserId, OnQueryDefini
...
         /// </param>
-        /// <param name="callback">
-        /// Invoked when the request from the server has completed - whether
-        /// successful or not.
-        /// </param>
-        private static void QueryPlayerAchievements(ProductUserId productUserId, OnQueryPlayerAchievementsCompleteCallback callback)
+        /// <returns>
+        /// List of PlayerAchievement objects.
+        /// </returns>
+        private Task<List<PlayerAchievement>> QueryPlayerAchievementsAsync(ProductUserId productUserId)
```
WispyMouse, line 277: If the above Task function should be async, this should also be async. Otherwise, ignore this comment.

Assets/Scripts/EOSAchievementManager.cs
```diff
@@ -489,21 +438,31 @@ public static uint GetAchievementsCount()
         /// <param name="achievementId">
         /// The id of the achievement to unlock for the current player.
         /// </param>
-        /// <param name="callback">
-        /// Invoked upon the completion of the attempt to unlock the achievement
-        /// for the given player.
-        /// </param>
-        public void UnlockAchievement(string achievementId, OnUnlockAchievementsCompleteCallback callback)
+        public Task UnlockAchievementAsync(string achievementId)
```
WispyMouse, line 463: Similar comment wrt `async Task`.

Assets/Scripts/ServiceManagers/ServiceManager.cs
```diff
@@ -0,0 +1,117 @@
...
+        /// Implementation of this function should refresh the locally cached
+        /// data managed by the service manager in question.
+        /// </summary>
+        public abstract void Refresh();
+
+        /// <summary>
+        /// Trigger the notification callbacks that have been registered,
+        /// letting any consumers know that the data base been updated.
+        /// </summary>
+        protected void NotifyUpdated()
```
WispyMouse, line 143: Is there ever a world where `_updateCallbacks` has a callback, and during `action()` being called it calls `RemoveUpdateCallback`, modifying this list? That'd break because of the foreach not liking its collection being modified while enumerating.

Assets/Scripts/ServiceManagers/StatsManager.cs
```diff
@@ -0,0 +1,249 @@
...
+
+            TaskCompletionSource<List<Stat>> tcs = new();
+
+            statInterface.QueryStats(ref statsOptions, null, (ref OnQueryStatsCompleteCallbackInfo queryStatsCompleteCallbackInfo) =>
+            {
+                if (queryStatsCompleteCallbackInfo.ResultCode != Result.Success)
+                {
+                    // TODO: handle error
+                    Log($"Failed to query stats, result code: {queryStatsCompleteCallbackInfo.ResultCode}");
+                    tcs.SetResult(new List<Stat>());
```
WispyMouse: Do you prefer that to the setting exception done in some other places?

Assets/Scripts/EOSAchievementManager.cs
```diff
@@ -55,17 +57,33 @@ public class EOSAchievementManager : IEOSSubManager, IEOSOnConnectLogin
...
+
+        #region Singleton Implementation
+
         /// <summary>
-        /// Maps a given user to a list of player statistics.
+        /// Lazy instance for singleton allows for thread-safe interactions with
+        /// the EOSAchievementManager
         /// </summary>
-        private IDictionary<ProductUserId, List<Stat>> _playerStats = new Dictionary<ProductUserId, List<Stat>>();
+        private static readonly Lazy<EOSAchievementManager> s_LazyInstance = new(() => new EOSAchievementManager());
```
WispyMouse, line 67: This is a really clever singleton method. Should we consider using this, over the previous `GetOrCreateManager` method?

Assets/Scripts/ServiceManagers/ServiceManager.cs
WispyMouse: Or alternatively while this is running, it somehow causes something to add to the list of callbacks. That'd also break it.

Assets/Scripts/UI/Achievements/UIAchievementsMenu.cs
```diff
@@ -134,48 +131,35 @@ public void HideMenu()
...
-                {
-                    Debug.Log("UnlockAchievement Succeed"); 
-                    achievementManager.Refresh();
-                }
-            });
+                await EOSAchievementManager.Instance.UnlockAchievementAsync(definition.AchievementId);
+            }
+            catch (Exception e)
+            {
+                Debug.Log(e.Message);
```
WispyMouse: Should this either use the print method, or perhaps log at error or warning?

WispyMouse:
I went to the Achievements sample,
logged in using PersistentAuth,
and upon landing on the main page, there were no achievements displayed.
When I pressed "Referesh Data", the achievements then appeared.
Is it intended that the achievements are visible immediately upon logging in to the sample?

WispyMouse:
It would seem every time I press `Increase Login Count` in the sample, it now needs to refresh the images of the achievements. That feels like it's extra work that isn't strictly necessary. Maybe this suggests the pictures aren't being cached or something?

WispyMouse:
When attempting to look at my achievement info, there's a message of "Player achievement info not found." I'm pointing at the correct deployment, and looking at the Epic back end I'm incrementing my login count each time I press Increase Login Count. The sample does not reflect that I have earned the login count achievement, still showing the X.

It could be that I'm configured incorrectly at the moment, but perhaps there's more there?

WispyMouse:
Let's treat this comment as the hub for this question:

Why is it that some functions have Task return type without the async signature, and some methods have the async signature with the void return type? Is there a rhyme/reason to the significance of when each of these are applied? Would things be better/worse/thesame/different if we turned every one of these methods in to an async Task?

WispyMouse:
Something I'm concerned about in the change to using async is making sure everything makes it back to the main thread when it's done. I'm theorizing a situation where someone says "after a refresh, change this label", and then Unity getting upset because that is only allowed to happen on the main thread.

Assets/Scripts/EOSAchievementManager.cs
paulhazen: I concur.

Assets/Scripts/EOSAchievementManager.cs
paulhazen: It should not have an `async` decorator, because nothing is `await`ed within the function. The function _itself_ is awaitable.

Assets/Scripts/EOSAchievementManager.cs
paulhazen: Good catch, fixed.

Assets/Scripts/EOSAchievementManager.cs
paulhazen: This is a good catch. `async void` functions cannot be awaited, whereas `async Task` functions can - moreover they can more properly handle exceptions called from the async context. The default should be `async Task` over `async void`. The latter should really only be used when the function is handling an event.

This instance has been corrected.

Assets/Scripts/EOSAchievementManager.cs
paulhazen: I have updated the implementation to utilize `await` for each of these calls.

Assets/Scripts/EOSAchievementManager.cs
paulhazen: Excellent observation - `AddOrUpdate` is much more appropriate in this context because it is thread-safe, which is the main purpose behind using `ConcurrentDictionary`. I have updated the implementation here, and in the function within `StatsManager` that updates the player stats.

Assets/Scripts/EOSAchievementManager.cs
paulhazen: Yep - another good catch. Implementation has been updated accordingly.

Assets/Scripts/EOSAchievementManager.cs
paulhazen: Yes - that is a valid critique of the EOS SDK in my opinion. That the SDK itself caches the achievements though I believe is a by-product of anti-cheat implementation - not necessarily a way to prevent the application from mis-representing achievements, but as a way to encourage use of _their_ cache as the source of truth over whatever locally cached version one might have made oneself.

Assets/Scripts/EOSAchievementManager.cs
paulhazen: Good idea. I've updated it to do so.

Assets/Scripts/EOSAchievementManager.cs
paulhazen: As before - does not need to be async, as it makes no await calls within it. This does not prevent it from being `await` called.

Assets/Scripts/EOSAchievementManager.cs
paulhazen: Same as above - does not need to be async, as it makes no await calls within it. This does not prevent it from being called with the `await` keyword.

Assets/Scripts/ServiceManagers/ServiceManager.cs
paulhazen: Because `NotifyUpdated` is _not_ async, and because none of the functions that update it are asynchronous, the list will not (as far as I understand things) be updated from another thread.

Assets/Scripts/ServiceManagers/StatsManager.cs
paulhazen: Good call - so I updated it to _not_ call the `SetResult` function, but instead call the `SetException` function passing an exception with a message equal to what was previously sent to the `Log` function.

Assets/Scripts/EOSAchievementManager.cs
paulhazen: Ultimately, yes, we should do that. However, I wanted to limit the scope of the change to only the manager classes that were being updated pursuant to the changes that this PR is intended to encompass. As other managers get refactored, we can make those changes - so the change happens more gradually.

Assets/Scripts/UI/Achievements/UIAchievementsMenu.cs
paulhazen: The `UIAchievementsMenu` class does not use the `print` function. I did however update it to log an error instead of just a standard log.

paulhazen:
> Let's treat this comment as the hub for this question:
> 
> Why is it that some functions have Task return type without the async signature, and some methods have the async signature with the void return type? Is there a rhyme/reason to the significance of when each of these are applied? Would things be better/worse/thesame/different if we turned every one of these methods in to an async Task?

This has been addressed on a per-instance basis for each of the places it was brought up.

paulhazen:
> Something I'm concerned about in the change to using async is making sure everything makes it back to the main thread when it's done. I'm theorizing a situation where someone says "after a refresh, change this label", and then Unity getting upset because that is only allowed to happen on the main thread.

I am not certain how to address this. The purpose of the `async` / `await` pattern is to accomplish this - that is, to perform some task and "wait" for it to be done before doing something else.

paulhazen:
> It would seem every time I press `Increase Login Count` in the sample, it now needs to refresh the images of the achievements. That feels like it's extra work that isn't strictly necessary. Maybe this suggests the pictures aren't being cached or something?

Is this something that you are observing visually, or are you stepping through the code and observing that the image data is not being cached?

WispyMouse:
> > It would seem every time I press `Increase Login Count` in the sample, it now needs to refresh the images of the achievements. That feels like it's extra work that isn't strictly necessary. Maybe this suggests the pictures aren't being cached or something?
> 
> Is this something that you are observing visually, or are you stepping through the code and observing that the image data is not being cached?

Observing visually. Click the button, see the flicker.

Assets/Scripts/UI/Achievements/UIAchievementsMenu.cs
```diff
@@ -79,17 +77,16 @@ private void Awake()
             achievementListItems = new List<UIAchievementButton>();
 
             HideMenu();
-            achievementManager = EOSManager.Instance.GetOrCreateManager<EOSAchievementManager>();
```
paulhazen, line 82: Explain why _not_ calling this method to get the manager is not a problem - because it does some hookup work for us, and it's concerning that is now circumvented by the singleton pattern.

Assets/Scripts/EOSAchievementManager.cs
WispyMouse: `await Task.WhenAll` is a great use of the C# API here, I had just thought about awaiting for each but this is much better and more thread safe.

Assets/Scripts/EOSAchievementManager.cs
WispyMouse: I was under the impression that all awaitable things needed to be asynchronous, but doing some reading it appears as though it is unnecessary in some situations, including this situation here. So good to go.

Assets/Scripts/EOSAchievementManager.cs
WispyMouse: Copy that, you've got it right here.

Assets/Scripts/EOSAchievementManager.cs
WispyMouse: Yup, you've got it right, this doesn't need to be async.

Assets/Scripts/ServiceManagers/ServiceManager.cs
WispyMouse: It doesn't have to be another thread. What if one of the callbacks itself results in modifying this list?

WispyMouse (approved):

Assets/Scripts/ServiceManagers/ServiceManager.cs
paulhazen: Excellent point! So, I changed the implementation so that before iterating over the callback list, a copy of the list is made, and the copy is iterated over instead.

paulhazen:
> I went to the Achievements sample, logged in using PersistentAuth, and upon landing on the main page, there were no achievements displayed. When I pressed "Referesh Data", the achievements then appeared. Is it intended that the achievements are visible immediately upon logging in to the sample?

That is the intended behavior - and the code has been updated to make that happen.

Assets/Scripts/UI/Achievements/UIAchievementsMenu.cs
paulhazen: Not calling this method is acceptable, because (now) the base implementation within `ServiceManager` is to add itself as a connect listener, and on dispose remove itself. @andrew-hirata-playeveryware for viz.

paulhazen:
> > > It would seem every time I press `Increase Login Count` in the sample, it now needs to refresh the images of the achievements. That feels like it's extra work that isn't strictly necessary. Maybe this suggests the pictures aren't being cached or something?
> > 
> > 
> > Is this something that you are observing visually, or are you stepping through the code and observing that the image data is not being cached?
> 
> Observing visually. Click the button, see the flicker.

This happens because of the implementation of `UIAchievementsMenu` - whenever refresh is called the buttons are actually _recreated_ which causes the blinking.

Assets/Scripts/ServiceManagers/ServiceManager.cs
```diff
@@ -0,0 +1,215 @@
...
+ *
+ * Author: Paul Hazen (paul.hazen@playeveryware.com)
+ *
+ * Date:   2024-06-17
+ *
+ * Notes: This file contains implementation of functionality that is common to
+ *        many of the different managers. Not all managers make use of it - as
+ *        transitioning will be done gradually and as areas are refactored.
+ *        Consequently, only EOSAchievementManager and StatsManager make use of
+ *        this base class.
```
paulhazen: Please remove these - we only want to have the license at the top of the file. Ownership and other administrative tasks should be manage outside the scope of the code file.

Assets/Scripts/ServiceManagers/ServiceManager.cs
```diff
@@ -0,0 +1,215 @@
...
+
+        /// <summary>
+        /// Determines whether the user is currently authenticated using the
+        /// Auth Interface provided by the EOS SDK.
+        /// </summary>
+        /// <returns>
+        /// True if the user is authenticated using the Auth Interface, false
+        /// otherwise.
+        /// </returns>
+        private static bool IsAuthenticatedByEpicAuthInterface()
```
paulhazen: Change this to `IsLoggedInWithEpicAuthInterface()`

Assets/Scripts/ServiceManagers/ServiceManager.cs
```diff
@@ -0,0 +1,215 @@
...
+
+        /// <summary>
+        /// Determines whether the user is currently authenticated using the
+        /// Connect Interface provided by the EOS SDK.
+        /// </summary>
+        /// <returns>
+        /// True if the user is authenticated using the Connect Interface, false
+        /// otherwise.
+        /// </returns>
+        private static bool IsAuthenticatedByConnectInterface()
```
paulhazen: Change this to `IsLoggedInWithConnectInterface()`

Assets/Scripts/ServiceManagers/ServiceManager.cs
```diff
@@ -0,0 +1,203 @@
...
+        /// Stores a list of functions to be called whenever data related to
+        /// the manager has been updated.
+        /// </summary>
+        protected IList<Action> _updateCallbacks = new List<Action>();
+
+        /// <summary>
+        /// Indicates whether the service needs to have a user authenticated in
+        /// order to function properly.
+        /// </summary>
+        private bool _requiresAuthentication;
```
andrew-hirata-playeveryware: I think this needs to be expanded into requiring ConnectLogin or AuthLogin (or potentially both).

Assets/Scripts/ServiceManagers/ServiceManager.cs
```diff
@@ -0,0 +1,203 @@
...
+
+        /// <summary>
+        /// Determines whether the user is currently authenticated using the
+        /// Auth Interface provided by the EOS SDK.
+        /// </summary>
+        /// <returns>
+        /// True if the user is authenticated using the Auth Interface, false
+        /// otherwise.
+        /// </returns>
+        private static bool IsLoggedInWithEpicAuthInterface()
```
paulhazen: Because neither `EOSAchievementManager` nor `StatsManager` utilize logging in using the Auth Interface provided by Epic, this particular function should be removed for the time being. It's possible that in future iterations of this class it can be re-introduced, but adding it now lends confusion.

@andrew-hirata-playeveryware for visibility - since this is per discussion with them.

paulhazen: Aforementioned changes have been made. Consequently, I am marking this conversation as 'resolved.'

andrew-hirata-playeveryware (approved):

Assets/Scripts/ServiceManagers/ServiceManager.cs
paulhazen: Donzo.

PR #733, opened 2024-06-19
Correct a misplaced `#endif`

paulhazen:
Corrects a bug in the placement of an `#endif` that cuts things off before it should have.

Thanks to @arthur740212 for discovering!


andrew-hirata-playeveryware (approved):
Looks code to me!

arthur740212 (approved):

PR #734, opened 2024-06-20
chore(native) : Restoring gitattributes with eol=lf instructions

WispyMouse:
Adds a .gitattributes file back to the NativeCode section. Only the file types that are actually in NativeCode are inside this. Added `eol=lf` to `*.cs`, which wasn't present before.

Having a more specific .gitattributes file will increase the likelihood that the Native library is following a consistent set of rules for all users of the plugin. The highest level .gitattributes file in the repository didn't include the C++ types, `.h` `.cpp` and `.c`, so this needs to have those rules. The `eol=lf` makes the line endings consistent regardless of client used to write/upload it.

### DIFF
--- a/lib/NativeCode/.gitattributes
+++ b/lib/NativeCode/.gitattributes
@@ -1,0 +1,4 @@
+*.cs diff=csharp text eol=lf
+*.h text eol=lf
+*.c text eol=lf
+*.cpp text eol=lf


arthur740212 (approved):

paulhazen (approved):
One nit I have regarding the description for this PR: it doesn't explain the "why" behind this change. Does this resolve some problem? If so - explaining the problem and how this resolves it would be good.

Also - add yourself as the "Assignee" for this PR. Helps provide a record of who is responsible for the change - regardless of who ends up actually performing the merge. Also helps expand the visibility of your impact on the project.

paulhazen (changes requested):
After some consideration - I think the comment I made on this PR probably merits a change - even though the changes I'm suggesting are not changes to the code changes, but changes I am requesting you make with respect to the format and administration of the PR as it exists within GitHub.

paulhazen (approved):

PR #736, opened 2024-06-21
Combine `IEOSAuthOnLogin` and `IEOSAuthOnLogout` into `IAuthInterfaceListener` interface.

paulhazen:
This PR clarifies the naming convention for `IEOSAuthOnLogin`, `IEOSAuthOnLogout`, and `IEOSConnectLogin` interfaces.

## Proper "is a" Design Concept

These changes ensure implementing classes properly reflect the "is a" relationship to their interfaces.

`EOSAchievementManager` is _not_ an `IEOSAuthOnLogin` or an `IEOSAuthOnLogout`; it listens for login and logout events passed to it by the Auth Interface via the `EOSManager`. Therefore, the two interface definitions have been combined into `IAuthInterfaceListener`, correcting the "is a" relationship such that `EOSAchievementManager` _is a_ Auth Interface event listener.

## Combines Common Functionality

Only the `EOSMetricsManager` implements logging out without needing to implement logging in. It now implements the login interface with an empty implementation, as loading metrics is an on-demand action.

> [!NOTE]
> Future improvement could involve loading new metrics when a user logs in, but this is beyond the scope of this PR.

## Naming Convention Disparity

This PR resolves inconsistencies in naming conventions for the Connect and Auth Interface behaviors. For example, the Connect Interface login was named `IEOSConnectLogin` instead of `IEOSConnectOnLogin`.


andrew-hirata-playeveryware (changes requested):
At the very least this needs a meeting if we're going to be breaking and changing our public API.

paulhazen:
Closing this PR until I'm able to strangler-fig the changes so that the change does not break current API.

PR #737, opened 2024-06-21
Combine `IEOSAuthOnLogin` and `IEOSAuthOnLogout` into `IAuthInterfaceEventListener` interface.

paulhazen:
This PR clarifies the naming convention for `IEOSAuthOnLogin`, `IEOSAuthOnLogout`, and `IEOSConnectLogin` interfaces.

## Proper "is a" Design Concept

These changes ensure implementing classes properly reflect the "is a" relationship to their interfaces.

`EOSAchievementManager` is _not_ an `IEOSAuthOnLogin` or an `IEOSAuthOnLogout`; it listens for login and logout events passed to it by the Auth Interface via the `EOSManager`. Therefore, the two interface definitions have been combined into `IAuthInterfaceEventListener`, correcting the "is a" relationship such that `EOSAchievementManager` _is a_ Auth Interface event listener.

## Combines Common Functionality

Only the `EOSMetricsManager` implements logging out without needing to implement logging in. It now implements the login interface with an empty implementation, as loading metrics is an on-demand action.

> [!NOTE]
> Future improvement could involve loading new metrics when a user logs in, but this is beyond the scope of this PR.

## Naming Convention Disparity

This PR resolves inconsistencies in naming conventions for the Connect and Auth Interface behaviors. For example, the Connect Interface login was named `IEOSConnectLogin` instead of `IEOSConnectOnLogin`.

> [!IMPORTANT]
> Unlike the previous PR that contained much of the same changes - this one introduces the combined interface definitions as interfaces that _inherit from_ the previously defined interfaces. All references internal to the project to any of the old interfaces have been replaced with references to the new combined interfaces.
>
> This means that there are no breaking changes to the API. Notes about how to reduce the redundancy in future iterations where breaking the API is deemed to be appropriate have been added in the appropriate places within the code.


Assets/Plugins/Source/Core/EOSManager.cs
```diff
@@ -323,64 +323,50 @@ static void print(string toPrint)
             }
 
             //-------------------------------------------------------------------------
-            public void AddConnectLoginListener(IEOSOnConnectLogin connectLogin)
+            public void AddConnectLoginListener(IConnectInterfaceEventListener connectLogin)
```
andrew-hirata-playeveryware: This is still changing the API from a `IEOSOnConnectLogin` type to a `IConnectInterfaceEventListner`, which changes the API signature.

Assets/Plugins/Source/Core/EOSManager.cs
```diff
@@ -323,64 +323,50 @@ static void print(string toPrint)
...
-            }
-
-            public void AddAuthLogoutListener(IEOSOnAuthLogout authLogout)
-            {
-                s_onAuthLogoutCallbacks.Add(authLogout.OnAuthLogout);
+                s_onConnectLoginCallbacks.Remove(connectLogin.OnConnectLogin);
             }
 
-            public void AddApplicationCloseListener(Action listener)
+            public void AddAuthInterfaceEventListener(IAuthInterfaceEventListener listener)
```
andrew-hirata-playeveryware: Likewise for this, it is also changing the API parameters when it shouldn't.

Assets/Plugins/Source/Core/EOSManager.cs
```diff
@@ -323,64 +323,50 @@ static void print(string toPrint)
...
-                T manager = default;
                 Type type = typeof(T);
+                T manager;
                 if (!s_subManagers.ContainsKey(type))
                 {
                     manager = new T();
                     s_subManagers.Add(type, manager);
 
-                    if (manager is IEOSOnConnectLogin)
+                    if (manager is IConnectInterfaceEventListener connectListener)
```
andrew-hirata-playeveryware: These also don't need to change.

andrew-hirata-playeveryware (changes requested):

Assets/Plugins/Source/Core/EOSManager.cs
paulhazen: Makes sense - I reverted to the original signatures. The file still shows some changes, but they are whitespace changes and I think a re-ordering of the function definitions.

Assets/Plugins/Source/Core/EOSManager.cs
paulhazen: They have been restored to the original implementation.

Assets/Plugins/Source/Core/EOSManager.cs
paulhazen: Fixed :)

Assets/Scripts/EOSStoreManager.cs
```diff
@@ -55,10 +55,15 @@ public void OnLoggedIn()
...
+        public void OnAuthLogin(LoginCallbackInfo loginCallbackInfo)
         {
             OnLoggedIn();
         }
+
+        public void OnAuthLogout(LogoutCallbackInfo logoutCallbackInfo)
+        {
+            OnLoggedOut();
+        }
 #endif
```
paulhazen, line 67: Before merging, make certain that if you define `EOS_DISABLE` in the scripting defines section of the player build settings, this does not break (I'm pretty sure that it will, and some additional `#if` statements will need to be added).

WispyMouse (approved):

Assets/Scripts/EOSStoreManager.cs
paulhazen: I have tested this and determined that it is not an issue, since when `EOS_DISABLE` is defined, everything within `Assets/Scripts/` is excluded from the build (including `EOSStoreManager.cs`).

andrew-hirata-playeveryware (approved):

PR #738, opened 2024-06-21
Upgrade android api level to 33

arthur740212:
Minimum API Level that EOS SDK 1.16.2 supports is 33


paulhazen (approved):

andrew-hirata-playeveryware (approved):

TrentSterling:
Is this change actually required? I see EOS SDK 1.16.4 supports api level 23 as mentioned here: https://dev.epicgames.com/docs/epic-online-services/platforms/android

My projects have been using a fairly old build of the eos_plugin_for_unity, but when I moved to Unity 6, I found that this higher android API requirement is causing issues with submission to the Meta store. I tried to simply change the target back to 32 (this is the recommended version on meta store), but one of the dependencies started complaining during the gradle build process.

What am I missing here?

PR #742, opened 2024-06-24
feature (Steamworks) : SteamApiInterfaceVersionsArray

WispyMouse:
Starting in Steamworks v1.58a onwards, if `LibraryManagedBySDK` is defined as a Steam platform flag, then it is required to pass the version of all interface APIs in Steamworks.

This set of changes allows for definition of the Steamworks API versions inside the Steam window for EOS Configuration. You can manually insert each record, or if you're using Steamworks.NET you can press the "Update from Steamworks.NET" button to populate the list from its list of constants.

image

Facepunch users will have to manually insert these values, because those values are not exposed in a way that the plugin can programmatically pull from. It will be required for anyone who is using `LibraryManagedBySDK` moving forward to go to the EOS Configuration window once and update and save their configuration.


Assets/Plugins/Source/Editor/Utility/Steam/Steamworks_Utility.cs
```diff
@@ -28,6 +28,7 @@
 
 #if !DISABLESTEAMWORKS
 using Steamworks;
+using System.Collections.Generic;
```
arthur740212: Because `GetSteamInterfacesVersions()` returns a `List`, this namespace needs to be outside of `!DISABLESTEAMWORKS` scope

Assets/Plugins/Source/Editor/Utility/Steam/Steamworks_Utility.cs
```diff
@@ -40,6 +41,48 @@ public static string GetSteamworksVersion()
...
+        /// <summary>
+        /// This is to populate the SteamConfig's steamApiInterfaceVersionsArray. It needs to be configured exactly like this for Steamworks v1.58a onwards.
+        /// This value is identical to steam_api.h's pszInternalCheckInterfaceVersions value.
+        /// https://dev.epicgames.com/docs/en-US/api-ref/structs/eos-integrated-platform-steam-options
+        /// </summary>
+        /// <returns></returns>
+        public static List<string> GetSteamInterfaceVersions()
+        {
+#if DISABLESTEAMWORKS
+            return string.Empty;
```
arthur740212: Return type mismatch if `DISABLESTEAMWORKS`
null might work here, just need to check if it affects the steam config window.

arthur740212 (changes requested):
Great job on a new feature.
Only small picks for when Steamworks was disabled.

Assets/Plugins/Source/Editor/Utility/Steam/Steamworks_Utility.cs
WispyMouse: Good catch! Changed to be outside this directive.

Assets/Plugins/Source/Editor/Utility/Steam/Steamworks_Utility.cs
WispyMouse: Oh, yup. This used to return a string and I didn't come back to change this after it became a list. Good catch!

paulhazen:
> Starting in Steamworks v1.58a onwards, if `LibraryManagedBySDK` is defined as a Steam platform flag, then it is required to pass the version of all interface APIs in Steamworks.
> 
> This set of changes allows for definition of the Steamworks API versions inside the Steam window for EOS Configuration. You can manually insert each record, or if you're using Steamworks.NET you can press the "Update from Steamworks.NET" button to populate the list from its list of constants.
> 
> image
> 
> Facepunch users will have to manually insert these values, because those values are not exposed in a way that the plugin can programmatically pull from. It will be required for anyone who is using `LibraryManagedBySDK` moving forward to go to the EOS Configuration window once and update and save their configuration.

Just to confirm - does a user _need_ to press the "Update values" button for this to work? If so, is there a way that we can make this happen automatically instead?

WispyMouse:
> Just to confirm - does a user _need_ to press the "Update values" button for this to work? If so, is there a way that we can make this happen automatically instead?

I hooked up something to BuildRunner to attempt to populate the list automatically. Hopefully that'll be more convenient! It only does this population if the array is not present, or empty, in the SteamConfig file.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/dllmain.cpp
```diff
@@ -1244,6 +1257,37 @@ void eos_create(EOSConfig& eosConfig)
...
+            {
+                char currentCharacter = currentFullValue[characterIndex];
+                steamApiInterfaceVersionsAsCharArray.push_back(currentCharacter);
+                totalBytes++;
+            }
+
+            steamApiInterfaceVersionsAsCharArray.push_back('\0');
+            totalBytes++;
+        }
+        steamApiInterfaceVersionsAsCharArray.push_back('\0');
```
andrew-hirata-playeveryware, line 1276: Isn't `totalBytes` always equal to `steamApiInterfaceVersionsAsCharArray.size()`?

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/dllmain.cpp
```diff
@@ -1244,6 +1257,37 @@ void eos_create(EOSConfig& eosConfig)
         steam_platform.SteamMajorVersion = eos_steam_config.steamSDKMajorVersion;
         steam_platform.SteamMinorVersion = eos_steam_config.steamSDKMinorVersion;
 
+        // For each element in the array (each of which is a string of an api version information)
+        // iterate across each character, and at the end of a string add a null terminator \0
+        // then add one more null terminator at the end of the array
+        std::vector<char> steamApiInterfaceVersionsAsCharArray = std::vector<char>();
```
andrew-hirata-playeveryware: The assignment here is not needed.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/dllmain.cpp
```diff
@@ -1244,6 +1257,37 @@ void eos_create(EOSConfig& eosConfig)
...
+
+        // steam_platform needs to have a count of how many bytes the "array" is, stored in SteamApiInterfaceVersionsArrayBytes
+        // This has some fuzzy behavior; if you set it to 0 or count it up properly, there won't be a logged problem
+        // if you put a non-zero amount that is insufficient, there will be an unclear logged error message
+        // HACK: For now the right method seems to be counting each char added to the array, including the null terminator
+        uint32_t totalBytes = 0;
+
+        for (int apiInterfaceVersionIndex = 0; apiInterfaceVersionIndex < eos_steam_config.steamApiInterfaceVersionsArray.size(); apiInterfaceVersionIndex++)
+        {
+            std::string currentFullValue = eos_steam_config.steamApiInterfaceVersionsArray[apiInterfaceVersionIndex];
```
andrew-hirata-playeveryware: I believe you can get away with using a `const std::string&` here instead of a `std::string` so that the string doesn't get copied.

andrew-hirata-playeveryware (changes requested):

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/dllmain.cpp
WispyMouse: I made this go through the motions and then use `.size()`, and that did indeed get the correct size! I had been using `sizeof`, and that was only counting up until the first `\0`. `size()` is functioning correctly, so we'll use that instead of counting up in the loop like this.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/dllmain.cpp
WispyMouse: Ah, I see! This makes it point at the value in the array instead of copying it in to memory. That makes sense, I'll utilize that.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/dllmain.cpp
WispyMouse: You're right. I had assumed that I needed to set it to something before I can start managing it, but it appears as though `std::vector<>` lets you use it right away without the assignment. Learned something new about C++!

arthur740212 (approved):

paulhazen (approved):

andrew-hirata-playeveryware (approved):

PR #746, opened 2024-06-25
Modest improvements to `UnityProjectUtility`

paulhazen:
This PR provides modest improvements to `UnityProjectUtility`, which aids in the discovery of missing script references, missing prefab references, and the identification of unreferenced prefabs.

### DIFF
--- a/Assets/Plugins/Source/Editor/Utility/UnityProjectUtility.cs
+++ b/Assets/Plugins/Source/Editor/Utility/UnityProjectUtility.cs
@@ -92,8 +92,8 @@ namespace PlayEveryWare.EpicOnlineServices.Editor.Utility
                 // Open the scene
                 Scene currentScene = EditorSceneManager.OpenScene(scenePath, OpenSceneMode.Single);
 
-                // Get all the game objects
-                GameObject[] gameObjects = Object.FindObjectsOfType<GameObject>();
+                // Get all the game objects (including inactive objects)
+                GameObject[] gameObjects = Object.FindObjectsOfType<GameObject>(true);
 
                 // Iterate over all the game objects
                 foreach (GameObject go in gameObjects)
@@ -116,6 +116,7 @@ namespace PlayEveryWare.EpicOnlineServices.Editor.Utility
             }
             else
             {
+                Debug.LogWarning($"There were {missingScriptMessages.Count} scripts found to be missing in scenes.");
                 foreach ((string message, Object context) in missingScriptMessages)
                 {
                     Debug.LogWarning(message, context);
--- a/Assets/Plugins/Source/Editor/Utility/UnityProjectUtility.cs
+++ b/Assets/Plugins/Source/Editor/Utility/UnityProjectUtility.cs
@@ -21,7 +21,7 @@
  */
 
 // Uncomment the following define to enable menu items detailed below
-// #define PROJECT_UTILITY_ENABLED
+//#define PROJECT_UTILITY_ENABLED
 
 #if PROJECT_UTILITY_ENABLED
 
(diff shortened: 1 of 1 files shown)


arthur740212 (approved):

Assets/Plugins/Source/Editor/Utility/UnityProjectUtility.cs
```diff
@@ -77,16 +79,18 @@ public static bool CanFindMissingScriptsInPrefabs()
...
+
             // Iterate over all the scenes
             foreach (string scenePath in EditorBuildSettingsScene.GetActiveSceneList(EditorBuildSettings.scenes))
             {
                 // Open the scene
-                EditorSceneManager.OpenScene(scenePath);
+                Scene currentScene = EditorSceneManager.OpenScene(scenePath, OpenSceneMode.Single);
 
                 // Get all the game objects
                 GameObject[] gameObjects = Object.FindObjectsOfType<GameObject>();
```
WispyMouse: If an object is inactive in the scene, will this find it? Should we pass in "includeInactive: true" to this function?

Assets/Plugins/Source/Editor/Utility/UnityProjectUtility.cs
```diff
@@ -101,15 +105,25 @@ private static void FindMissingScripts()
...
-                            $"Script \"{FullGameObjectPath(go)}\" in scene \"{Path.GetFileNameWithoutExtension(scenePath)}\" is missing.",
-                            go);
+                        missingScriptMessages.Add(($"Script \"{FullGameObjectPath(go)}\" in scene \"{Path.GetFileNameWithoutExtension(scenePath)}\" is missing.", go));
                     }
                 }
             }
 
+            if (0 == missingScriptMessages.Count)
+            {
+                Debug.Log("There were no scripts found to be missing in any scene.");
```
WispyMouse, line 115: I like this log; can we add a similar one for if errors are found? Just something that spits out the message "There were {missingScriptMessages.Count} missing scripts found in scenes".

Assets/Plugins/Source/Editor/Utility/UnityProjectUtility.cs
```diff
@@ -122,51 +136,56 @@ private static bool CanFindMissingScriptsInScenes()
...
-                {
-                    string path = AssetDatabase.GUIDToAssetPath(prefabGuid);
-                    Debug.Log("Unreferenced Prefab: " + path);
-                }
+                // Skip if the prefab is referenced
+                if (referencedPrefabGuids.Contains(prefabGuid))
+                    continue;
+
+                string path = AssetDatabase.GUIDToAssetPath(prefabGuid);
+                Debug.LogWarning($"Unreferenced Prefab: {path}");
```
WispyMouse, line 177: thinking out loud, not essentially a thing that needs to be changed: This does exactly as it is supposed to, find all prefabs without references. But what about things that get loaded from Streaming Assets by their file path? Is there a world where this search is constrained to not search inside of Resources and Streaming Assets folders?

And what if the only reference to it is in Addressable bundles? E.x. I have an addressables bundle that contains GameObjects with cosmetic information in it. Those prefabs aren't referenced elsewhere in the game, just in being downloaded by their bundle...

Anyways that all said, this tool seems useful, and we already had it. Perhaps people who will utilize it will already be aware of the limitations of actually trying to answer "is this resource used by anything".

WispyMouse (approved):
Looks good. While not something directly changed by this changeset, please consider making the FindMissingScriptsInScenes function to FindObjectsOfType to include inactive objects.

Assets/Plugins/Source/Editor/Utility/UnityProjectUtility.cs
paulhazen: This is an interesting point, I will keep it in mind.

Assets/Plugins/Source/Editor/Utility/UnityProjectUtility.cs
paulhazen: Good catch - I didn't realize that it defaulted to exclude inactive objects. I'll change that.

Assets/Plugins/Source/Editor/Utility/UnityProjectUtility.cs
paulhazen: Not certain if you are looking for something other than what is done on lines 119-122. If there are missing scenes, a warning is logged for each one that is missing.

Assets/Plugins/Source/Editor/Utility/UnityProjectUtility.cs
paulhazen: Fixed.

Assets/Plugins/Source/Editor/Utility/UnityProjectUtility.cs
paulhazen: I think I know what you mean - like before listing each script that's missing, first just print out how many there are. So I changed the implementation to match that.

PR #751, opened 2024-06-26
Log Filter

paulhazen:
This PR introduces a log filter, which can stop certain log messages from being logged, depending on a set of terms from which to exclude messages based off of.

This feature is disabled to "off", but for circumstances where the log window is overwhelmed with `<Redacted>` log messages from the EOS SDK, enabling this can help clarify what is going on.


WispyMouse (approved):

arthur740212 (approved):

andrew-hirata-playeveryware (changes requested):
I believe this needs either a feature meeting or a task where we define what this is for.

andrew-hirata-playeveryware (approved):

PR #754, opened 2024-06-28
Make use of `async` keyword conditional on whether SteamWorks is enabled or not.

paulhazen:
In a recent PR changes were introduced to the `SteamManager` that make one of the function calls `async`. However, if SteamWorks is disabled, no function is called with the `await` keyword, and a compiler warning is generated that suggest removing the `async` keyword. 

This resolves that issue by making the use of the `async` keyword conditional on whether or not SteamWorks is enabled.

### DIFF
--- a/Assets/Scripts/Steam/SteamManager.cs
+++ b/Assets/Scripts/Steam/SteamManager.cs
@@ -414,21 +414,15 @@ namespace PlayEveryWare.EpicOnlineServices.Samples.Steam
         /// NOTE: This conditional is here because if EOS_DISABLE is enabled, the members referenced
         ///       in this code block will not exist on EOSManager.
 #if !EOS_DISABLE
-        public
-            // NOTE: This conditional is here because the function should only
-            //       be marked as 'async' if it calls an awaitable function 
-            //       using the 'await' keyword. When SteamWorks is disabled, 
-            //       there is no such call, which results in a compiler warning.
-#if !DISABLESTEAMWORKS
-            async 
-#endif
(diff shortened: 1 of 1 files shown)


WispyMouse (approved):
I'm not thrilled with this solution, but I don't have a better one. We could for example tell the function to await something that finishes immediately. But since this is an `async void` rather than `async Task`, I suppose there's no downside to removing the async conditionally.

Assets/Scripts/Steam/SteamManager.cs
```diff
@@ -414,8 +414,16 @@ public void RequestAppTicket(Action<string> callback)
...
-        {
+        public
+            // NOTE: This conditional is here because the function should only
+            //       be marked as 'async' if it calls an awaitable function 
+            //       using the 'await' keyword. When SteamWorks is disabled, 
+            //       there is no such call, which results in a compiler warning.
+#if !DISABLESTEAMWORKS
+            async 
+#endif
+            void StartLoginWithSteam(EOSManager.OnAuthLoginCallback onLoginCallback)
```
paulhazen: It might be better to have an empty `await` call within this - so that the `async` keyword is not conditionally applied - which _might_ cause some weird behavior - but more to the point - it will be clearer in the code what is happening if we await an empty thing.

paulhazen: Changes have been made as suggested above, marking conversation as resolved.

arthur740212 (approved):

PR #762, opened 2024-07-02
Release `v3.3.0`

paulhazen:
In rebuilding the native code libraries for the upcoming release - I discovered that changes made to the native code files introduced some code that generated compiler warnings, particularly regarding the comparison of signed and unsigned integers. 

This PR does the following:

- Changes `for` loop variables to be `unsigned` instead of `int`, so as to avoid the compiler warning.
- Remove a duplicate macro definition that was also generating a warning.
- Adds the PEW copyright comment header block to all relevant code files.
- Accordingly updates the binaries that already exist within the repository with ones that result from the compilation of the aforementioned changes.
- Updates the git build sha within the `package.json` file

This _should_ be the final pull request for the latest release branch. Upon merging it - I will subsequently open a PR into `stable` from `release-3.3.0`, and then tag / announce / make the release (updating the corresponding upm repository as well).


lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/dllmain.cpp
```diff
@@ -413,7 +436,7 @@ void eos_set_loglevel(const LogLevelConfig& log_config)
 {
     if (EOS_Logging_SetLogLevel_ptr != nullptr)
     {
-        for (int i = 0; i < log_config.category.size() - 1; i++)
+        for (unsigned i = 0; i < log_config.category.size() - 1; i++)
```
andrew-hirata-playeveryware: From a mildly pedantic perspective, I would prefer this to be `size_t` instead of unsigned.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/dllmain.cpp
```diff
@@ -1353,11 +1376,18 @@ void eos_create(EOSConfig& eosConfig)
         steamApiInterfaceVersionsAsCharArray.push_back('\0');
 
         steam_platform.SteamApiInterfaceVersionsArray = reinterpret_cast<char*>(steamApiInterfaceVersionsAsCharArray.data());
+                
+        auto size = steamApiInterfaceVersionsAsCharArray.size();
+
+        if (size > (std::numeric_limits<uint32_t>::max)()) {
+            log_error("Size given for SteamApiInterfaceVersionsAsCharArray exceeds the maximum value for uint32_t.");
+            throw std::overflow_error("Size exceeds the maximum value for uint32_t");
```
andrew-hirata-playeveryware: I'm intrigued this worked because I thought we had exceptions disabled.

andrew-hirata-playeveryware (dismissed):

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/dllmain.cpp
paulhazen: I wrote this without knowing that they were disabled, just seemed like the appropriate implementation just looking at the narrow bit of code here.

Is it possible that exceptions are disabled in the sense that throwing them does not actually _do_ anything?

Either way, should the implementation be changed so that a default value is used instead of an exception being thrown like it currently is?

What _would have_ happened previously if there was an overflow error?

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/dllmain.cpp
paulhazen: Yeah, that's a good point.

paulhazen: Changed both instances to actually use the for each loop syntax instead.

paulhazen: Closing this issue out as the problem addressed no longer exists.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/dllmain.cpp
```diff
@@ -1353,11 +1373,18 @@ void eos_create(EOSConfig& eosConfig)
         steamApiInterfaceVersionsAsCharArray.push_back('\0');
 
         steam_platform.SteamApiInterfaceVersionsArray = reinterpret_cast<char*>(steamApiInterfaceVersionsAsCharArray.data());
+                
+        auto size = steamApiInterfaceVersionsAsCharArray.size();
+
+        if (size > (std::numeric_limits<uint32_t>::max)()) {
```
WispyMouse: I should have done this when setting this up, but if we're doing this comparison; there's also a maximum size this message can be, maybe we should also check against that.

WispyMouse: https://dev.epicgames.com/docs/en-US/api-ref/structs/eos-integrated-platform-steam-options

> Size of the SteamApiInterfaceVersionsArray in bytes. Cannot exceed EOS_INTEGRATEDPLATFORM_STEAM_MAX_STEAMAPIINTERFACEVERSIONSARRAY_SIZE.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/dllmain.cpp
```diff
@@ -1353,11 +1373,18 @@ void eos_create(EOSConfig& eosConfig)
...
+            log_error("Size given for SteamApiInterfaceVersionsAsCharArray exceeds the maximum value for uint32_t.");
+            throw std::overflow_error("Size exceeds the maximum value for uint32_t");
```
paulhazen: A change should be made to avoid throwing an exception (or entering a situation where one is going to be thrown).

lib/NativeCode/include/windows/framework.h
```diff
@@ -11,5 +11,3 @@
 #define STATIC_EXPORT(return_type) extern "C" return_type
 
 #define DLL_EXPORT(return_value) extern "C" __declspec(dllexport) return_value  __stdcall
-
-#define FUN_EXPORT(return_value) DLL_EXPORT(return_value)
```
WispyMouse, line 15: Intentional exclusion? Is this the,

> Remove a duplicate macro definition that was also generating a warning.

referenced in the description?

WispyMouse: Per your description in the pull request meeting, this is an intentional clean up.

WispyMouse (approved):

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/dllmain.cpp
paulhazen: Also place the open curly on a new line.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/dllmain.cpp
paulhazen: Steps to resolve the concerns described in this conversation have been articulated in a separate conversation, so I am marking this one as resolved.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/dllmain.cpp
paulhazen: Exception throw statement has been removed.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/dllmain.cpp
paulhazen: The if statement now just compares against the aforementioned constant.

arthur740212 (approved):

PR #763, opened 2024-07-03
feature (sessions): Reworking Sample, P2P Communication Part

WispyMouse:
This commit includes:
- P2P messaging in the Sessions Sample
- Some scenarios (join, leave, destroy, modify, start, end) have P2P messages sent to members of session, registering and unregistering users
- GetSessionById for getting a local session by id

This commit lacks:
- Taking the Refresh actions implied by received messages
- Support for server-owned sessions
- Several common situations where a user would trigger a refresh to the owner

I would like feedback on:
- Are all appropriate protocols for P2P messaging and connection management being followed?
- Is the managing of common error states appropriately done?
- Does anyone have concerns I hadn't considered about this process?
- Does my use of #region look appropriate?
- Is the position of event subscription appropriate?

### DIFF
--- a/Assets/Scripts/EOSSessionsManager.cs
+++ b/Assets/Scripts/EOSSessionsManager.cs
@@ -20,18 +20,24 @@
 * SOFTWARE.
 */
 
-using System;
-using System.Collections.Generic;
-
-using UnityEngine;
-
-using Epic.OnlineServices;
-using Epic.OnlineServices.Sessions;
-using Epic.OnlineServices.Presence;
-using Epic.OnlineServices.UI;
+// Uncomment this define, or define it in build symbols, to get non-error debugging messages from this Manager
+// #define ENABLE_DEBUG_SESSIONS_MANAGER
 
 namespace PlayEveryWare.EpicOnlineServices.Samples
 {
+    using System;
+    using System.Collections.Generic;
+
+    using UnityEngine;
+
+    using Epic.OnlineServices;
+    using Epic.OnlineServices.Sessions;
+    using Epic.OnlineServices.Presence;
+    using Epic.OnlineServices.UI;
+    using Epic.OnlineServices.P2P;
+    using System.Text.RegularExpressions;
+    using System.Text;
+
     /// <summary>
     /// Class represents a session search and search results
     /// </summary>
@@ -359,6 +365,70 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
 
         private bool subscribtedToGameInvites = false;
         private bool userLoggedIn = false;
+
+        #region Peer2Peer Messaging Variables
+        /// <summary>
+        /// When sending or receiving Peer2Peer messages relating to Session Status upkeep, always use this same socket name.
+        /// This should be a distinct and unique socket name, different than any other socket name utilized in the program.
+        /// </summary>
+        private const string P2PSessionStatusSocketName = "SESSIONSTATUS";
+
+        /// <summary>
+        /// When sending or receiving Peer2Peer messages relating to Session status upkeep, always use this same channel.
+        /// This should be a distinct and unique channel, different than any other socket name utilized in the program, if possible.
+        /// </summary>
+        private const byte P2PSessionStatusUpdateChannel = 0xF;
+
+        /// <summary>
+        /// Messages about sessions should all start with this message, so that parsing the message for information is easy.
+        /// This is formatted in <see cref="P2PInformSessionMessageFormat"/> while sending messages.
+        /// </summary>
+        private const string P2PInformSessionMessageBase = "SESSIONINFORMATION";
+
+        /// <summary>
+        /// Messages about sessions follow this format, starting with <see cref="P2PInformSessionMessageBase"/>.
+        /// {0} - The back-end <see cref="Session.Id"/> of the session to message about.
+        /// {1} - The <see cref="ProductUserId"/> of the user sending the message.
+        /// {2} - Additional information about the message.
+        /// <see cref="P2PJoiningSessionMessageElement"/>
+        /// <see cref="P2PLeavingSessionMessageElement"/>
+        /// <see cref="P2PRefreshSessionMessageElement"/>
+        /// <see cref="P2PSessionOwnerDestroyedSessionMessageElement"/>
+        /// </summary>
+        private const string P2PInformSessionMessageFormat = P2PInformSessionMessageBase + " ({0}) ({1}) ({2})";
+
+        /// <summary>
+        /// Messages with this as the {2} parameter of <see cref="InformSessionOwnerMessageFormat"/> indicate a user is joining the session.
+        /// </summary>
+        private const string P2PJoiningSessionMessageElement = "JOIN";
+
+        /// <summary>
+        /// Messages with this as the {2} parameter of <see cref="P2PInformSessionMessageFormat"/> indicate a user is leaving the session.
+        /// </summary>
+        private const string P2PLeavingSessionMessageElement = "LEAVE";
+
+        /// <summary>
+        /// Messages with this as the {2} parameter of <see cref="P2PInformSessionMessageFormat"/> indicate that a user should re-acquire and refresh session information.
+        /// </summary>
+        private const string P2PRefreshSessionMessageElement = "REFRESH";
+
+        /// <summary>
+        /// Messages with this as the {2} parameter of <see cref="P2PInformSessionMessageFormat"/> indicate that the owner of a session has destroyed the session,
+        /// so members of the session should also remove themselves from the session.
+        /// </summary>
+        private const string P2PSessionOwnerDestroyedSessionMessageElement = "DESTROY";
+
+        /// <summary>
+        /// When subscribing to peer request connection messages, this Id is held as a way to later remove the subscription.
+        /// </summary>
+        private ulong P2PSessionPeerRequestConnectionNotificationId { get; set; }
+
+        /// <summary>
+        /// When subscribing to peer disconnection messages, this Id is held as a way to later remove the subscription.
+        /// </summary>
+        private ulong P2PSessionPeerDisconnectConnectionNotificationId { get; set; }
+        #endregion
+
         public bool IsUserLoggedIn
         {
             get { return userLoggedIn; }
@@ -376,6 +446,18 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
             CurrentSearch = new SessionSearch();
             Invites = new Dictionary<Session, SessionDetails>();
             CurrentInvite = null;
+        }
+
+        /// <summary>
+        /// Conditionally executed proxy function for Unity's log function.
+        /// Errors are logged using <see cref="UnityEngine.Debug.LogError(object)"/>.
+        /// If this is included, it is used to log out non-error debug messages.
+        /// </summary>
+        /// <param name="toPrint">The message to log.</param>
+        [System.Diagnostics.Conditional("ENABLE_DEBUG_SESSIONS_MANAGER")]
+        private static void Log(string toPrint)
+        {
+            UnityEngine.Debug.Log(toPrint);
         }
 
         public Dictionary<Session, SessionDetails> GetInvites()
@@ -465,12 +547,14 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
         {
             DestroyAllSessions();
             UnsubscribeFromGameInvites();
+            UnsubscribeToSessionMessageConnectionRequests();
         }
 
         //-------------------------------------------------------------------------
         public bool Update()
         {
             bool stateUpdates = false;
+            HandleReceivedP2PMessages();
 
             //Update active session from time to time
             foreach (KeyValuePair<string, Session> kvp in CurrentSessions)
@@ -517,7 +601,10 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
                 return;
             }
 
+            EOSManager.Instance.SetLogLevel(Epic.OnlineServices.Logging.LogCategory.AllCategories, Epic.OnlineServices.Logging.LogLevel.Warning);
+
             SubscribteToGameInvites();
+            SubscribeToSessionMessageConnectionRequests();
 
             CurrentInvite = null;
 
@@ -533,6 +620,7 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
             }
 
             UnsubscribeFromGameInvites();
+            UnsubscribeToSessionMessageConnectionRequests();
 
             LeaveAllSessions();
 
@@ -793,6 +881,19 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
             return null;
         }
 
+        public Session GetSessionById(string id)
+        {
+            foreach (Session curSession in CurrentSessions.Values)
+            {
+                if (curSession.Id.Equals(id))
+                {
+                    return curSession;
+                }
+            }
+
+            return null;
+        }
+
         public void StartSession(string name)
         {
             if (CurrentSessions.TryGetValue(name, out Session session))
@@ -833,7 +934,7 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
             registerOptions.PlayersToRegister = new ProductUserId[] { friendId };
 
             SessionsInterface sessionInterface = EOSManager.Instance.GetEOSPlatformInterface().GetSessionsInterface();
-            sessionInterface.RegisterPlayers(ref registerOptions, null, OnRegisterCompleteCallback);
+            sessionInterface.RegisterPlayers(ref registerOptions, sessionName, OnRegisterCompleteCallback);
         }
 
         public void UnRegister(string sessionName, ProductUserId friendId)
@@ -843,7 +944,7 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
             unregisterOptions.PlayersToUnregister = new ProductUserId[] { friendId };
 
             SessionsInterface sessionInterface = EOSManager.Instance.GetEOSPlatformInterface().GetSessionsInterface();
-            sessionInterface.UnregisterPlayers(ref unregisterOptions, null, OnUnregisterCompleteCallback);
+            sessionInterface.UnregisterPlayers(ref unregisterOptions, sessionName, OnUnregisterCompleteCallback);
         }
 
         public void InviteToSession(string sessionName, ProductUserId friendId)
@@ -1273,6 +1374,8 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
                         CurrentSessions.Remove(sessionName);
                     }
                 }
+
+                InformSessionMembers(sessionName, P2PRefreshSessionMessageElement);
             }
         }
 
@@ -1379,6 +1482,8 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
                     {
                         CurrentSessions[session.Name] = session;
                     }
+
+                    InformSessionOwnerWithMessage(session.Name, P2PJoiningSessionMessageElement);
                 }
                 callback?.Invoke(result);
             }
@@ -1599,6 +1704,12 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
             Debug.LogFormat("Session Matchmaking(OnStartSessionCompleteCallback): Started session: {0}", sessionName);
 
             //OnSessionStarted(sessionName); // Needed for C# wrapper?
+
+            // ClientData should contain the local sessionName
+            if (data.ClientData is string localSessionName)
+            {
+                InformSessionMembers(localSessionName, P2PRefreshSessionMessageElement);
+            }
         }
 
         private void OnEndSessionCompleteCallback(ref EndSessionCallbackInfo data)
@@ -1620,6 +1731,12 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
             Debug.LogFormat("Session Matchmaking(OnEndSessionCompleteCallback): Ended session: {0}", sessionName);
 
             //OnSessionEnded(sessionName); // Not used in C# wrapper
+
+            // ClientData should contain the local sessionName
+            if (data.ClientData is string localSessionName)
+            {
+                InformSessionMembers(localSessionName, P2PRefreshSessionMessageElement);
+            }
         }
 
         private void OnDestroySessionCompleteCallback(ref DestroySessionCallbackInfo data)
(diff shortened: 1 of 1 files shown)


Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -359,6 +365,70 @@ public class EOSSessionsManager : IEOSSubManager
 
         private bool subscribtedToGameInvites = false;
         private bool userLoggedIn = false;
+
+        #region Peer2Peer Messaging Variables
```
paulhazen, line 369: nit: Add an empty line after the `#region` statement.

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -359,6 +365,70 @@ public class EOSSessionsManager : IEOSSubManager
...
+        /// <summary>
+        /// When sending or receiving Peer2Peer messages relating to Session Status upkeep, always use this same socket name.
+        /// This should be a distinct and unique socket name, different than any other socket name utilized in the program.
+        /// </summary>
+        private const string P2PSessionStatusSocketName = "SESSIONSTATUS";
+
+        /// <summary>
+        /// When sending or receiving Peer2Peer messages relating to Session status upkeep, always use this same channel.
+        /// This should be a distinct and unique channel, different than any other socket name utilized in the program, if possible.
```
paulhazen: nit: Discrepancy between your qualifications of uniqueness between `P2PSessionStatusSocketName` and `P2PSessionStatusUpdateChannel`. Is this intentional, or is it okay to use a non-unique channel?

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -359,6 +365,70 @@ public class EOSSessionsManager : IEOSSubManager
...
+        /// </summary>
+        private const byte P2PSessionStatusUpdateChannel = 0xF;
+
+        /// <summary>
+        /// Messages about sessions should all start with this message, so that parsing the message for information is easy.
+        /// This is formatted in <see cref="P2PInformSessionMessageFormat"/> while sending messages.
```
paulhazen: nit: This is not formatted - it is used as a _component_ of formatting.

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -359,6 +365,70 @@ public class EOSSessionsManager : IEOSSubManager
...
+        /// <summary>
+        /// Messages with this as the {2} parameter of <see cref="P2PInformSessionMessageFormat"/> indicate that the owner of a session has destroyed the session,
+        /// so members of the session should also remove themselves from the session.
+        /// </summary>
+        private const string P2PSessionOwnerDestroyedSessionMessageElement = "DESTROY";
+
+        /// <summary>
+        /// When subscribing to peer request connection messages, this Id is held as a way to later remove the subscription.
+        /// </summary>
+        private ulong P2PSessionPeerRequestConnectionNotificationId { get; set; }
```
paulhazen, line 431: What happens if there is more than one peer request connection message?

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -359,6 +365,70 @@ public class EOSSessionsManager : IEOSSubManager
...
+
+        /// <summary>
+        /// When subscribing to peer disconnection messages, this Id is held as a way to later remove the subscription.
+        /// </summary>
+        private ulong P2PSessionPeerDisconnectConnectionNotificationId { get; set; }
+        #endregion
```
paulhazen, line 440: nit: Similar to the opening of the region, I believe it's appropriate to always have an empty line precede the closing of a region.

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -517,7 +601,10 @@ public void OnLoggedIn()
                 return;
             }
 
+            EOSManager.Instance.SetLogLevel(Epic.OnlineServices.Logging.LogCategory.AllCategories, Epic.OnlineServices.Logging.LogLevel.Warning);
```
paulhazen: I believe that this should be configured via our new log level setting window. I suspect that you added it here for diagnostic purposes, but it should not be explicitly set here. Particularly because this sets the log level of _all categories_ to warning, which would override the values set in the log level config system we have now.

@arthur740212 can you confirm?

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -517,7 +601,10 @@ public void OnLoggedIn()
...
+
             SubscribteToGameInvites();
```
paulhazen, line 614: Not a component of your PR - but I think it's erroneous enough that you should correct this function signature while you are making changes to this file.

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -793,6 +881,19 @@ public Session GetSession(string name)
...
         }
 
+        public Session GetSessionById(string id)
+        {
+            foreach (Session curSession in CurrentSessions.Values)
+            {
+                if (curSession.Id.Equals(id))
+                {
+                    return curSession;
+                }
```
paulhazen: nit: Invert the if statement to reduce nesting. For `continue` statements, it's acceptable to exclude the curly brackets. The result would be:

```csharp
if (!curSession.Id.Equals(id))
  continue;

return curSession;
```

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -1599,6 +1704,12 @@ private void OnStartSessionCompleteCallBack(ref StartSessionCallbackInfo data)
             Debug.LogFormat("Session Matchmaking(OnStartSessionCompleteCallback): Started session: {0}", sessionName);
 
             //OnSessionStarted(sessionName); // Needed for C# wrapper?
+
+            // ClientData should contain the local sessionName
+            if (data.ClientData is string localSessionName)
```
paulhazen, line 1719: +1 for using inline declaration 😎

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -1862,5 +2021,419 @@ public void DeclineLobbyInvite()
...
+                P2PSessionPeerRequestConnectionNotificationId = 0;
+            }
+
+            if (P2PSessionPeerDisconnectConnectionNotificationId != 0)
+            {
+                EOSManager.Instance.GetEOSPlatformInterface().GetP2PInterface().RemoveNotifyPeerConnectionClosed(P2PSessionPeerDisconnectConnectionNotificationId);
+                P2PSessionPeerDisconnectConnectionNotificationId = 0;
+            }
+
+            CloseConnectionsOptions closeOptions = new CloseConnectionsOptions() { LocalUserId = EOSManager.Instance.GetProductUserId(), SocketId = new SocketId() { SocketName = P2PSessionStatusSocketName } };
```
paulhazen: For the sake of greater clarity, I would suggest putting each initialized value on a new line.

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -1644,6 +1761,36 @@ private void OnDestroySessionCompleteCallback(ref DestroySessionCallbackInfo dat
...
+                }
+            }
+            else
+            {
+                Debug.LogError($"Session Matchmaking (OnDestroySessionCompleteCallback): Could not find local Session and associated ActiveSession, so could not inform owner/members of destruction.");
+            }
+
             if (!string.IsNullOrEmpty(sessionName))
             {
                 OnSessionDestroyed(sessionName);
```
paulhazen: Similar to the loop scenario - within a function some careful consideration of the conditional design can greatly reduce the indentation, and greatly increase the readability of the code. The above code for example could be rewritten like so:

```csharp
// Before removing the session from our local data, we need to inform the owner of the session that we've left the session, if we're not the owner
// TODO: Validate that this gets to the members/owners of the session in time, and that we haven't already deleted the local information needed to get session information
string sessionName = (string)data.ClientData;
Session localSession = GetSession(sessionName);
if (localSession == null || localSession.ActiveSession == null)
{
    Debug.LogError(
        $"Session Matchmaking (OnDestroySessionCompleteCallback): Could not find local Session and associated ActiveSession, so could not inform owner/members of destruction.");
    return;
}

ActiveSessionCopyInfoOptions copyOptions = new ActiveSessionCopyInfoOptions() { };

Result localCopyResult =
    localSession.ActiveSession.CopyInfo(ref copyOptions, out ActiveSessionInfo? outActiveSessionInfo);

// TODO: Add comment clarifying in plain language what question this conditional is answering
if (localCopyResult != Result.Success || !outActiveSessionInfo.HasValue ||
    !outActiveSessionInfo.Value.SessionDetails.HasValue)
{
    Debug.LogError(
        $"Session Matchmaking (OnDestroySessionCompleteCallback): Failed to copy local information for session {sessionName}, so could not inform owner/members of destruction. Result code {localCopyResult}");
    return;
}

if (!outActiveSessionInfo.Value.SessionDetails.Value.OwnerUserId.Equals(EOSManager.Instance
        .GetProductUserId()))
{
    // Inform the owner that we've left the session
    InformSessionOwnerWithMessage(sessionName, P2PLeavingSessionMessageElement);
}
else
{
    // We're the owner of the session, inform everyone that it was destroyed
    InformSessionMembers(sessionName, P2PSessionOwnerDestroyedSessionMessageElement);
}

if (!string.IsNullOrEmpty(sessionName))
{
    OnSessionDestroyed(sessionName);
}
```

Another thing is that I would add a comment before the more complicated conditional that you have - describing in plain language what question the conditional is answering. In my example re-write above, I added a "TODO" comment to that effect.

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -1862,5 +2021,419 @@ public void DeclineLobbyInvite()
...
+        }
+
+        /// <summary>
+        /// Whenever a user attempts to create a connection, this method handles their connection request.
+        /// By default, accept all incoming connections.
+        /// </summary>
+        /// <param name="data">Data, containing the product user id of the connecting request.</param>
+        private void OnIncomingSessionsConnectionRequest(ref OnIncomingConnectionRequestInfo data)
+        {
+            if (!(bool)data.SocketId?.SocketName.Equals(P2PSessionStatusSocketName))
```
paulhazen: This is a little confusing to my mind. It would be much cleaner to write:

```csharp
if (data.SocketId?.SocketName != P2PSessionStatusSocketName))
{
    /* ... */
}
```

Nothing is lost with this - as `String` objects use the same code path to determine equality when using `.Equals` and the `==` operator (since in this case no overloaded versions of `Equals` are used that employ culture-aware comparison logic).

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -1862,5 +2021,419 @@ public void DeclineLobbyInvite()
...
+                LocalUserId = EOSManager.Instance.GetProductUserId(),
+                RemoteUserId = data.RemoteUserId,
+                SocketId = socketId
+            };
+
+            Result result = EOSManager.Instance.GetEOSPlatformInterface().GetP2PInterface().AcceptConnection(ref options);
+
+            if (result != Result.Success)
+            {
+                Debug.LogErrorFormat("EOSSessionsManager (OnIncomingSessionsConnectionRequest): Eror while accepting connection, code: {0}", result);
```
paulhazen: Typo: "Eror" should be "Error".

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -1862,5 +2021,419 @@ public void DeclineLobbyInvite()
...
+                LocalUserId = EOSManager.Instance.GetProductUserId(),
+                RemoteUserId = data.RemoteUserId,
+                SocketId = socketId
+            };
+
+            Result result = EOSManager.Instance.GetEOSPlatformInterface().GetP2PInterface().CloseConnection(ref closeOptions);
+
+            if (result != Result.Success)
+            {
+                Debug.LogErrorFormat("EOSSessionsManager (OnIncomingSessionsDisconnect): Eror while closing connection, code: {0}", result);
```
paulhazen: Very small nit, and arguably a preference thing, but I notice that for formatted error logs you tend to use `LogErrorFormat` function, and when using the `Log` function you use the string format thing that starts the string literal with a `$` symbol.

Personally, I prefer to exclusively use the latter method - so that things are consistent in the code and there aren't two different kinds of string formatting happening in the same space.

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -1862,5 +2021,419 @@ public void DeclineLobbyInvite()
...
+            // Identify the owner of the session
+            // Send them a packet informing them of joining status
+
+            Session localSession = GetSession(localSessionName);
+
+            if (localSession == null)
+            {
+                Debug.LogErrorFormat("EOSSessionsManager (InformSessionOwnerWithMessage): No local session with name \"{0}\" was found.", localSessionName);
+                return;
+            }
```
paulhazen: Given the frequency with which this operation is done, you may want to consider changing the function signature of `GetSession` to be:

```csharp
public bool TryGetSession(string name, out Session session)
```

Then you could have the log statement happen inside that function, removing the need to repeat it every time after you call `GetSession`.

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -1862,5 +2021,419 @@ public void DeclineLobbyInvite()
...
+                // This session is owned by a server, so we can't message it using a product user id.
+                // TODO: How do we message it?
+                // This isn't exactly an error state, so we won't log a message, just going to return.
+                return;
+            }
+
+            // Quickly check; is this local user the owner of this session? If so, we shouldn't be informing ourself
+            if (EOSManager.Instance.GetProductUserId().Equals(ownerUserId))
+            {
+                // Oh nevermind, then! This user is the owner of this, and shouldn't message themself.
```
paulhazen, line 2230: In contrast with the situation earlier where I suggest reducing the nesting of the conditional - it seems you did it the way I was suggesting in this particular instance, and I think it makes for much clearer code.

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -1862,5 +2021,419 @@ public void DeclineLobbyInvite()
...
+                return;
+            }
+
+            ActiveSessionGetRegisteredPlayerCountOptions countOptions = new ActiveSessionGetRegisteredPlayerCountOptions() { };
+            uint playerCount = activeSession.GetRegisteredPlayerCount(ref countOptions);
+
+            Log($"EOSSessionsManager (InformSessionMembers): There are {playerCount} registered members in {localSession.Name}, informing users of {messageDetail} (excluding self)");
+
+            string messageToSend = string.Format(P2PInformSessionMessageFormat, copiedInfo.Value.SessionDetails.Value.SessionId, EOSManager.Instance.GetProductUserId(), messageDetail);
+            for (uint ii = 0; ii < playerCount; ii++)
```
paulhazen, line 2295: I recall you mentioning sometime when we were discussing things in-person that you had a reason for this - but I find the use of `ii` to be an odd choice for a loop variable. Could you elaborate on your thought process for doing so?

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -1862,5 +2021,419 @@ public void DeclineLobbyInvite()
...
+            Log($"EOSSessionsManager (HandleReceivedP2PMessages): Message parts: [{nameof(sessionId)}: {sessionId}] [{nameof(userId)}: {userId}] [{nameof(messageElement)}: {messageElement}]");
+
+            Session session = GetSessionById(sessionId);
+            ProductUserId messagingUserId = ProductUserId.FromString(userId);
+
+            if (session == null)
+            {
+                Debug.LogError($"EOSSessionsManager (HandleReceivedP2PMessages): Message received regarding sessionId {sessionId}, but no local sessions have that id. Message: {message}");
+                return;
+            }
```
paulhazen: This is another good example of careful organization of conditionals that helps reduce indentation and increase code clarity.

paulhazen (changes requested):
Lots of detailed feedback. Overall, I think these are great changes, but I think there are some areas of low-hanging fruit for improvement before the changes are merged.

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: Sure thing, space added. That's not something I had considered as part of code hygiene, but I'll do that moving forward.

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: It _appears_ as though you can have a socket name collision, as long as they're on different channels. Because `HandleReceivedP2PMessages` uses only the channel in the `ReceivePacketOptions` and not the socket to get incoming messages, it's most important that the channel is unique. Otherwise someone implementing this could receive messages meant for another socket in this method, or get messages meant for session management in another method.

I'm noticing in this comment that it's a little unclear, so we're gonna update this text a bit.

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: Sure, that's a significant linguistic difference. Do you think this new version is better?

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: You asking this might suggest that we need to make the comment include this information.

This notification id is for all incoming requests. You don't need one of these per peer, just the one to allow peers to connect.

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: Sure thing. A space has been added appropriately.

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: This one is entirely a mistake, it was configured while I was trying to get the log to stop complaining about an irrelevant communication. This is removed.

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: Yup, removed!

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: I'm not entirely convinced on this change, but I'm willing to float it by other people to see how they feel about the no-bracket continue statement.

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: This construction seems lofty enough to be worth lining it out like that, sure. Can you look at how now this is set up?

```
CloseConnectionsOptions closeOptions = new CloseConnectionsOptions()
{
    LocalUserId = EOSManager.Instance.GetProductUserId(), 
    SocketId = new SocketId() 
    { 
        SocketName = P2PSessionStatusSocketName 
    } 
};
```

Would SocketId here be good to be on one line, or is this preferable?

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: Flattening out like this definitely looks better. I'm _slightly_ concerned about pushing down the `OnSessionDestroyed` from not happening. I wanna do all the messaging to owner/members before getting to `OnSessionDestroyed`, but I still want it to run.

That said, I'll be working on validating that this behavior always works when it should. TODO!

Trying to iterate on the above design, it feels like you've arrived at the right place. Didn't make any alterations other than writing a comment.

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: This is definitely confusing! I borrowed it from `EOSPeer2PeerManager`, which does it this way.

Changed over to the way you've suggested here. Probably wanna update those other places, too? Maybe when those get reworks.

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: Couple of places had this mistake. `I Am Eror`, huhuhuhu. Fixed!

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: This was mostly following the conventions that were established in this file already. I'm very comfortable changing things over to the $"" string interpolation. Shall I pull the trigger on changing those here?

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: Mayyyyyybe... Here, it is an error. We're expecting that you should be in a session with this localSessionName. Not every time that we try to get a session does it result in an error.

That said I do generally prefer the TryGet model; it avoids null comparisons. Which isn't a big deal for the Session object, but is a big deal for Unity capital O Objects. Looking at the `GetSession` function, which existed before this PR, it kind of hilarious is just a wrapper around something actually doing TryGetValue. Since I'm the only consumer of this function, I'm comfortable changing it over... or adding another one? This is a place where the API of the sample would be changing, but nothing _in_ our samples currently relied on this.

WispyMouse: I've implemented the TryGetSession, but I'm not sold on putting the comment inside there.

WispyMouse:
> @paulhazen 

Thank you for your review! I learned a 'lil bit, and it's good to be kept honest on excessive if statement nesting. Feeling better about this stage of the work after implementing your review comments. Please resolve things you see fit as completely resolved, and continue discussion on things that need iteration.

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: Compared to `i`, `ii` is less frequent in code. That means if I use `ii` as an iterator, I can search for `ii` using `CTRL+F` and see more places where it's used. If I use `i`, then searching the text for `i` will get more hits that I don't care about.

It doesn't _stand_ for anything, like there's no acronym or abbreviation here. I just find `ii` more clear when trying to read code than using `i`.

Assets/Scripts/EOSSessionsManager.cs
paulhazen: That is a salient point - and clarifies my own understanding of the functionality.

Assets/Scripts/EOSSessionsManager.cs
paulhazen: Just to clarify - I was referring to the function `SubscribteToGameInvites()`, and how we should rename it in this PR to be `SubscribeToGameInvites()`, fixing the spelling error (likely a typo) within the function name as it existed before.

Assets/Scripts/EOSSessionsManager.cs
paulhazen: Good push-back - I approve of disagreement in this arena. I think it could go either way personally.

Assets/Scripts/EOSSessionsManager.cs
paulhazen: You know, I considered mentioning that one - and I think that is more of a subjective decision. I think _I_ would probably have done this, but that doesn't mean it's the "correct" way to do it.

Assets/Scripts/EOSSessionsManager.cs
paulhazen: I had similar thoughts regarding the `OnSessionDestroyed` call, but I wondered whether if the session name is empty or does not exist whether it makes sense to try and destroy that session or not. I suppose if it passes that test but fails early otherwise, you _would_ still want to call it, so that's a good observation.

Assets/Scripts/EOSSessionsManager.cs
paulhazen: I would say update those when changes are made to those pieces of code; otherwise leave them alone to limit the scope of the changes in the PR to reflect a focused set of changes.

Assets/Scripts/EOSSessionsManager.cs
paulhazen: I would pull the trigger on making those changes for new implementations. I understand how this is somewhat subjective, because other guidance regarding refactors dictates that you follow the format and styling that already exists within the file. 

I would clarify that strategy to say that you should always match existing style and formatting _if_ that style and formatting is consistent. If it's _already_ not consistent, then introduce changes that only include style and format decisions that are consistent internal to the changes being made (in other words, don't continue to use inconsistent styling, try to pick the best one shown in the file and follow that one everywhere your changes touch).

Assets/Scripts/EOSSessionsManager.cs
paulhazen: That's a fair observation. In similar circumstances, I have opted to include a `boolean` parameter in the `TryGet` function that determines whether an error is thrown. Typically, I default the value of that `boolean` parameter value to whichever is the most common use-case.

Assets/Scripts/EOSSessionsManager.cs
paulhazen: Interesting, I remember you mentioning that now. My only concern is that it is irregular enough that it might be something that other developers snag on - but only as a curiosity. However, that concern doesn't rise to a level where I would request it be changed.

paulhazen (approved):
Approving these changes. I have some topics of conversation with respect to some standards and code formatting things - but those are probably best brought up once these changes are to be merged into the main `development` branch, so I marked those conversations as resolved.

PR #764, opened 2024-07-03
chore (documentation, Lobby): Adding clarifying documentation mention subscription

WispyMouse:
Comment-only change set.

Adds comments to the various "AddNotify" methods indicating that subscription is required for them to work.


Assets/Scripts/EOSLobbyManager.cs
```diff
@@ -525,30 +525,59 @@ private bool IsLobbyNotificationValid(NotifyEventHandle handle)
         }
 
         // Helper method to keep the code cleaner
+
+        /// <summary>
+        /// Use to access functionality of [EOS_Lobby_AddNotifyLobbyUpdateReceived](https://dev.epicgames.com/docs/api-ref/functions/eos-lobby-add-notify-lobby-update-received)
+        /// The callback will only run if a listener is subscribed, which is done in <see cref="SubscribeToLobbyUpdates"/>.
+        /// </summary>
+        /// <param name="lobbyInterface">Handle to the lobby interface.</param>
+        /// <param name="notificationFn">Callback to receive notification when lobby member update is received</param>
```
paulhazen: This should be "Callback to receive notification when lobby update is received." (looks like a copy - paste mistake). Also missing return value comment.

Assets/Scripts/EOSLobbyManager.cs
```diff
@@ -525,30 +525,59 @@ private bool IsLobbyNotificationValid(NotifyEventHandle handle)
...
             var options = new AddNotifyLobbyUpdateReceivedOptions();
             return lobbyInterface.AddNotifyLobbyUpdateReceived(ref options, null, notificationFn);
         }
 
+        /// <summary>
+        /// Use to access functionality of [EOS_Lobby_AddNotifyLobbyMemberUpdateReceived](https://dev.epicgames.com/docs/api-ref/functions/eos-lobby-add-notify-lobby-member-update-received)
+        /// The callback will only run if a listener is subscribed, which is done in <see cref="SubscribeToLobbyUpdates"/>.
+        /// </summary>
+        /// <param name="lobbyInterface">Handle to the lobby interface.</param>
+        /// <param name="notificationFn">Callback to receive notification when lobby member update is received</param>
```
paulhazen: This comment block is missing the component for documenting the return value.

Assets/Scripts/EOSLobbyManager.cs
```diff
@@ -525,30 +525,59 @@ private bool IsLobbyNotificationValid(NotifyEventHandle handle)
...
             var options = new AddNotifyLobbyMemberUpdateReceivedOptions();
             return lobbyInterface.AddNotifyLobbyMemberUpdateReceived(ref options, null, notificationFn);
         }
 
+        /// <summary>
+        /// Use to access functionality of [EOS_Lobby_AddNotifyLobbyMemberStatusReceived](https://dev.epicgames.com/docs/api-ref/functions/eos-lobby-add-notify-lobby-member-status-received)
+        /// The callback will only run if a listener is subscribed, which is done in <see cref="SubscribeToLobbyUpdates"/>.
+        /// </summary>
+        /// <param name="lobbyInterface">Handle to the lobby interface.</param>
+        /// <param name="notificationFn">Callback to receive notification when lobby member update is received</param>
```
paulhazen: Another copy-paste error - it should be "Callback to receive notification when lobby member status update is received." This block also excludes comments documenting the return value.

Assets/Scripts/EOSLobbyManager.cs
```diff
@@ -525,30 +525,59 @@ private bool IsLobbyNotificationValid(NotifyEventHandle handle)
...
             var options = new AddNotifyLobbyMemberStatusReceivedOptions();
             return lobbyInterface.AddNotifyLobbyMemberStatusReceived(ref options, null, notificationFn);
         }
 
+        /// <summary>
+        /// Use to access functionality of [EOS_Lobby_AddNotifyLeaveLobbyRequested](https://dev.epicgames.com/docs/api-ref/functions/eos-lobby-add-notify-leave-lobby-requested)
+        /// The callback will only run if a listener is subscribed, which is done in <see cref="SubscribeToLobbyUpdates"/>.
+        /// </summary>
+        /// <param name="lobbyInterface">Handle to the lobby interface.</param>
+        /// <param name="notificationFn">Callback to receive notification when lobby member update is received</param>
```
paulhazen: Similar to comment blocks before: the notificationFn comment is incorrect, and comment describing the return value is missing.

paulhazen (changes requested):
See notes. The comments you add are incorrect in some cases, and incomplete in others. From a philosophical perspective one might argue that only the incorrect comments need to be corrected, but I feel that when we're adding comments to the code we should do so completely (in this case include the comment section for the function return value).

Assets/Scripts/EOSLobbyManager.cs
WispyMouse: Oh yea, this does have a return value. It's the intptr for where to find this notification for the purposes of subscription and unsubscription, right? I'll take a shot at communicating that, please review afterwards. `AchievementsInterface.AddNotifyAchievementsUnlocked` uses this line of text;

> handle representing the registered callback

So we'll use that.

`member` here is extra from copypaste, that text was removed.

Assets/Scripts/EOSLobbyManager.cs
WispyMouse: Added the return, using the same message as `AchievementsInterface.AddNotifyAchievementsUnlocked`.

Assets/Scripts/EOSLobbyManager.cs
WispyMouse: Return value added.

Should it be "Callback to receive notification when lobby member status update is received" or "Callback to receive notification when lobby member status is received" (no update) given the name of the function?

Assets/Scripts/EOSLobbyManager.cs
WispyMouse: Return value added, changed it to "when lobby member leave request is received."

WispyMouse:
> @paulhazen 

I still wish that GitHub had a way to respond to the message directly... Anyways, what do you think about these adjustments? Is this return statement appropriate?

Assets/Scripts/EOSLobbyManager.cs
paulhazen: I would say whichever makes the most sense to you - I think there are decent arguments to be made for either - so long as it indicates that it happens when the lobby member status has changed.

paulhazen: Feel free to mark this conversation as resolved.

paulhazen (approved):

arthur740212 (approved):

PR #769, opened 2024-07-08
feature (Sessions, Sample): Refresh UI, Standardized Logging

WispyMouse:
The main feature of this pull request is the UI refresh. When a member has received an appropriate message to refresh their UI, then it will now search up the associated ID and update it in the UI by running a callback. Is this callback made in a way that is compliant and appropriate with our samples?

Additionally per a conversation with Paul, I've standardized all the logging in the function. I'm trying out a stylistic choice; logs look like this:

`Debug.LogError($"{nameof(EOSSessionsManager)} ({nameof(OnSearchResultsReceived)}): searchHandle is null");`

Using `nameof(CLASSNAME)` and `nameof(FUNCTIONNAME)` makes it crystal clear where the logs come from at a glance, and feels more coherent than just saying "Sessions Matchmaking" or other phrases that were in use.

### DIFF
--- a/Assets/Scripts/StandardSamples/UI/MatchMaking/UISessionsMatchmakingMenu.cs
+++ b/Assets/Scripts/StandardSamples/UI/MatchMaking/UISessionsMatchmakingMenu.cs
@@ -373,12 +373,7 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
             {
                 UISessionEntry thisEntry = childTransform.GetComponent<UISessionEntry>();
 
-                if (thisEntry == null)
-                {
-                    continue;
-                }
-
-                if (thisEntry.RepresentedSession == null || thisEntry.RepresentedSession.Id != sessionId)
+                if (null == thisEntry || null == thisEntry.RepresentedSession || thisEntry.RepresentedSession.Id != sessionId)
                 {
                     continue;
                 }
--- a/Assets/Scripts/StandardSamples/UI/MatchMaking/UISessionsMatchmakingMenu.cs
+++ b/Assets/Scripts/StandardSamples/UI/MatchMaking/UISessionsMatchmakingMenu.cs
@@ -79,6 +79,8 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
             InviteFromVal.text = string.Empty;
 
             HideMenu();
+
+            GetEOSSessionsManager.UIOnSessionRefresh = OnSessionRefresh;
         }
 
         /*private void Start()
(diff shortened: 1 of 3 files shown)


Assets/Scripts/StandardSamples/UI/MatchMaking/UISessionsMatchmakingMenu.cs
```diff
@@ -432,5 +366,45 @@ public void HideMenu()
...
+
+                if (thisEntry == null)
+                {
+                    continue;
+                }
+
+                if (thisEntry.RepresentedSession == null || thisEntry.RepresentedSession.Id != sessionId)
+                {
+                    continue;
+                }
```
paulhazen: Tiny nit would be to change this to:

```csharp
if (null == thisEntry || thisEntry.RepresentedSession == null || thisEntry.RepresentedSession.Id != sessionId)
{
    continue;
}
```

paulhazen (approved):
Looks good!

Assets/Scripts/StandardSamples/UI/MatchMaking/UISessionsMatchmakingMenu.cs
WispyMouse: Sure, can do! I have noticed that other programmers seem to prefer these kind of single-if statement checks, whereas I've preferred multiple statements. In this particular case, though, I think it really does make sense to do it the way you have.

Only question is, why put `null == thisEntry` instead of `thisEntry == null`? The order of comparison matters in PowerShell, but in C# it shouldn't.

WispyMouse: > Totally! So the reason I put null first, is because it's not possible to assign a value to null. In general, I put the immutable thing on the left-hand side. So it's the same for string literals and number values. The reason for all of this is because some languages allow assigning a value in a conditional, and if I do it the other way around, I might make the mistake of having the conditional look like this: if (thisEntry = null), which could be a difficult error to diagnose or hunt down. So for the sake of forming a habit that avoids that potential problem in any language, I put the literal value to the left.

(from conversation with Paul)

paulhazen: We discussed this offline, but for the sake of the history: Putting the immutable value on the left-hand side of the conditional prevents accidental assignment happening in the conditional. Some languages allow for `thisEntry = null` inside the `if` statement. By forming a habit of putting the immutable thing first, I avoid making that mistake in languages that support doing so.

PR #772, opened 2024-07-09
feature (Sessions): Rearranging Manager Code into Regions

WispyMouse:
This change set seeks to rearrange the methods inside EOSSessionsManager to group them by purpose. The types of activity someone would want to manage with the Manager are categorized in to regions, and inside those regions things are generally organized by the flow of when they would be called.

Notes:

- The **ONLY** changes made here are moving functions around. There is no change to the contents of those functions or the comments surrounding them, just moving them around.
- Git's comparison is probably going to look atrocious, it can't really reconcile this kind of reorganization.
- With regards to variables, only Peer2Peer Messaging Variables and Peer2Peer UX Variables are in regions, and those are at the top of the class. We *could* organize the variables by the group they apply to, but I think it's appropriate to leave variables at the top of classes. I do not believe putting the remaining variables in to regions would be significantly beneficial, but I could be convinced otherwise.
- `SetJoinInfo` and `OnSetPresenceCompleteCallback` are inside the `Manager Set Up And Utility` region because they are not used in this class. They are only used by the `UIP2PTransportMenu`, which suggests these should be in another class.
- `MakeSessionHandleFromSearch` is unused and unimplemented. Should probably be removed or implemented properly.
- Notifications (things that aren't called by the Plugin but are invoked by the SDK on event triggering) are subsectioned in to regions.

### DIFF
--- a/Assets/Scripts/EOSSessionsManager.cs
+++ b/Assets/Scripts/EOSSessionsManager.cs
@@ -1487,20 +1487,6 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
             return string.Format("{0}{1}", JOINED_SESSION_NAME, JoinedSessionIndex);
         }
 
-        public void OnJoinGameAcceptedListener(ref JoinGameAcceptedCallbackInfo data) // OnPresenceJoinGameAcceptedListener
-        {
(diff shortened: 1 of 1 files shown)


paulhazen (approved):
Looks good - I am assuming this is _just_ reorganizing into regions, no functional change. I'll take your word for that since it is a merge that goes into a feature branch.

andrew-hirata-playeveryware (approved):

PR #775, opened 2024-07-10
Editor UI Improvements

paulhazen:
This PR encompasses some improvements to the editor UI during MDF.

The label for input fields is automatically set to be long enough to read:

image

As for code improvements, it removes the need for `ConfigEditor` classes for editor specific configurations, removing a ton of redundancies, and decreasing the size of the codebase.


Assets/Plugins/Source/Editor/ConfigEditors/ConfigEditor.cs
```diff
@@ -22,29 +22,146 @@
...
+        /// Use reflection to retrieve a collection of fields that have been
+        /// assigned custom ConfigFieldAttribute attributes, grouping by group,
+        /// and sorting by group.
+        /// </summary>
+        /// <returns>A collection of config fields.</returns>
+        private static IOrderedEnumerable<IGrouping<int, (FieldInfo FieldInfo, ConfigFieldAttribute FieldDetails)>> GetFieldsByGroup()
+        {
+            var returnValue = typeof(T).GetFields(BindingFlags.Public | BindingFlags.Instance)
+                .Where(field => field.GetCustomAttribute<ConfigFieldAttribute>() != null)
+                .Select(info => (info, info.GetCustomAttribute<ConfigFieldAttribute>()))
```
WispyMouse, line 71: In the parameterless constructor, it attempts to get this `ConfigGroupAttribute`, and then there's a null check against it. Here, it selects it and uses it without null checking. Is that okay? Is there ever a circumstance where this can be null, and that breaks the rest of this LINQ query?

paulhazen: Nope - the reason is that the `Where` clause excludes any results for which the field is `null`, meaning only those that are not null are subsequently processed by the `Select` clause.

Assets/Plugins/Source/Editor/Configs/LibraryBuildConfig.cs
```diff
@@ -21,10 +21,19 @@
  */
 namespace PlayEveryWare.EpicOnlineServices.Editor.Config
 {
+    using System;
+
+    [Serializable]
```
WispyMouse, line 26: I noticed this EditorConfig is serializable, but not `PackagingConfig`. Should all EditorConfig files be marked with this?

Assets/Plugins/Source/Editor/EditorWindows/EOSPluginSettingsWindow.cs
```diff
@@ -104,10 +105,7 @@ protected override void RenderWindow()
             {
                 foreach (var configurationSectionEditor in configEditors)
                 {
-                    GUILayout.Label(configurationSectionEditor.GetLabelText(), EditorStyles.boldLabel);
-                    GUIEditorUtility.HorizontalLine(Color.white);
-                    configurationSectionEditor.Render();
-                    EditorGUILayout.Space();
+                    configurationSectionEditor.RenderAsync();
```
WispyMouse: This has a signature of `async Task`. Should this be `await`'d?

Assets/Plugins/Source/Editor/Utility/GUIEditorUtility.cs
```diff
@@ -24,12 +24,13 @@ namespace PlayEveryWare.EpicOnlineServices.Editor.Utility
 {
     using System;
     using System.Collections.Generic;
-    using System.Security.Cryptography;
     using UnityEditor;
     using UnityEngine;
 
     public static class GUIEditorUtility
     {
+        private const float MaximumButtonWidth = 100f;
```
WispyMouse, line 32: nit: SCREAMING_SNAKE_CASE

Assets/Plugins/Source/Editor/ConfigEditors/ConfigEditor.cs
WispyMouse: Ah, I was thrown off a bit by the LINQ naming. The `.Where` uses `field` where the `.Select` uses `info`, so I read that as different comparisons. Seems good!

WispyMouse (approved):
UX appears to function, the code is clean. Seems good to me!

Assets/Plugins/Source/Editor/Utility/GUIEditorUtility.cs
paulhazen: I'm split on this. @andrew-hirata-playeveryware Do we have a final answer on our standards for this? I recall making a comment regarding standards being different between C# and C++, but I do not recall what the outcome of that discussion was.

Assets/Plugins/Source/Editor/EditorWindows/EOSPluginSettingsWindow.cs
paulhazen: Interestingly, this is not seen as an issue because the list of configEditors implements the `IConfigEditor` interface, and it is not possible to specify that the implementation of that interface should be async, it just specifies that it returns a task. To ameliorate the situation, I used the discard pattern to avoid that - even though it doesn't cause problems during runtime, and doesn't throw warnings in the editor or during compilation.

Assets/Plugins/Source/Editor/Configs/LibraryBuildConfig.cs
paulhazen: For the sake of clarity, it should be. Functionally, the `[Serializable]` attribute is actually inherited by extending the `EditorConfig` class, which itself has that attribute. For the sake of consistency however, I have added it here.

Assets/Plugins/Source/Editor/Configs/LibraryBuildConfig.cs
```diff
@@ -21,10 +21,19 @@
...
+    [ConfigGroup("Platform Library Build Settings")]
     public class LibraryBuildConfig : EditorConfig
     {
+        [FilePathField("MSBuild path", "exe")]
         public string msbuildPath;
+
+        [FilePathField("Make path", "")]
```
paulhazen: No extension for Linux, but it _does_ support having make on windows, so this should probably be `*`.

Assets/Plugins/Source/Editor/Configs/LibraryBuildConfig.cs
andrew-hirata-playeveryware: I agree with Paul.

andrew-hirata-playeveryware (approved):

PR #777, opened 2024-07-11
refactor: remove `LoadDelegatesWithReflection` and `LoadDelegatesByHand`

andrew-hirata-playeveryware:
This code was originally added to support other runtime binding systems that ended up not being used. They are safe to remove because they are not used on any platform, and there is little to no chance that they will be used in the future.


WispyMouse (approved):

paulhazen (approved):
Yay! Less code 😁

PR #778, opened 2024-07-11
feat: Add ability configure the plugin to not unload the EOS SDK on shutdown.

andrew-hirata-playeveryware:
# Overview
This feature is the result of work done to assist a plugin user that was experiencing hangs in the editor during play. In a sense, it is kind of a bug fix, but in order to fix that bug, and given that the bug shouldn't always be 'fixed', this issue is being resolved via a new 'feature'.

Enabling this define will work around issues where the Editor can hang during shutdown / startup while playing in the Unity editor. This workaround, while useful, also makes it necessary to reboot the editor if any EOS boot strap based configuration needs to be changed. It also carries with it the potential for "leaked" EOS State to carry over between play in editor runs.

### DIFF
--- a/Assets/Plugins/Source/Core/EOSManager.cs
+++ b/Assets/Plugins/Source/Core/EOSManager.cs
@@ -29,6 +29,16 @@
 #define EOS_CAN_SHUTDOWN
 #endif
 
+//#define EOS_DO_NOT_UNLOAD_SDK_ON_SHUTDOWN
+#if (UNITY_EDITOR_OSX || UNITY_EDITOR_LINUX)
+#define EOS_DO_NOT_UNLOAD_SDK_ON_SHUTDOWN
+#endif
+
+#if !EOS_DO_NOT_UNLOAD_SDK_ON_SHUTDOWN
+// If defined, the EOS SDK will be unloaded
+#define EOS_UNLOAD_SDK_ON_SHUTDOWN
+#endif
+
 #if !UNITY_EDITOR
 #define USE_STATIC_EOS_VARIABLE
 #endif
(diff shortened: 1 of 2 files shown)


Assets/Plugins/Source/Core/EOSManager.cs
```diff
@@ -29,6 +29,16 @@
 #define EOS_CAN_SHUTDOWN
 #endif
 
+//#define EOS_DO_NOT_UNLOAD_SDK_ON_SHUTDOWN
```
paulhazen, line 33: Would probably be good to remove this. I assume it was here during development for testing purposes.

paulhazen (changes requested):
I think that a way to significantly improve the clarity of these changes would be to introduce a private member to `EOSManager`:

```csharp
private static readonly bool s_shouldUnloadSDKOnShutdown = 
    #if EOS_DO_NOT_UNLOAD_SDK_ON_SHUTDOWN
        false
    #else
        true
    #endif
        ;
```

Then you can use the value lower in the code file to determine whether to load or unload. I think this implementation would lead to greater clarity, and less of a need to maintain scripting-define conditionals - which at times can become complicated.

Assets/Plugins/Source/Core/EOSManager.cs
```diff
@@ -635,24 +645,32 @@ public void Init(IEOSCoroutineOwner coroutineOwner, string configFileName)
...
-                    LoadEOSLibraries();
 
-                    var secondTryResult = InitializePlatformInterface(loadedEOSConfig);
-                    Debug.Log($"EOSManager::Init: InitializePlatformInterface: initResult = {secondTryResult}");
+                    if (initResult != Result.Success)
+                    {
+#if UNITY_EDITOR
 
-                    if (secondTryResult != Result.Success)
+#if EOS_UNLOAD_SDK_ON_SHUTDOWN
```
WispyMouse: It surprises me to see a region named `EOS_UNLOAD_SDK_ON_SHUTDOWN` surround a statement, and then immediately afterwards a call to `UnloadAllLibraries` is made outside it.

Assets/Plugins/Source/Core/EOSManager.cs
```diff
@@ -1640,13 +1658,20 @@ public void OnApplicationShutdown()
...
                     Debug.Log("Shutting down the platform interface.");
                     ShutdownPlatformInterface();
+#endif
+
                     SetEOSPlatformInterface(null);
+
+
 #endif
 #if UNITY_EDITOR
+#if EOS_UNLOAD_SDK_ON_SHUTDOWN
```
WispyMouse: This relates to a previous comment; here, you do put the `UnloadAllLibraries` inside the `#if EOS_UNLOAD_SDK_ON_SHUTDOWN`. Should it be called above?

docs/c_sharp_defines.md
```diff
@@ -0,0 +1,10 @@
+## Overview
```
WispyMouse, line 1: Oh exciting! I'm looking forward to expanding this list.

docs/c_sharp_defines.md
```diff
@@ -0,0 +1,10 @@
...
+This document lists the "public" conditional compilation directive defines ("Preprocessor" defines) that control different behavior of the Epic Online Services for Unity Plugin.
+
+## Defines
+### `EOS_DO_NOT_UNLOAD_SDK_ON_SHUTDOWN`
```
WispyMouse, line 5: Should we mention how this is automatically defined in the OSX and Linux editors? Should we define what conditions defines are automatically made in, in this document?

Assets/Plugins/Source/Core/EOSManager.cs
andrew-hirata-playeveryware: I actually leave these in so that it's easier for plugin developers to turn it on and off quickly without launching the editor.

Assets/Plugins/Source/Core/EOSManager.cs
andrew-hirata-playeveryware: That's fair; I original kept it there because `UnloadAllLibraries` doesn't actually unload all the libraries in the editor, but instead just ensures that dispose is called correctly on any of the `DLLHandles` that we have loaded, which nominally should always be called. That being said, I can make it all be inside the `EOS_UNLOAD_SDK_ON_SHUTDOWN` because that would be clearer.

Assets/Plugins/Source/Core/EOSManager.cs
andrew-hirata-playeveryware: In light of making the code less confusing, I'm going to change the code above.

docs/c_sharp_defines.md
andrew-hirata-playeveryware: Oh we should!

WispyMouse (approved):

Assets/Plugins/Source/Core/EOSManager.cs
```diff
@@ -181,6 +189,14 @@ public partial class EOSSingleton
             // Need to keep track for shutting down EOS after a successful platform initialization
             static private bool s_hasInitializedPlatform;
 
+            private static readonly bool s_eosUnloadSdkOnShutdown =
```
paulhazen: Is this consistent with our standards regarding initialisms?

paulhazen (approved):

PR #781, opened 2024-07-13
refactor: Move PackageTemplate files to new upm layout directory

andrew-hirata-playeveryware:
These changes move the PackageTemplate directory files to the new UPM layout directory.


arthur740212 (approved):

paulhazen (commented):
Am I understanding correctly that the `com.playeveryware.eos` directory is a temporary directory that exists until everything has been moved into it, at which point its contents will become the new root and that directory name will be removed?

It seems that most UPMs are deployed as a tarball with the name `com.[company_name].[product].tgz`, but the contents of that tarball contain no such directory.

andrew-hirata-playeveryware:
@paulhazen It's not temporary: the end goal is to move all the parts that we plan to ship into that directory but keep our own Unity project so that we can still have a shared Unity project to work with.

paulhazen (approved):

PR #783, opened 2024-07-15
docs (Sessions): Commenting Functions

WispyMouse:
This set of changes is just comments on the functions and regions of the code. Minimal amounts of reorganization or renaming has been done. There should be no functional changes.

There's lots of TODOs to remind myself of what reworks are needed for each function. This was a useful exercise for determining things that I think need to change at a glance, and was a great opportunity to take inventory.

I would like feedback on the commenting style and details, if anyone has any. I'll merge this over by midday tomorrow.

### DIFF
--- a/Assets/Scripts/EOSSessionsManager.cs
+++ b/Assets/Scripts/EOSSessionsManager.cs
@@ -433,7 +433,7 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
 
         /// <summary>
         /// Messages about sessions follow this format, starting with <see cref="P2P_INFORM_SESSION_MESSAGE_BASE"/>.
-        /// {0} - The back-end <see cref="Session.Id"/> of the session to message about.
+        /// {0} - The EOS Game Services <see cref="Session.Id"/> of the session to message about.
         /// {1} - The <see cref="ProductUserId"/> of the user sending the message.
         /// {2} - Additional information about the message.
         /// <see cref="P2P_JOINING_SESSION_MESSAGE_ELEMENT"/>
@@ -596,7 +596,7 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
                 return;
             }
 
-            SubscribteToGameInvites();
+            SubscribeToGameInvites();
             SubscribeToSessionMessageConnectionRequests();
 
             CurrentInvite = null;
@@ -987,7 +987,7 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
         #region Local Session Lookup and Search
 
         /* These functions are separated out to highlight that these functions are entirely local.
-         * While they may call in to the EOS SDK, none of these functions call the Epic Online Services backend.
+         * While they may call in to the EOS SDK, none of these functions call the EOS Game Services backend.
          * */
 
         /// <summary>
@@ -1110,7 +1110,7 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
         /// <summary>
         /// Tries to get a local Session that is joined with the provided id.
         /// </summary>
-        /// <param name="id">The id of the Session to try to get. This is the back-end id, shared for all users.</param>
+        /// <param name="id">The id of the Session to try to get. This is the EOS Game Services id, shared for all users.</param>
         /// <param name="session">Out parameter with the Session, if found.</param>
         /// <returns>True if a Session with the provided name is found.</returns>
         public bool TryGetSessionById(string id, out Session session)
@@ -1137,7 +1137,7 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
         /// </summary>
         /// <param name="sessionId">
         /// The id of the Session to try to get.
-        /// This is the back-end id, shared for all users.
+        /// This is the EOS Game Services id, shared for all users.
         /// Can only get results from the most recent search.
         /// </param>
         /// <returns>Session Details handle, if found.</returns>
@@ -1244,7 +1244,7 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
         /// It won't be able to find private invite only Sessions.
         /// <see cref="OnFindSessionsCompleteCallback"/> is called with the results.
         /// A handle for retriving results will be stored in <see cref="CurrentSearch"/>.
-        /// This is the back-end id, shared for all users.
+        /// This is the EOS Game Services id, shared for all users.
         /// 
         /// TODO: This should also have a callback or return type to indicate successful search attempt.
         /// </summary>
@@ -1391,7 +1391,7 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
         }
 
         /// <summary>
-        /// Identifies a local session by its <paramref name="localSessionName"/>, gets its back end <see cref="Session.Id"/>,
+        /// Identifies a local session by its <paramref name="localSessionName"/>, gets its EOS Game Services <see cref="Session.Id"/>,
         /// and then attempts to use the Session search API to look for this Session on the Epic Online Services back end.
         /// If it is able to find it, then a UI refresh action is called to inform the UI to update the Session's displayed information.
         /// While similar to <see cref="SearchById(string)"/>, this function uses <see cref="P2PSessionRefreshSessionSearch"/> instead of <see cref="CurrentSearch"/>,
@@ -1519,7 +1519,7 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
                 return;
             }
 
-            // Now that we have the back-end session information, update the existing session
+            // Now that we have the EOS Game Services session information, update the existing session
             if (!TryGetSessionById(sessionInfo.Value.SessionId, out Session existingLocalSession))
             {
                 Log($"{nameof(EOSSessionsManager)} ({nameof(OnRefreshSessionFindSessionsCompleteCallback)}): Successfully queried Epic Online Services for Session, but was unable to find a local session with {nameof(Session.Id)} \"{sessionInfo.Value.SessionId}\".");
@@ -1540,7 +1540,7 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
         /// Indicates that a Session should be "destroyed".
         /// This informs Epic Game Services that this local user is leaving a Session.
         /// <see cref="OnDestroySessionCallback"/> runs with results, which should then locally remove all of the Session's information.
-        /// If this user is the Owner of the Session, the Session is destroyed on the back end, and every member should leave it.
+        /// If this user is the Owner of the Session, the Session is destroyed with EOS Game Services, and every member should leave it.
         /// </summary>
         /// <param name="name">The local name of the Session to destroy.</param>
         public void DestroySession(string name)
@@ -1817,7 +1817,7 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
         /// This can only find Sessions that have their <see cref="Session.PermissionLevel"/> set to either
         /// <see cref="OnlineSessionPermissionLevel.JoinViaPresence"/> or <see cref="OnlineSessionPermissionLevel.PublicAdvertised"/>.
         /// </summary>
-        /// <param name="sessionId">The back end Session Id.</param>
+        /// <param name="sessionId">The EOS Game Services Session Id.</param>
         private void JoinPresenceSessionById(string sessionId)
         {
             JoinPresenceSessionId = sessionId;
@@ -1920,7 +1920,7 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
          * */
 
         /// <summary>
-        /// Starts an existing Session in EOS Game Services. This is not used to create new Sessions.
+        /// Starts an existing Session with EOS Game Services. This is not used to create new Sessions.
         /// This is mostly used for standardizing a state across clients,
         /// but if <see cref="Session.AllowJoinInProgress"/> is set to false, then users can only join it if it hasn't been started.
         /// This will set the <see cref="Session.SessionState"/> to <see cref="OnlineSessionState.Starting"/>,
@@ -1987,7 +1987,7 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
         }
 
         /// <summary>
-        /// Ends an existing Session in EOS Game Services. This is not used to leave or destroy Sessions.
+        /// Ends an existing Session with EOS Game Services. This is not used to leave or destroy Sessions.
         /// A Session does not need to "End" in order to be left or destroyed.
         /// This is mostly used for standardizing a state across clients,
         /// but if <see cref="Session.AllowJoinInProgress"/> is set to false, then users can only join it if it hasn't been started.
@@ -2021,7 +2021,7 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
         /// When successful, the Session owner will attempt to inform Session members of the state change.
         /// </summary>
         /// <param name="data">
-        /// Callback information about the attempted end.
+        /// Callback information about the attempted end operation.
         /// <see cref="EndSessionCallbackInfo.ClientData"/> must contain the local name of the Session that ended.
         /// </param>
         private void OnEndSessionCompleteCallback(ref EndSessionCallbackInfo data)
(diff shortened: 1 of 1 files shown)


Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -348,17 +351,40 @@ public class EOSSessionsManager : IEOSSubManager
         private SessionDetails JoiningSessionDetails = null;
         private ulong JoinedSessionIndex = 0;
 
+        /// <summary>
+        /// When joining a Session you did not create, the Session's name is not shared with you.
```
paulhazen: nit: It would be clearer to say simply, "...the Session name..."
I would also rephrase to avoid use of any pronouns (ie "you").

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -348,17 +351,40 @@ public class EOSSessionsManager : IEOSSubManager
...
+        /// Instead the Session is given a local name, starting with this, when <see cref="GenerateJoinedSessionName(bool)"/> is called.
+        /// This local name doesn't have to follow any particular formatting, but does need to be locally unique and consistent.
```
paulhazen: nit: Actually, as a matter of habit I try to avoid using contractions at all. However, sometimes it is unavoidable.

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -348,17 +351,40 @@ public class EOSSessionsManager : IEOSSubManager
...
         private const string BUCKET_ID = "SessionSample:Region";
 
+        /// <summary>
+        /// When creating a Session or searching for one, this is the attribute used to filter for buckets.
+        /// In these sample functions, <see cref="CreateSession(Session, bool, Action)"/> sets the BucketId
+        /// in <see cref="CreateSessionModificationOptions"/> as well as an <see cref="SessionAttribute"/>.
+        /// </summary>
         private const string EOS_SESSIONS_SEARCH_BUCKET_ID = "bucket";
+
+        // TODO: All three of these are unused, remove?
```
paulhazen: nit: I would categorize this as a `NOTE:`, the `TODO:` version would be "Confirm that the three proceeding values are unused, and if so remove them."

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -348,17 +351,40 @@ public class EOSSessionsManager : IEOSSubManager
...
         private const string EOS_SESSIONS_SEARCH_EMPTY_SERVERS_ONLY = "emptyonly";
         private const string EOS_SESSIONS_SEARCH_NONEMPTY_SERVERS_ONLY = "nonemptyonly";
         private const string EOS_SESSIONS_SEARCH_MINSLOTSAVAILABLE = "minslotsavailable";
 
         // UI Parameterized Callbacks
+        // TODO: None of these are subscribed to, several of them aren't implemented fully. Remove? Implement more fully?
+        // UIOnSessionCreated is actually utilized
```
paulhazen: I'd be more specific with respect to which you mean when you say, "these," particularly since you follow up to specify one in particular. Simpler would be to just limit the note to specifically reference the ones it applies to, or be more verbose.

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -377,6 +403,12 @@ public class EOSSessionsManager : IEOSSubManager
 
         #region Peer2Peer Messaging Variables
 
+        /* The Sessions Interface does not provide a way to notify Session Owners and Members when things change,
+         * such as users joining, leaving, or the session starting, ending, being modified, or bieng destroyed.
```
paulhazen: spelling: bieng = being

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -458,14 +494,21 @@ public class EOSSessionsManager : IEOSSubManager
 
         /// <summary>
         /// After a successful <see cref="RefreshSession(string)"/> and <see cref="OnRefreshSessionFindSessionsCompleteCallback(ref SessionSearchFindCallbackInfo)"/> callback call,
-        /// this is run to inform UI of the updated Session information.
+        /// this is run to inform UI of the updated Session information. This should be subscribed to by your program's UI.
```
paulhazen: A bit redundant. Did you intend to replace the comment instead of amending it? Either way, capitalization.

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -458,14 +494,21 @@ public class EOSSessionsManager : IEOSSubManager
...
         /// </summary>
         public Action<Session, SessionDetails> UIOnSessionRefresh { get; set; }
 
         #endregion
 
         #region Manager Set Up And Utility
 
+        /* This region contains meta-management for this Manager class.
```
paulhazen: For multi-line block comments, always start the block with an empty comment line, and end with an empty comment line:

```csharp
/*
 * REGION NOTES:
 * 
 * This would be an example (the header of "REGION NOTES" isn't prescriptive, 
 * but can be a useful way to indicate the intent of the comment block that is
 * not otherwise adjacent to a function or class definition).
 */
```

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -494,6 +537,11 @@ internal static void Log(string toPrint)
             UnityEngine.Debug.Log(toPrint);
         }
 
+        /// <summary>
+        /// Checks for new Peer2Peer messages for Session management, as well as updating the OnlineSessionState of ActiveSessions.
+        /// This Manager is not a MonoBehaviour, so this function should be run by the program every update.
```
paulhazen, line 553: The comment is correct, and this is certainly outside the scope of the task of adding comments to existing code - but it brings up a point that might otherwise not have been noticed: is it possible that `EOSSessionsManager` _should_ be a `MonoBehaviour` (which would obviate the need to make this comment)?

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -544,14 +596,20 @@ public void OnLoggedIn()
...
 
             CurrentInvite = null;
 
             userLoggedIn = true;
         }
 
+        /// <summary>
+        /// Unsubscribes from invitations and Peer2Peer message requests and clears internal states.
+        /// This can be run before or after a successful logout, or preparing to shut down the application.
+        /// This must be called in order to remove Native notifications.
```
paulhazen: It may be wise to be clearer regarding what Native notifications means in this context.

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -574,13 +632,31 @@ public void OnLoggedOut()
             userLoggedIn = false;
         }
 
+        /// <summary>
+        /// Unsubscribes from invitations and Peer2Peer message requests.
+        /// This should be called before the application finishes shutting down in order to properly remove Native notifications.
+        /// TODO: This function has no references, and is incomplete in its implementation. This will likely be removed.
```
paulhazen, line 649: I think more properly it _should_ be called, as (just going based off of the contents of the function) it seems to be freeing some resources. I have noticed that this seems to be the case with several of the managers.

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -574,13 +632,31 @@ public void OnLoggedOut()
...
+        /// </summary>
+        [Obsolete]
```
paulhazen, line 651: Neat! I didn't know this was something that could be done. Thanks for teaching me something new! 😀

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -574,13 +632,31 @@ public void OnLoggedOut()
...
             UnsubscribeToSessionMessageConnectionRequests();
         }
 
+        /// <summary>
+        /// Informs the <see cref="UIInterface"/> that a UI Event has been processed.
+        /// Some functions in the EOS SDK return a UIEventId. When this happens, it is required that this function be called to released by the SDK.
+        /// For example, if you are subscribed to <see cref="OnJoinSessionAcceptedListener(ref JoinSessionAcceptedCallbackInfo)"/>, after a successful call to <see cref="SessionsInterface.JoinSession(ref JoinSessionOptions, object, OnJoinSessionCallback)"/>,
+        /// the notification function will receive a <see cref="JoinSessionAcceptedCallbackInfo.UiEventId"/>.
+        /// You can use this event id to create a <see cref="SessionDetails"/> object by calling <see cref="MakeSessionHandleByEventId(ulong)"/>.
+        /// After you've done so, or if you aren't going to use this handle, you must call this function to release the handle in the EOS SDK.
```
paulhazen: An example of how to rephrase to remove pronouns:

```csharp
/// For example, if a subscription is made to <see cref="OnJoinSessionAcceptedListener(ref JoinSessionAcceptedCallbackInfo)"/>, after a successful call to <see cref="SessionsInterface.JoinSession(ref JoinSessionOptions, object, onJoinSessionCallback)"/>, 
/// the notification function will receive a <see cref="JoinSessionAcceptedCallbackInfo.UiEventId"/>.
/// This event id can be used to create a <see cref="SessionDetails"/> object by calling <see cref="MakeSessionHandleByEventId(ulong)"/>.
/// After doing so, this function must be called to release the handle within the EOS SDK.
```

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -574,13 +632,31 @@ public void OnLoggedOut()
...
+        /// Most of the functions in the EOS SDK that require this will mention it in the comments of the callback function. Anything with a UiEventId should have this called.
+        /// 
+        /// TODO: This is hard coded to only handle the <see cref="JoinUiEvent"/> case. It must be changed to also accept the event's id in order to function properly.
```
paulhazen, line 668: nit: Another example where the contraction can be removed without altering the meaning of the comment.

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -596,11 +672,25 @@ private void AcknowledgeEventId(Result result)
...
+        /// Sets the current user's Social Presence.
+        /// 
+        /// TODO: This is not called by this manager. It either should be utilized, or moved to another manager, perhaps some sort of EOSPresenceManager.
+        /// </summary>
+        /// <param name="joinInfo">If provided, this string will be displayed to users in the Social Overlay as your current presence.
+        /// This can be used to indicate things like current level, looking for party status, etc.</param>
+        /// <param name="onLoggingOut">Indicates if the user is currently attempting or has already logged out.
+        /// If true, then if there's an error while trying to change your presence, the error will be ignored.
+        /// 
+        /// TODO: The error handling in both scenarios is almost identical. Perhaps an error state should be returned in the situation where this is false and an error is encountered.</param>
```
paulhazen: If you choose to have line breaks within your comment tags, you should put the open and close tags on their own lines.

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -965,6 +1151,17 @@ public SessionDetails MakeSessionHandleFromSearch(string sessionId)
 
         #region Online Session Lookup and Search
 
+        /// <summary>
+        /// Performs an online Search to look up a Session, where all attributes provided are met.
+        /// <see cref="OnFindSessionsCompleteCallback"/> is called with the results.
+        /// A handle for retriving results will be stored in <see cref="CurrentSearch"/>.
+        /// 
+        /// TODO: I feel like this should run a callback or return a Result in case of errors when trying to set up the search.
```
paulhazen: Particularly for comments that contain personal statements or opinions, you should either indicate that it is a `NOTE:` to yourself or rephrase the comment such that it is more objective. As it is written, it is not really a `TODO:` item that is written for any developer to take action on.

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -965,6 +1151,17 @@ public SessionDetails MakeSessionHandleFromSearch(string sessionId)
...
+        /// TODO: Only finds up to 10 results. Should have some method for finding next sets of results.
```
paulhazen, line 1181: This is a great example of what I think of as an appropriate `TODO:` comment. It clearly states what needs to be changed, and gives context for why.

paulhazen (approved):
I ran out of time to go through everything, apologies for being incomplete. Any comment I left that starts with a `nit:` you can feel free to take it or leave it and mark the conversation as resolved once you decide what to do with the feedback.

I'm approving, but I _do_ have some things I would change in the comments if this was being merged into `development`, so I'm depending on the requirement that conversations be resolved prior to merging to serve as a gating mechanism.

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: Noted! I am definitely awkward with pronouns in these comments, and it seems really easy to rephrase those away.

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: "doesn't" easily becomes "does not". But "Session's" is a possessive, is that still a contraction?

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: This is definitely meant to be followed up on, so I'll rephrase this to be more like a TODO comment.

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: Specified. I decided the clearest way to do that was to put `UIOnSessionCreated` in to its own singular spot, make a line break, and then comment about the following being unimplemented.

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: Good catch, typo fixed.

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: I don't see the capitalization fault, though I do see a "you/your" pronoun usage to replace.

I want a clear way to communicate, HEY USER! Subscribe to this or you're missing out!

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: Sure, that's a simple pattern to implement. I'll put that in all the regional comments.

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: Mayyyyyyybe? I don't think it should be, because none of the other Manager classes from the samples are MonoBehaviours. Unless it is okay to break that pattern? Either way it would change how this particular manager needs to be implemented.

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: Sure, I think the easiest way to do that here is just to point at OnLoggedIn, so changed the comments for that.

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: This already uses the word "should". Would this be better as "must"?

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: Handy tag is handy! ^o^ You can also use it to point users towards the new function that should be called, in the argument for this attribute.

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: Seems like a clear rewrite, I'll use most of your text. I do think the "or if you aren't going to use this handle" is important, so I preserved it in the form of "or if the handle will not be utilized". Otherwise I think it implies "after using this call Acknowledge" rather than what is really intended of "after it's not useful then call Acknowledge".

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: This contraction got eliminated with the rephrase, but I'll note to use "you have" if this ever comes up again.

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: That seems like a reasonable and clean guideline, will implement that.

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: I'm willing to put on my executive action pants on, and restate this to "this should be done" rather than "I feel like this should".

WispyMouse:
All "you"-s have been replaced with language about "the local user" or "the application".

PR #784, opened 2024-07-15
More Unit Tests!

paulhazen:
Adds Unit Tests to the project.

This is still a draft for the following reasons:

~- A determination needs to be made regarding how to handle package creation in light of the additions made. (This has simply not been explored).~

_The reason above has been determined to not be an issue. A package was created, and no errors are generated when loading the resulting package into a fresh unity project (Unit Tests are not shipped with the package)._

~- A more complete description of the changes needs to be written here.~

_The title of this PR actually does completely describe what has been changed, although not in granular detail._

~- Documentation needs to be added regarding the changes introduced.~

_As the intention of this PR is to introduce Unit Tests as they currently exist, the Unit Tests themselves have not been audited sufficiently to be able to write documentation that is consistent with what exists. Both these tasks should be accomplished in a separate PR(s)._

~It _is_ opened as a draft because it represents a significant number of file changes, and I want to maximize the window of time during which these changes can be inspected.~

This PR is now ready for review. Previous reservations are documented above, with reasons why they are no longer concerning factors.


WispyMouse (approved):
Unity doesn't complain about this change, and nothing unusual is included in the commits. Approved.

.github/workflows/unity-version-badge.yml
```diff
@@ -40,7 +40,7 @@ jobs:
           git config --global user.name 'github-actions'
           git config --global user.email 'github-actions@github.com'
           git add README.md
-          git add etc/PackageTemplate/README.md
```
andrew-hirata-playeveryware: I think this is supposed to be in a different PR?

andrew-hirata-playeveryware (changes requested):

.github/workflows/unity-version-badge.yml
paulhazen: Yep! This change is actually also in a PR that is opened [here](https://github.com/PlayEveryWare/eos_plugin_for_unity/pull/786). 

The reason I merged it into this PR is because _this_ PR triggers the workflow file that has been changed, and GitHub's user interface displays a "failed check" for the PR because it is broken in `development`, so I pre-emptively merged it into this PR to avoid the confusion that such a warning would generate.

paulhazen: As an example - _any_ PR opened into `development` after the merge that moved the `README.md` file into `com.playeveryware.eos` will have a failed check. It just so happens that this PR is the first one to trigger it.

andrew-hirata-playeveryware (approved):

PR #787, opened 2024-07-16
refactor(sessions): Session State Management Mirroring

WispyMouse:
- Users that are not the Owner of a Session cannot Start, Stop, or Modify a Session. The UI for those operations have been hidden in the sample.
- When a User joins a Session, the Owner sends them a message indicating the current Start/End status of the Session
- When the Owner Starts/Ends a Session, inform all users so they can mirror the status
- When the Owner Registers/UnRegisters a user, all users are sent the list of Registered users, which are then locally Registered/UnRegistered
- I capitalized "Session" and "Owner" more consistently

This has been tested with three accounts in a public Session.

### DIFF
--- a/Assets/Scripts/EOSSessionsManager.cs
+++ b/Assets/Scripts/EOSSessionsManager.cs
@@ -195,15 +195,21 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
         public bool UpdateInProgress = true;
         public OnlineSessionState SessionState = OnlineSessionState.NoSession;
 
+        /// <summary>
+        /// Cached value for <see cref="DoesLocalUserOwnSession"/>.
+        /// Answers if the local user owns this Session, which is information that should not change after determined.
+        /// </summary>
+        private bool? doesLocalUserOwnSession = null;
+
         //private Session InvalidSession;
 
         //-------------------------------------------------------------------------
-        public bool InitFromInfoOfSessionDetails(SessionDetails Session)
+        public bool InitFromInfoOfSessionDetails(SessionDetails session)
         {
             //SessionDetails
 
             SessionDetailsCopyInfoOptions copyOptions = new SessionDetailsCopyInfoOptions();
-            Result result = Session.CopyInfo(ref copyOptions, out SessionDetailsInfo? outSessionInfo);
+            Result result = session.CopyInfo(ref copyOptions, out SessionDetailsInfo? outSessionInfo);
 
             if (result != Result.Success)
             {
@@ -211,7 +217,7 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
             }
 
             //session.Release();  // Crashes EOS on JoinSession if session is released here
-            InitFromSessionInfo(Session, outSessionInfo);
+            InitFromSessionInfo(session, outSessionInfo);
             return true;
         }
 
@@ -290,11 +296,8 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
         {
             if (!string.IsNullOrEmpty(Name))
             {
-                if (ActiveSession != null)
-                {
-                    ActiveSession.Release();
-                    ActiveSession = null;
-                }
+                ActiveSession?.Release();
+                ActiveSession = null;
 
                 CopyActiveSessionHandleOptions copyOptions = new CopyActiveSessionHandleOptions();
                 copyOptions.SessionName = Name;
@@ -355,57 +358,62 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
         /// False if either it was determined that the user doesn't own this Session,
         /// or if the operation to query that information failed.
         /// </returns>
-        public bool IsLocalUserOwnerOfSession
+        public bool DoesLocalUserOwnSession()
         {
-            get
+            ProductUserId localUserId = EOSManager.Instance?.GetProductUserId();
+
+            // If we can't even get the local user's id, that state suggests they shouldn't be owning any Sessions
+            if (localUserId == null)
             {
-                ProductUserId localUserId = EOSManager.Instance?.GetProductUserId();
-
-                // If we can't even get the local user's id, that state suggests they shouldn't be owning any Sessions
-                if (localUserId == null)
-                {
-                    return false;
-                }
-
-                // If the ActiveSession isn't set, then this user isn't a part of the Session
-                // Can't possibly own it
-                if (ActiveSession == null)
-                {
-                    return false;
-                }
-
-                ActiveSessionCopyInfoOptions options = new ActiveSessionCopyInfoOptions() { };
-                Result copyResult = ActiveSession.CopyInfo(ref options, out ActiveSessionInfo? sessionInfo);
-
-                if (copyResult != Result.Success || !sessionInfo.HasValue || !sessionInfo.Value.SessionDetails.HasValue)
-                {
-                    return false;
-                }
-
-                return sessionInfo.Value.SessionDetails.Value.OwnerUserId.Equals(localUserId);
+                return false;
             }
+
+            // If the ActiveSession isn't set, then this user isn't a part of the Session
+            // Can't possibly own it
+            if (ActiveSession == null)
+            {
+                return false;
+            }
+
+            if (doesLocalUserOwnSession.HasValue)
+            {
+                return doesLocalUserOwnSession.Value;
+            }
+
+            ActiveSessionCopyInfoOptions options = new ActiveSessionCopyInfoOptions() { };
+            Result copyResult = ActiveSession.CopyInfo(ref options, out ActiveSessionInfo? sessionInfo);
+
+            if (copyResult != Result.Success || !sessionInfo.HasValue || !sessionInfo.Value.SessionDetails.HasValue)
+            {
+                return false;
+            }
+
+            doesLocalUserOwnSession = sessionInfo.Value.SessionDetails.Value.OwnerUserId.Equals(localUserId);
+            return doesLocalUserOwnSession.Value;
         }
 
         public bool TryGetRegisteredUsers(out List<ProductUserId> userIds)
         {
+            userIds = null;
+
             if (ActiveSession == null)
             {
-                userIds = null;
                 return false;
             }
 
-            ActiveSessionCopyInfoOptions copyInfoOptions = new ActiveSessionCopyInfoOptions() { };
-            ActiveSessionInfo? copiedInfo;
-            Result activeSessionInfoCopyResult = ActiveSession.CopyInfo(ref copyInfoOptions, out copiedInfo);
+            ActiveSessionCopyInfoOptions copyInfoOptions = new();
+            Result activeSessionInfoCopyResult = ActiveSession.CopyInfo(
+                ref copyInfoOptions, 
+                out ActiveSessionInfo? copiedInfo
+                );
 
             if (activeSessionInfoCopyResult != Result.Success)
             {
                 Debug.LogError($"{nameof(EOSSessionsManager)} ({nameof(TryGetRegisteredUsers)}): Failed to copy session information. Result code {activeSessionInfoCopyResult}.");
-                userIds = null;
                 return false;
             }
 
-            ActiveSessionGetRegisteredPlayerCountOptions countOptions = new ActiveSessionGetRegisteredPlayerCountOptions() { };
+            ActiveSessionGetRegisteredPlayerCountOptions countOptions = new();
             uint playerCount = ActiveSession.GetRegisteredPlayerCount(ref countOptions);
 
             // First iterate across the list, getting all registered users
@@ -413,7 +421,7 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
 
             for (uint playerIndex = 0; playerIndex < playerCount; playerIndex++)
             {
-                ActiveSessionGetRegisteredPlayerByIndexOptions indexOptions = new ActiveSessionGetRegisteredPlayerByIndexOptions() { PlayerIndex = playerIndex };
+                ActiveSessionGetRegisteredPlayerByIndexOptions indexOptions = new() { PlayerIndex = playerIndex };
                 ProductUserId userId = ActiveSession.GetRegisteredPlayerByIndex(ref indexOptions);
                 userIds.Add(userId);
             }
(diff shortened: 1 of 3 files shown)


Assets/Scripts/StandardSamples/UI/MatchMaking/UISessionEntry.cs
```diff
@@ -84,13 +102,21 @@ public void EnableButtonsBySessionState(bool updating, OnlineSessionState state)
...
+                    StartButton.interactable = !updating && (state == OnlineSessionState.Pending || state == OnlineSessionState.Ended);
+                    EndButton.interactable = !updating && state == OnlineSessionState.InProgress;
+                    ModifyButton.interactable = !updating;
+                }
+                else
+                {
+                    StartButton.interactable = false;
+                    EndButton.interactable = false;
+                    ModifyButton.interactable = false;
+                }
```
paulhazen: I would insert a block of comments first thing within each clause of the conditional explaining in plain language (doesn't have to be super specific - just general) what is happening in each clause and why.

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -198,29 +198,29 @@ public class Session
         //private Session InvalidSession;
 
         //-------------------------------------------------------------------------
-        public bool InitFromInfoOfSessionDetails(SessionDetails session)
+        public bool InitFromInfoOfSessionDetails(SessionDetails Session)
```
paulhazen: Parameters should start with lowercase letters.

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -290,6 +290,12 @@ public void InitActiveSession()
         {
             if (!string.IsNullOrEmpty(Name))
             {
+                if (ActiveSession != null)
+                {
+                    ActiveSession.Release();
+                    ActiveSession = null;
+                }
```
paulhazen: Using null propagation, you can change this to the (arguably) clearer:
```csharp
ActiveSession?.Release();
ActiveSession = null;
```

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -331,6 +345,81 @@ public override int GetHashCode()
...
+
+                ActiveSessionCopyInfoOptions options = new ActiveSessionCopyInfoOptions() { };
+                Result copyResult = ActiveSession.CopyInfo(ref options, out ActiveSessionInfo? sessionInfo);
+
+                if (copyResult != Result.Success || !sessionInfo.HasValue || !sessionInfo.Value.SessionDetails.HasValue)
+                {
+                    return false;
+                }
+
+                return sessionInfo.Value.SessionDetails.Value.OwnerUserId.Equals(localUserId);
```
paulhazen: This is a lot of work to do within a Property field member. The expectation is that Properties should have a "runtime" as close to a bare field member as possible. I think you should either refactor this property to cache the value returned so that it is only retrieved once, or change it from a Property to a static function with the signature:
```csharp
public static bool DoesLocalUserOwnSession(ActiveSession session)
{
    /* ... */
}
```

You could also make the above `static` function private, and call it from a public instance function or something.

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -331,6 +345,81 @@ public override int GetHashCode()
...
+            }
+        }
+
+        public bool TryGetRegisteredUsers(out List<ProductUserId> userIds)
+        {
+            if (ActiveSession == null)
+            {
+                userIds = null;
+                return false;
+            }
```
paulhazen: You can simplify several of the conditionals within this function if you start the function with:
```csharp
userIds = null;
```

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -331,6 +345,81 @@ public override int GetHashCode()
...
+
+            ActiveSessionCopyInfoOptions copyInfoOptions = new ActiveSessionCopyInfoOptions() { };
+            ActiveSessionInfo? copiedInfo;
+            Result activeSessionInfoCopyResult = ActiveSession.CopyInfo(ref copyInfoOptions, out copiedInfo);
```
paulhazen: It's preferable to inline the declaration of the out variable like this:

```csharp
Result activeSessionInfoCopyResult = ActiveSession.CopyInfo(
    ref copyInfoOptions, 
    out ActiveSessionInfo? copiedInfo
);
```

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -331,6 +345,81 @@ public override int GetHashCode()
...
+
+            if (activeSessionInfoCopyResult != Result.Success)
+            {
+                Debug.LogError($"{nameof(EOSSessionsManager)} ({nameof(TryGetRegisteredUsers)}): Failed to copy session information. Result code {activeSessionInfoCopyResult}.");
+                userIds = null;
+                return false;
+            }
+
+            ActiveSessionGetRegisteredPlayerCountOptions countOptions = new ActiveSessionGetRegisteredPlayerCountOptions() { };
```
paulhazen: I think it is acceptable - nay preferable to (particularly when the type name is long) use the simple `new()` means. In this particular case, you can also omit the curly brackets at the end. This may be a preference thing (the use of `new()`, not the curly brackets - either way it makes sense to remove those).

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -431,43 +520,82 @@ public class EOSSessionsManager : IEOSSubManager
...
+        /// so members of the Session should also remove themselves from the Session.
         /// </summary>
         private const string P2P_SESSION_OWNER_DESTROYED_SESSION_MESSAGE_ELEMENT = "DESTROY";
 
+        /// <summary>
+        /// Messages with this as the {2} parameter of <see cref="P2P_INFORM_SESSION_MESSAGE_FORMAT"/> indicate that the Owner of a Session has started the Session.
+        /// Members of the Session should start their Session locally.
+        /// When a user joins a Session, if the Session is <see cref="OnlineSessionState.Starting"/> or <see cref="OnlineSessionState.InProgress"/>,
+        /// the Owner should inform the joining user that the Session should be Started locally.
+        /// Otherwise when the Owner starts a Session, they should inform the users to start theirs locally.
```
paulhazen, line 580: This is quite a bit of information to include for a `const` definition. It seems like only line `568` is required for the comment. The rest of the content seems important, but seems like it might belong elsewhere.

Similar thoughts regarding the following two `const` definitions.

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -1595,36 +1723,36 @@ private void OnDestroySessionCompleteCallback(ref DestroySessionCallbackInfo dat
                 return;
             }
 
-            // Before removing the session from our local data, we need to inform the owner of the session that we've left the session, if we're not the owner
-            // TODO: Validate that this gets to the members/owners of the session in time, and that we haven't already deleted the local information needed to get session information
+            // Before removing the Session from our local data, we need to inform the Owner of the Session that we've left the Session, if we're not the Owner
+            // TODO: Validate that this gets to the members/owners of the Session in time, and that we haven't already deleted the local information needed to get Session information
```
paulhazen: "...and that the local information needed to get Session information has not already been deleted." is a version that removes the use of the word, "we."

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -1595,36 +1723,36 @@ private void OnDestroySessionCompleteCallback(ref DestroySessionCallbackInfo dat
...
+            // If we can't get the Owner, we can't determine who should be messaged, or if we are the Owner of this Session
             if (localCopyResult != Result.Success || !outActiveSessionInfo.HasValue || !outActiveSessionInfo.Value.SessionDetails.HasValue)
             {
-                Debug.LogError($"{nameof(EOSSessionsManager)} ({nameof(OnDestroySessionCompleteCallback)}): Failed to copy local information for session {sessionName}, so could not inform owner/members of destruction. Result code {localCopyResult}");
+                Debug.LogError($"{nameof(EOSSessionsManager)} ({nameof(OnDestroySessionCompleteCallback)}): Failed to copy local information for Session {sessionName}, so could not inform Owner/members of destruction. Result code {localCopyResult}");
                 return;
             }
 
-            if (outActiveSessionInfo.Value.SessionDetails.Value.OwnerUserId.Equals(EOSManager.Instance.GetProductUserId()))
+            if (localSession.IsLocalUserOwnerOfSession)
```
paulhazen: I can see why you wrote this property. I still think that it should either be cached or converted to a static function as I commented earlier, but it's clear why implementing it one way or another is a win. Nice job!

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -2090,6 +2220,35 @@ public void Register(string sessionName, ProductUserId userIdToRegister)
...
+        /// Other users should be informed by the Owner to Register users, so they have their information.
+        /// P2P messages can be used to inform the Owner and users when to run this method.
+        /// </summary>
+        /// <param name="sessionName">The name of the local Session to register the user to.</param>
+        /// <param name="userIdsToRegister">
+        /// The ProductUserId of the users to Register to the Session.
+        /// If this array is empty, the function exits without performing an operation.
+        /// Uses <paramref name="sessionName"/> for ClientData to identify the Session to register the user to.
+        /// </param>
+        public void Register(string sessionName, ProductUserId[] userIdsToRegister)
```
paulhazen: This is a great circumstance in which to utilize the `params` keyword. Opinions may differ on best practices, but if you're unfamiliar with the keyword this would be a good circumstance to become familiar with it (regardless of whether this is an instance in which to make use of it).

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -2108,9 +2267,9 @@ private void OnRegisterCompleteCallback(ref RegisterPlayersCallbackInfo data)
             // ClientData should contain the local sessionName
             if (data.ClientData is string localSessionName)
             {
-                // Refresh the owner's local UI, and also inform members
+                // Refresh the Owner's local UI, and also inform members
                 RefreshSession(localSessionName);
-                InformSessionMembers(localSessionName, P2P_REFRESH_SESSION_MESSAGE_ELEMENT);
+                InformUsersOfRegistrationStatusChange(localSessionName);
```
paulhazen, line 2277: This is a great change 👍

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -2134,6 +2296,35 @@ public void UnRegister(string sessionName, ProductUserId userIdToUnRegister)
...
+                return;
+            }
+
+            UnregisterPlayersOptions unregisterOptions = new UnregisterPlayersOptions();
+            unregisterOptions.SessionName = sessionName;
+            unregisterOptions.PlayersToUnregister = userIdsToUnRegister;
+
+            SessionsInterface sessionInterface = EOSManager.Instance.GetEOSPlatformInterface().GetSessionsInterface();
+            sessionInterface.UnregisterPlayers(ref unregisterOptions, sessionName, OnUnregisterCompleteCallback);
+        }
```
paulhazen: I would change the signature of this function to emphasize that it is the user id(s) who are being unregistered from the session. That may be the only kind of unregistration that takes place, but the clarity seems like a win.

paulhazen (changes requested):
A few very small changes. Please feel free to mark the conversations as resolved when you have resolved them to your satisfaction - be it through a change or a determination that a change is not warranted. I'll promptly rereview with an approval.

Assets/Scripts/StandardSamples/UI/MatchMaking/UISessionEntry.cs
WispyMouse: Sure, that seems good. This is a seemingly arbitrary list of things doing things.

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: Ahhhh, missed one. I had done a full replace for all "session" to become "Session" before realizing that would affect parameter names. Tried to hit as many of these as I could, will fix!

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: This is a great use of the null safe operation, good suggestion.

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: I'm comfortable with the idea that this is too big for a property. It's not *expensive* work, but it is a lot of logic.

I didn't want to cache the value, in the case that this is accessed before the Session has had all of its information saturated properly... but upon looking at this, I realized that I don't have to actually store the cached value until the operation gets to a comfortable point. If ActiveSession etc is null, just return false without hitting the cache.

WispyMouse: This function is part of the Session object already, so I don't need to pass in the parameter.

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: Oh, huh. I never considered doing that for out parameter things. This would remove some redundancy.

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: I can see where you're coming from, and will make that change. In this case since the variables aren't complex, I don't mind it all being on one line, but I suppose for out parameters that can be different?

WispyMouse: Ah I see now, your change incorporates designating the variable too.

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: Honestly most of the time we have these { }s after these structs, it's to see if anything should go in there. The answer for this one was no, there's nothing inside.

Mint is not fully on board with the new() syntax, but at least in this case the name of the object is in the variable declaration, so it's fine here.

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: I think I'm comfortable with the idea that Register and such will document this for me. I'll snip all but the first two lines, which I think are necessary, and replace it with a `<see cref="InformUsersOfRegistrationStatusChange(string)"/>`

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: It feels like this is best called as an array. `params` seems more fit for something reading in command line arguments or something.

WispyMouse: Not sure how to use this limited set of emoji to suggest "I contest this change idea"... But ah, I don't think this one should be params-y, or at least there should also be this array option. Unless params covers the array option, which I think maybe it does..?

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: Sure! The Sessions Interface calls it `UnregisterPlayers`, so let's add "Players" and "Player" to the end of these function names.

andrew-hirata-playeveryware: Also it should be "Unregister" not "UnRegister".

Assets/Scripts/EOSSessionsManager.cs
andrew-hirata-playeveryware: I also feel it is more appropriate for this to be a method or static method instead of a property.

andrew-hirata-playeveryware (approved):

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: That's an easy change, sure. Maintained it from its previous name.

paulhazen (approved):

PR #790, opened 2024-07-18
refactor(sessions): Adding Friends UI Tab to Sessions

WispyMouse:
Added the Friends Tab to the Sessions scene. You can invite friends from that menu if your friend is online, and you're in a Presence Enabled Session. The invitation button is disabled if the Session you're in isn't applicable to invitations, such as rooms that are full or are marked as not allowing joining in progress.

Minor touchup of `UIFriendInteractionSource` to make it manage the dirty flag.

### DIFF
--- a/Assets/Scripts/EOSFriendsManager.cs
+++ b/Assets/Scripts/EOSFriendsManager.cs
@@ -436,20 +436,23 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
                 TargetUserId = data.TargetUserId
             };
 
-            Result joinInfoResult = PresenceHandle.GetJoinInfo(ref joinInfoOptions, out Utf8String joinInfo);
-            if (joinInfoResult != Result.Success)
-            {
-                joinInfo = null;
-            }
-
+            // If the user has a Presence JoinInfo, set that in this incoming Presence.
+            // NotFound just implies the user does not have an active joinable activity, it is not an error.
             PresenceInfo presenceInfo = new PresenceInfo()
             {
                 Application = presence?.ProductId,
                 Platform = presence?.Platform,
                 Status = (Status)(presence?.Status),
                 RichText = presence?.RichText,
-                JoinInfo = joinInfo
+                JoinInfo = null // Default to null until the value is determined
             };
+
+            Result joinInfoResult = PresenceHandle.GetJoinInfo(ref joinInfoOptions, out Utf8String joinInfo);
+
+            if (joinInfoResult == Result.Success)
+            {
+                presenceInfo.JoinInfo = joinInfo;
+            }
 
             if(CachedFriends.TryGetValue(data.TargetUserId, out FriendData friendData))
             {
--- a/Assets/Scripts/EOSFriendsManager.cs
+++ b/Assets/Scripts/EOSFriendsManager.cs
@@ -437,7 +437,6 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
             };
 
             // If the user has a Presence JoinInfo, set that in this incoming Presence.
-            // NotFound just implies the user does not have an active joinable activity, it is not an error.
             PresenceInfo presenceInfo = new PresenceInfo()
             {
                 Application = presence?.ProductId,
@@ -449,6 +448,8 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
 
             Result joinInfoResult = PresenceHandle.GetJoinInfo(ref joinInfoOptions, out Utf8String joinInfo);
 
+            // If the result of getting join info is "Result.NotFound", that implies the user does not have an active joinable activity.
+            // Only on a Success can the JoinInfo be used, but otherwise the JoinInfo is to be ignored.
             if (joinInfoResult == Result.Success)
             {
                 presenceInfo.JoinInfo = joinInfo;
--- a/Assets/Scripts/EOSSessionsManager.cs
+++ b/Assets/Scripts/EOSSessionsManager.cs
@@ -37,6 +37,7 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
     using Epic.OnlineServices.P2P;
     using System.Text.RegularExpressions;
     using System.Text;
+    using UnityEngine.Events;
 
     /// <summary>
     /// Class represents a Session search and search results
@@ -488,6 +489,11 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
         Queue<Action> UIOnJoinSession;
         Queue<Action> UIOnLeaveSession;
         Queue<Action> UIOnSessionSearchCompleted;
+
+        /// <summary>
+        /// Event to inform UI that it should refresh displayed Friends and Joining/Inviting buttons.
+        /// </summary>
+        public UnityEvent UIOnPresenceAffectingChange = new UnityEvent();
 
         public const ulong INVALID_NOTIFICATIONID = 0;
 
@@ -1194,6 +1200,13 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
         /// <returns>Returns true if there is a local Session with Presence enabled.</returns>
         public bool HasPresenceSession()
         {
+            // Can't have a presence session if we're not even logged in
+            if (!IsUserLoggedIn)
+            {
+                KnownPresenceSessionId = string.Empty;
+                return false;
+            }
+
             if (KnownPresenceSessionId.Length > 0)
             {
                 if (CurrentSessions.ContainsKey(KnownPresenceSessionId)) // TODO: validate
@@ -1716,6 +1729,7 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
             existingLocalSession.InitFromSessionInfo(sessionDetails, sessionInfo);
 
             UIOnSessionRefresh?.Invoke(existingLocalSession, sessionDetails);
+            UIOnPresenceAffectingChange?.Invoke();
         }
 
         #endregion
@@ -1812,6 +1826,8 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
                     CurrentSessions.Remove(sessionName);
                 }
             }
+
+            UIOnPresenceAffectingChange?.Invoke();
         }
 
         /// <summary>
@@ -1937,6 +1953,7 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
                         CurrentSessions[session.Name] = session;
                     }
 
+                    UIOnPresenceAffectingChange?.Invoke();
                     InformSessionOwnerWithMessage(session.Name, P2P_JOINING_SESSION_MESSAGE_ELEMENT);
                 }
                 callback?.Invoke(result);
@@ -2638,6 +2655,7 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
                 }
 
                 InformSessionMembers(sessionName, P2P_REFRESH_SESSION_MESSAGE_ELEMENT);
+                UIOnPresenceAffectingChange?.Invoke();
             }
         }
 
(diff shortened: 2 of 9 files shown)


andrew-hirata-playeveryware (approved):

Assets/Scripts/StandardSamples/UI/Friends/UIFriendInteractionSource.cs
```diff
@@ -37,6 +37,13 @@ public enum FriendInteractionState
             Enabled
         }
 
+        /// <summary>
+        /// Indicates that this object needs to refresh its UI.
+        /// When this is true, the next time <see cref="UIFriendsMenu.Update"/> runs,
+        /// <see cref="UIFriendsMenu.RefreshFriendUI"/> will be called.
+        /// </summary>
+        protected bool dirtyFlag { get; private set; } = true;
```
paulhazen: This should be `DirtyFlag`, since it is a property. It also would probably be more appropriate to call it `IsDirty`. 

IMO, when naming a bool variable or property, you should be able to use it in a simulacrum of English grammar:

The following is meaningful:
```
Question: IsDirty? 
Answer: True
```
Whereas this is confusing:
```
Question: DirtyFlag?
Answer: True
```

Inferring that the second example _means the same thing_ as the first may be an inference that is more difficult to people for make.

Assets/Scripts/StandardSamples/UI/MatchMaking/UISessionsMatchmakingMenu.cs
```diff
@@ -35,8 +35,37 @@
...
+            /// </summary>
+            ValidSessionToInviteTo
+        }
+
+        /// <summary>
+        /// Cached state that indicates if the local user can send invitations for a Session.
+        /// Processed in <see cref="ProcessInformationBeforeFriendsRefresh"/>,
+        /// and utilized in <see cref="GetFriendInteractionState(FriendData)"/>.
+        /// </summary>
+        protected OwnSessionInvitationAbilityState ownInvitationState { get; set; } = OwnSessionInvitationAbilityState.NoSessionToInviteTo;
```
paulhazen: All properties should start with a capital letter.

Assets/Scripts/StandardSamples/UI/MatchMaking/UISessionsMatchmakingMenu.cs
```diff
@@ -403,5 +434,112 @@ public void OnSessionRefresh(Session session, SessionDetails details)
...
+
+        #region UIFriendInteractionSource Implementations
+
+        /* 
+         * REGION NOTES:
+         * 
+         * The UIFriendsMenu popout uses the base class UIFriendInteractionSource to facilitate interactions.
+         * 
+         * NOTE: This current implementation is not complete.
+         * There is functionality suggested by the existance of this window that need to be added.
```
paulhazen: *needs

Assets/Scripts/StandardSamples/UI/MatchMaking/UISessionsMatchmakingMenu.cs
```diff
@@ -403,5 +434,112 @@ public void OnSessionRefresh(Session session, SessionDetails details)
...
+            if (ownInvitationState == OwnSessionInvitationAbilityState.InvalidSessionToInviteTo)
+            {
+                return FriendInteractionState.Disabled;
+            }
+
+            // The only thing remaining is yes, this user can be interacted with
+            return FriendInteractionState.Enabled;
+        }
+
+        public override void ProcessInformationBeforeFriendsRefresh()
```
paulhazen: I have a philosophical problem with this function name.

Functions should be named in a descriptive manner such that they clearly articulate the actions that are taken. Nothing more, and nothing less. 

They should _not_ be named in such a way that makes implications regarding _when_ the function should be called. Such articulations should be made through comments and perhaps supplemented with other documentation.

Assets/Scripts/EOSFriendsManager.cs
```diff
@@ -436,21 +436,24 @@ private void OnQueryPresenceCompleted(ref QueryPresenceCallbackInfo data)
...
             };
 
-            Result joinInfoResult = PresenceHandle.GetJoinInfo(ref joinInfoOptions, out Utf8String joinInfo);
-            if (joinInfoResult != Result.Success)
-            {
-                joinInfo = null;
-            }
-
+            // If the user has a Presence JoinInfo, set that in this incoming Presence.
+            // NotFound just implies the user does not have an active joinable activity, it is not an error.
```
paulhazen: It's unclear to me in this context what `NotFound` is referring to.

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -489,6 +490,11 @@ public class EOSSessionsManager : IEOSSubManager
         Queue<Action> UIOnLeaveSession;
         Queue<Action> UIOnSessionSearchCompleted;
 
+        /// <summary>
+        /// Event to inform UI that it should refresh displayed Friends and Joining/Inviting buttons.
+        /// </summary>
+        public UnityEvent UIOnPresenceAffectingChange = new UnityEvent();
```
paulhazen: I would name this simply `OnPresenceChange`. It is _used by_ the UI logic to make a change to the UI in _response to_ the presence state changing. It's possible that users or other areas of the code may want to subscribe to the event, and in that context this would be a confusion name.

Assets/Scripts/StandardSamples/UI/Friends/UIFriendInteractionSource.cs
```diff
@@ -52,13 +59,29 @@ public virtual void OnFriendInteractButtonClicked(FriendData friendData)
...
+        }
+
+        public virtual void ResetFriendsUIDirtyFlag()
         {
-            return false;
+            dirtyFlag = false;
         }
 
-        public virtual void ResetDirtyFlag()
+        public virtual void MarkFriendsUIDirty()
```
paulhazen: I believe that this introduces confusion by using the verb "Mark," vs in other instances where this value is referred to a "Flag," which in these contexts are synonymous, but using them interchangeably is (I believe) an inconsistency that can be easily avoided and lend greater clarity to what is happening.

Assets/Scripts/StandardSamples/UI/Friends/UIFriendInteractionSource.cs
```diff
@@ -52,13 +59,29 @@ public virtual void OnFriendInteractButtonClicked(FriendData friendData)
...
         {
+            dirtyFlag = true;
+        }
+
+        /// <summary>
+        /// This function is called before refreshing the <see cref="UIFriendsMenu"/>.
+        /// This is an opportunity to refresh any caches or do processing work that should be done once,
+        /// instead of processed for each call of <see cref="GetFriendInteractionState(FriendData)"/>.
+        /// </summary>
+        public virtual void ProcessInformationBeforeFriendsRefresh()
```
paulhazen: Is the intent of this virtual function being defined with an empty implementation that the _default_ implementation should be empty, and overridden by classes that derive from `UIFriendInteractionSource`?

If so, do any of the derivative classes require an empty implementation?

paulhazen (changes requested):
I think these are good changes, but I think the details of the implementation need to be changed - mostly regarding naming convention issues. See comments in the review for more details.

Assets/Scripts/StandardSamples/UI/Friends/UIFriendInteractionSource.cs
WispyMouse: That totally makes sense, and I feel like this was a great example for demonstrating that. I'll rename it to IsDirty.

Sounds like you also suggest that protected/properties should start capitalized?

WispyMouse: Just saw another comment saying yes.

Assets/Scripts/StandardSamples/UI/MatchMaking/UISessionsMatchmakingMenu.cs
WispyMouse: Yup, that's a valid grammar fixup. Changed.

Assets/Scripts/StandardSamples/UI/MatchMaking/UISessionsMatchmakingMenu.cs
WispyMouse: Hmmmm... Going to have a Slack conversation with you on this one!

Assets/Scripts/EOSFriendsManager.cs
WispyMouse: `Result.NotFound`, a possible result of the method coming after that comment. I think you expressing confusion makes it clear that this comment should be elaborated a bit, and moved to where it actually belongs.

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: Hrmmrmmrmm.... This is another hard one. I can accept that, I guess.

What's the naming difference between "UnityEvent you can subscribe to (possibly for UI)" and "This is the function that runs On something happening (and is not an event to subscribe to) such as `OnRefreshSessionFindSessionsCompleteCallback`? I guess the Callback at the end?

Assets/Scripts/StandardSamples/UI/Friends/UIFriendInteractionSource.cs
WispyMouse: What do you think this could be called instead? "Set"FriendsUIDirty?

Assets/Scripts/StandardSamples/UI/Friends/UIFriendInteractionSource.cs
WispyMouse: Yes. This is a timing window for derivative classes to do optional processing work. There's no work that can be done in the parent class. I don't think it needs to be `abstract`, because implementors can also choose to not implement anything.

Currently, only the `UISessionsMatchmakingMenu` overrides this function, because it has work that needs to be done at this time. (before starting to process friends UI states)

Assets/Scripts/StandardSamples/UI/Friends/UIFriendInteractionSource.cs
paulhazen: I've always been a fan of doing this:
```csharp
public virtual void SetDirtyFlag(bool isDirty = true)
```

Leaving the `isDirty` parameter to default maintains the correct grammar of "SetDirtyFlag", while also making clear that `SetDirtyFlag(false)` means to _unset_ the dirty flag.

Assets/Scripts/EOSSessionsManager.cs
paulhazen: The difference (in my mind) should be in the _tense_ of the verb in concert with the callback. So `OnPresenceChange` is _subscribable_, and a naming convention that consumers might use for their function that is triggered on this event would be `OnPresenceChangedCallback`. One indicates that an action is happening, the other implies that it's happening to _resolve_ the action of the presence changing. Curious what @andrew-hirata-playeveryware would have to say about this - he's a fan of grammar and these kinds of discussions.

Also to be super-duper clear - this is a tiny point of discussion. Overall these changes are great. I just noticed the opportunity for a conversation. Having to think more explicitly around it reveals to me that there is a degree to which I can't articulate exactly my reasoning - which is an excellent sign that I should understand that instinct better, and change my mind if I can't articulate it.

WispyMouse: Ah, I didn't mean to imply that "this is frustrating to implement". Mostly just that I hadn't had protocols for the naming of these kinds of things before, and trying to reconcile it with past information. You're good, and talking about small details is what I was looking for in the PR.

Imma change this to OnPresenceChange, and make sure callback style event things end in -Callback.

Assets/Scripts/StandardSamples/UI/Friends/UIFriendInteractionSource.cs
WispyMouse: That sounds simple enough, I'll do that!

WispyMouse: There's a message on Slack in relation to this. The subscribed event method is added in this changelist, would like your feedback on naming it.

paulhazen (approved):

PR #792, opened 2024-07-18
fix(android,fileIO) : remove file check that prevent file reading on android

### LINKED ISSUE #788: Android: `log_level_config.json` not found.
I'm seeing this error in Android build after updating to latest from `3.1.0`:

```csharp
Error Unity FileNotFoundException: Config file "jar:file:///data/app/~~GlkpSHNC1KZ_kWCHd5yOKw==/*package name*-6PIJZtVpcHpmNAccpkJn_Q==/base.apk!/assets/EOS/log_level_config.json" does not exist.
```

I see the file in `StreamingAssets/EOS`, and I'm not seeing the issue in editor and PC, PS5 builds. Not sure about other platforms.

Using Unity `2021.3.35`. Is there any special setup needed for the new config file on Android?

arthur740212:
`File.Exist` will always return false for android files, thus this is modified.

The UNITY_EDITOR `Write` can be isolated because `Write` is only meant to happen on editor play mode.

The file check for other platforms could be omitted because they are handled in `ReadAllText` function of their respective platform.

### DIFF
--- a/Assets/Plugins/Source/Core/Config.cs
+++ b/Assets/Plugins/Source/Core/Config.cs
@@ -280,22 +280,16 @@ namespace PlayEveryWare.EpicOnlineServices
         /// </summary>
         protected virtual void Read()
         {
-            bool configFileExists = File.Exists(FilePath);
-
-            if (configFileExists)
+            
+#if UNITY_EDITOR // This should only happen in Editor Runtime (Play Mode)
+            if (!File.Exists(FilePath)) 
             {
-                _lastReadJsonString = FileUtility.ReadAllText(FilePath);
-                JsonUtility.FromJsonOverwrite(_lastReadJsonString, this);
-            }
-            else
-            {
-#if UNITY_EDITOR
                 Write();
-#else
-                throw new FileNotFoundException(
-                    $"Config file \"{FilePath}\" does not exist.");
-#endif
             }
+#endif
+
+            _lastReadJsonString = FileUtility.ReadAllText(FilePath);
+            JsonUtility.FromJsonOverwrite(_lastReadJsonString, this);
         }
 
         // Functions declared below should only ever be utilized in the editor.
--- a/Assets/Plugins/Source/Core/Config.cs
+++ b/Assets/Plugins/Source/Core/Config.cs
@@ -337,9 +337,6 @@ namespace PlayEveryWare.EpicOnlineServices
             bool prettyPrint = true, 
             bool updateAssetDatabase = true)
         {
-            FileInfo configFile = new(FilePath);
-            configFile.Directory?.Create();
-
             var json = JsonUtility.ToJson(this, prettyPrint);
 
             // If the json hasn't changed since it was last read, then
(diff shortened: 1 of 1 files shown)


Assets/Plugins/Source/Core/Config.cs
```diff
@@ -280,22 +280,16 @@ protected virtual async Task ReadAsync()
         /// </summary>
         protected virtual void Read()
         {
-            bool configFileExists = File.Exists(FilePath);
-
-            if (configFileExists)
+            
+#if UNITY_EDITOR // This should only happen in Editor Runtime (Play Mode)
+            if (!File.Exists(FilePath))
```
paulhazen: During runtime when not in the editor - how do Config files get read?

I think most of them utilize `ReadAsync`, but I think that's just a coincidence.

paulhazen (changes requested):
My main concern is that I was fairly certain that the `Read()` function on the `Config` class _is_ utilized within contexts where the editor is not running.

Regardless of that factor - changes made to `Read()` should also be replicated in `ReadAsync()`.

Is it possible that a better solution would be to never use `File.Exists`, but instead implement a platform-aware method within `FileUtility` that can determine if a file exists on any platform (I believe only Android would need to have a unique implementation).

Assets/Plugins/Source/Core/Config.cs
arthur740212: During runtime on any platform (including editor), it does use this `Read()`
The comment here is talking about the `Write`.
I'll add a couple words to make it clear.

ReadAsync is used in editor when not in runtime (i.e. in EditorsWindows)

arthur740212:
> My main concern is that I was fairly certain that the `Read()` function on the `Config` class _is_ utilized within contexts where the editor is not running.
> 
To be precise.
`Read()` is for Player mode (including player mode on editor).
`ReadAsync()` is for editor mode (in other words, in editor but not in player mode).

> Regardless of that factor - changes made to `Read()` should also be replicated in `ReadAsync()`.
> 
`ReadAsync()` does not need these changes because it only runs in the editor, which means File.Exist will have expected behavior.

> Is it possible that a better solution would be to never use `File.Exists`, but instead implement a platform-aware method within `FileUtility` that can determine if a file exists on any platform (I believe only Android would need to have a unique implementation).
>
We can, but in this case wouldn't need to.

arthur740212:
Because `ReadAsync` is only called in Editor Windows (kudos for handling those). 
I am thinking if there is a way to separate the two and prevent cross usage. The easiest way is simply doing an #if UNITY_EDITOR around `ReadAsync`

andrew-hirata-playeveryware (approved):

Assets/Plugins/Source/Core/Config.cs
```diff
@@ -280,22 +280,16 @@ protected virtual async Task ReadAsync()
...
-            bool configFileExists = File.Exists(FilePath);
-
-            if (configFileExists)
-            {
-                _lastReadJsonString = FileUtility.ReadAllText(FilePath);
-                JsonUtility.FromJsonOverwrite(_lastReadJsonString, this);
-            }
-            else
+            
+#if UNITY_EDITOR // This should only happen in Editor Runtime (Play Mode)
```
andrew-hirata-playeveryware: We should leave a comment on _why_ this should only happen in editor mode.

paulhazen (approved):

PR #793, opened 2024-07-18
Fix log level native

arthur740212:
The log config in native used to be read with camel case, and it causes a crash.
This fix reads with the updated Pascal case and fixes the crash

The native libs for windows are updated alongside


lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/dllmain.cpp
```diff
@@ -975,11 +975,11 @@ static LogLevelConfig log_config_from_json_value(json_value_s* config_json)
                 struct json_object_element_s* pairs_iter = pairs_json_object->start;
                 while (pairs_iter != nullptr)
                 {
-                    if (!strcmp("category", pairs_iter->name->string))
+                    if (!strcmp("Category", pairs_iter->name->string))
                     {
                         log_config.category.push_back(json_value_as_string(pairs_iter->value)->string);
                     }
-                    else if (!strcmp("level", pairs_iter->name->string))
+                    else if (!strcmp("Level", pairs_iter->name->string))
```
andrew-hirata-playeveryware, line 982: Hmm it just occurred to me that we don't have a naming convention for our config file's fields. My personal preferences is that fields are either camelCase or snake_case.

paulhazen: I agree. Specifically, I think that JSON standards dictate using `snake_case`, but that is not consistent with our current implementation of using `camelCase`.

WispyMouse (approved):

paulhazen (approved):

PR #795, opened 2024-07-19
`EOSConfig` Simplification

paulhazen:
This PR simplifies a variety of mechanisms surrounding the `EOSConfig` class and how it is utilized.

This work is done as part of the continuing effort to ultimately introduce config class data validation.

The two significant areas of change are with respect to the parsing of a variety of different string values into Epic-defined enum values, and the assigning of override values for the ThreadAffinity on a platform-to-platform basis.

## Enum Parsing

Typically, most applications can simply use `Enum.TryParse` to perform parsing from string values into enum values, but in-order to provide backwards compatibility with what the values _used to be named_ we need a lookup of all the string values that should be possible to parse into enums.

Prior to this PR this was accomplished in some very large and difficult to maintain conditionals. With these changes, those lookup maps are sequestered to their own extension and utility classes, allowing the `EOSConfig` to simply be a place where data is stored, allowing it to be ignorant of the specific details regarding how to parse a particular string into a particular enum value.

## OverrideThreadAffinity Configuration

This PR appropriately makes private all of the accessor methods to the field members that pertain to thread affinity, and instead introduces a single function that accepts a reference to a thread affinity struct, and sets the values accordingly.


WispyMouse:
I have a compiler error.

> Assets\Plugins\Source\Core\PlatformSpecifics.cs(118,39): error CS1061: 'EOSConfig' does not contain a definition for 'SetOverrideThreadAffinity' and no accessible extension method 'SetOverrideThreadAffinity' accepting a first argument of type 'EOSConfig' could be found (are you missing a using directive or an assembly reference?)

CTRL+Fing through the code, I don't see a function named SetOverrideThreadAffinity.

WispyMouse (approved):
I have a compilation error, posted it. If you can fix that, then this gets approved by me.

paulhazen:
> I have a compiler error.
> 
> > Assets\Plugins\Source\Core\PlatformSpecifics.cs(118,39): error CS1061: 'EOSConfig' does not contain a definition for 'SetOverrideThreadAffinity' and no accessible extension method 'SetOverrideThreadAffinity' accepting a first argument of type 'EOSConfig' could be found (are you missing a using directive or an assembly reference?)
> 
> CTRL+Fing through the code, I don't see a function named SetOverrideThreadAffinity.

Good catch - I noticed this too and have since pushed a change that resolves it.

andrew-hirata-playeveryware:
This change modifies the public API of the EOSManager so I believe we need to have a prospectus meeting to talk about at least the route to deprecation if we decide to take these changes.

andrew-hirata-playeveryware (changes requested):
This change modifies the public API of the EOSManager so I believe we need to have a prospectus meeting to talk about at least the route to deprecation if we decide to take these changes.

paulhazen:
> This change modifies the public API of the EOSManager so I believe we need to have a prospectus meeting to talk about at least the route to deprecation if we decide to take these changes.

I went ahead and re-introduced the unused functions that altered the API, changing their implementation, but maintaining their functionality and signatures. There should not be any changes to public (or for that matter protected) function signatures.

Assets/Plugins/Source/Core/Utility/EnumUtility.cs
```diff
@@ -0,0 +1,180 @@
...
+            {
+                return true;
+            }
+
+            // Try to parse the string using the Enum.TryParse method, fall back
+            // on the custom mapping dictionary declared above if that fails.
+            if (!Enum.TryParse(enumValueString, out result))
+            {
+                Debug.LogError($"\"{enumValueString}\" was not recognized as a valid {nameof(TEnum)} value, and parsing failed.");
+                return false;
```
andrew-hirata-playeveryware: Shouldn't it check the custom mappings first before returns false on parsing? 
i.e. shouldn't it be the case that if the enum doesn't directly parse, it should try to look it up in the provided custom mappings before it returns false?

Assets/Plugins/Source/Core/Utility/Extensions/AuthScopeFlagsExtensions.cs
```diff
@@ -0,0 +1,89 @@
...
+            {"EOS_AS_FriendsList",       AuthScopeFlags.FriendsList},
+            {"EOS_AS_Presence",          AuthScopeFlags.Presence},
+            {"EOS_AS_FriendsManagement", AuthScopeFlags.FriendsManagement},
+            {"EOS_AS_Email",             AuthScopeFlags.Email},
+            {"EOS_AS_Country",           AuthScopeFlags.Country}
+        };
+
+        public static string GetDescription(this AuthScopeFlags flags)
+        {
+            return flags switch
```
andrew-hirata-playeveryware, line 53: I didn't know about this syntax in C# : it's nifty.

andrew-hirata-playeveryware (changes requested):

Assets/Plugins/Source/Core/Utility/EnumUtility.cs
paulhazen: Ah, yes. Good catch. I fixed the order of operations.

andrew-hirata-playeveryware (approved):

PR #797, opened 2024-07-24
refactor(sessions): Acknowledge Event

WispyMouse:
Whenever a call returns a UiEventId, it should be Acknowledged. Previously the AcknowledgeEventId function was hard wired to only deal with one event, now it can take any event.

Several member variables that were used to store temporary state have been reduced and removed. I disliked the way that important state information was juggled. There is still some, Invites mostly, that are still managed in a disagreeable way.

Removed the notification for "game joined". This notification was pretty misleading; it appears to be an event that fires when you invite a user over the social overlay, and they accept over the social overlay. That notification wasn't meaningfully being handled or used, so we cut it for now.

Some Action delegate pass-ins have been adjusted a bit. For example see JoinSession, which now creates a delegate that passes in additional information without needing to be as tied to the "ClientData" field. I'd like feedback on what to call what is currently `runOnSearchResultReceived`.

### DIFF
--- a/Assets/Scripts/EOSSessionsManager.cs
+++ b/Assets/Scripts/EOSSessionsManager.cs
@@ -90,6 +90,14 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
             return SearchResults;
         }
 
+        /// <summary>
+        /// Looks through the Sessions stored in <see cref="SearchResults"/> for a Session with a matching id.
+        /// If found, outputs the <see cref="SessionDetails"/> handle object.
+        /// This can only find Sessions that are part of this search's results, and only if <see cref="OnSearchResultReceived(Dictionary{Session, SessionDetails})"/> has stored those values.
+        /// </summary>
+        /// <param name="sessionId">The id of the Session to find.</param>
+        /// <param name="sessionHandle">Output parameter for the SessionDetails handle.</param>
+        /// <returns>True if a Session was found with the Id. False otherwise.</returns>
         public bool TryGetSessionHandleById(string sessionId, out SessionDetails sessionHandle)
         {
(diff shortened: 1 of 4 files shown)


Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -47,6 +47,14 @@ public class SessionSearch
         public Epic.OnlineServices.Sessions.SessionSearch SearchHandle;
         private Dictionary<Session, SessionDetails> SearchResults = new Dictionary<Session, SessionDetails>();
 
+        /// <summary>
+        /// Action to run when search results are successfully retrieved.
+        /// Set in <see cref="SetNewSearch(Epic.OnlineServices.Sessions.SessionSearch, Action{SessionSearch})"/>,
+        /// unset in <see cref="Release"/>, when a new search is set without the argument,
+        /// or after <see cref="OnSearchResultReceived(Dictionary{Session, SessionDetails})"/> is called.
+        /// </summary>
+        private Action<SessionSearch> runOnSearchResultReceived { get; set; }
```
paulhazen: I'm of the opinion that all properties should be capitalized, regardless of access modifiers used. For field _members_ I would say private ones should start with an underscore and be camel case. In summary: I believe that this should be capitalized.

paulhazen (approved):
Approving because the change I'm requesting is very minimal, and I trust you will resolve it before merging. Feel free to mark the conversation as resolved once the topic is put to rest to your satisfaction.

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: Agreeable! I'll go set that up now.

PR #798, opened 2024-07-24
Runtime Config

paulhazen:
This PR introduces the new data structure proposed for containing the configuration values used by the plugin at runtime. It is distinct from the other task that has been discussed - mainly identifying which of these features should be overridable on a platform-by-platform basis. 

It seemed prudent to me to start by identifying what values are configurable in total and separate from that the concept of there being different values for each platform.

In addition to defining these set of values in this PR, their data types are appropriately set. In anticipation of providing functionality for using a single enum value for the field members which should be enum types (moreover, in anticipation of needing to provide a user interface to edit those single enum values) corresponding functionality was added to the `EnumUtility` that allows for the result of bitwise operations on Enums to be decomposed into the discrete enum values from which they were made.

Because this is a work in progress, the `RuntimeConfig` file is excluded from compilation by default.


Assets/Plugins/Source/Core/RuntimeConfig.cs
```diff
@@ -0,0 +1,258 @@
...
+    using System;
+
+#if EOS_RUNTIME_NEW_CONFIG_SYSTEM
+    /// <summary>
+    /// Contains the information necessary to configure the EOS SDK to connect
+    /// properly to Epic Online Services. Is ignorant of platform, because it
+    /// _only_ represents the _current_ runtime values to use. The values of the
+    /// field members defined within will need to be determined at runtime.
+    ///
+    /// _Project Settings_ is the area where there will be options to define
```
WispyMouse: not pr crit: Your use of markdown here makes me think there's a place that this can be read in markdown format. Is there?

Assets/Plugins/Source/Core/RuntimeConfig.cs
```diff
@@ -0,0 +1,258 @@
...
+    /// during development. Those settings are distinct from these set of
+    /// values.
+    ///
+    /// Most of the values for these field members come from the
+    /// [Development Portal](https://dev.epicgames.com/portal/)
+    /// </summary>
+    public readonly struct RuntimeConfig
+    {
+
+        /*
```
WispyMouse: To match the style of the recent Sessions rework comment regions, should this comment be inside the region?

Assets/Plugins/Source/Core/RuntimeConfig.cs
```diff
@@ -0,0 +1,258 @@
...
+         * The following region contains values that are required in order to
+         * use the EOS SDK.
+         */
+        #region EOS SDK Configuration Values
+
+        /// <summary>
+        /// Product Name defined in the
```
WispyMouse: Seems like an incomplete thought

Assets/Plugins/Source/Core/RuntimeConfig.cs
```diff
@@ -0,0 +1,258 @@
...
+
+        /// <summary>
+        /// Product Id
+        /// </summary>
+        public readonly Guid ProductId;
+
+        /// <summary>
+        /// Sandbox Id
+        /// </summary>
+        public readonly Guid SandboxId;
```
WispyMouse: Are sandbox ids always guids? I looked at the dev portal and saw that our sandbox id for dev is `p-pxx285jnqt8p3abnsekbcvhjwcgtk7` which is definitely not a guid. But then I looked at the sandbox for live `30f622bbe1f447889a26eb696d9fbb2a` and that's a guid.

Deployments on our product look like a GUID. The Product ID looks like a GUID.

Assets/Plugins/Source/Core/RuntimeConfig.cs
```diff
@@ -0,0 +1,258 @@
...
+                Guid.Parse(config.productID),
+                config.productName,
+                Version.Parse(config.productVersion),
+                Guid.Parse(config.sandboxID),
+                Guid.Parse(config.deploymentID),
+                new ClientCredentials() { ClientId = config.clientID, ClientSecret = config.clientSecret },
+                config.encryptionKey,
+                config.GetPlatformFlags(),
+                config.GetAuthScopeFlags(),
+                IntegratedPlatformManagementFlags.Disabled,
```
WispyMouse: Is there any way to pull the IntegratedPlatformManagementFlags from the incoming EOSConfig? It probably shouldn't always be disabled.

WispyMouse (changes requested):
Had some thoughts in the comments. Will switch to Approve if all of them are NO-OP or addressed.

Assets/Plugins/Source/Core/RuntimeConfig.cs
paulhazen: Not as far as I know, but it was the best way that I could think of to make use of explicit emphasis.

Assets/Plugins/Source/Core/RuntimeConfig.cs
paulhazen: Yup! Nice catch. I fixed this one, marking conversation as resolved.

Assets/Plugins/Source/Core/RuntimeConfig.cs
paulhazen: Ultimately, this will be done, however, as we discovered in the course of this work the `EOSConfig` does not actually _have_ those flags as field members. This is a bug in the current implementation. This particular item will be fixed once that bug is resolved.

Assets/Plugins/Source/Core/RuntimeConfig.cs
paulhazen: Good catch! It looks like if the sandbox is a "Private" sandbox, then it starts with a p, and is not itself a GUID. I wonder what it _is_.

arthur740212 (approved):

Assets/Plugins/Source/Core/Utility/EnumUtility.cs
```diff
@@ -197,5 +197,37 @@ private static TEnum Combine(TEnum current, TEnum toAdd)
...
+        /// values.
+        /// </summary>
+        /// <param name="bitFlag">
+        /// The possible result of combining enum values using bitwise
+        /// operations.
+        /// </param>
+        /// <returns>
+        /// The discrete values of which this combined flag is composed of.
+        /// </returns>
+        public static IEnumerable<TEnum> Decompose(TEnum bitFlag)
```
arthur740212: This looks like a non config specific thing. Maybe it can be in a "Core" adjacent class

paulhazen: It is a bit of a gray area. My thought process was that the `RuntimeConfig` object would not be able to display a user interface for an enum value that was the result of a bitwise OR operation on multiple values. This code was included to make it clear that aspect of the change was considered. You are correct that it's not strictly speaking config related, even though it will exclusively be used within the context of config operations.

It's a small enough case that I'm comfortable including it, since it doesn't _change_ the functionality of anything, as the function is not called. But I could go either way on it. Lemme know your thoughts and I'll take action.

Assets/Plugins/Source/Editor/ConfigEditors/PlatformConfigEditor.cs
```diff
@@ -54,12 +58,26 @@ public virtual void RenderOverrides()
...
+            //       values are the only ones within "overrideValues" that are
+            //       actually being used to override the otherwise defined
+            //       values within EOSConfig. Changing the values in the fields
+            //       for which the input fields are rendered above does not 
+            //       seem to have any affect. 
+            // 
+            //       This is a bug, but is not relevant to the current task as 
+            //       of this writing, which is to simply add the field member
+            //       "integratedPlatformManagementFlags" to EOSConfig, and make
+            //       sure that field is editable within the editor.
```
paulhazen, line 79: Part of my thought process for not fixing this here is because I started to fix it - and found that doing so is actually not straightforward, and the work would need to be redone in light of the changes we are planning to make to Config validation and platform variation.

Basically, I determined that the time it would take to fix it the way it is was greater than the time it would take to continue on the current development plan.

Assets/Plugins/Source/Core/RuntimeConfig.cs
paulhazen: Marking this conversation as "unresolved," because I chose to include the changes to include the `IntegratedPlatformManagementFlags` field member being added to `EOSConfig` for the sake of clarity. I unresolved it for visibility, feel free to re-resolve it if you're satisfied with this solution.

WispyMouse (approved):
All comments addressed, good to go.

Assets/Plugins/Source/Core/EOSConfig.cs
```diff
@@ -120,6 +123,11 @@ static EOSConfig()
         /// </summary>
         public List<string> authScopeOptionsFlags;
 
+        /// <summary>
+        /// Flags used to configure platform integration management flags.
+        /// </summary>
+        public List<string> integratedPlatformManagementFlags;
```
andrew-hirata-playeveryware: I believe this is better separated from EOSConfig because I don't believe there is a meaningful default besides 'disabled'. Keeping it separate also better mirrors how the EOS SDK itself keeps the flags separate.

Assets/Plugins/Source/Core/RuntimeConfig.cs
```diff
@@ -0,0 +1,259 @@
...
+    /// allows users who are making their game with EOS to define different
+    /// properties for each platform, and to store sandbox / deployment
+    /// overrides so that it is easy to switch between deployment environments
+    /// during development. Those settings are distinct from these set of
+    /// values.
+    ///
+    /// Most of the values for these field members come from the
+    /// [Development Portal](https://dev.epicgames.com/portal/)
+    /// </summary>
+    public readonly struct RuntimeConfig
```
andrew-hirata-playeveryware: There are two kinds of 'runtime' configuration, bootstrap config values and then 'everything else'. If possible, I would prefer the 'everything else' not to be mixed with the bootstrap configuration values. The reason they were mixed together in the first place is because 1: it was faster 2: it allowed per platform overrides because it could hitch along with the existing override system.

Assets/Plugins/Source/Core/Utility/EnumUtility.cs
andrew-hirata-playeveryware: I think I prefer names of the style `ToEnumerable` or something similar than 'Decompose'. I think it's because `Decompose` is too similar to `Dispose`.

andrew-hirata-playeveryware (changes requested):

Assets/Plugins/Source/Core/EOSConfig.cs
paulhazen: The problem with separating the value from `EOSConfig` proper is that it makes it impossible to make use of the flags on Windows, since the WindowsConfig was not designed to be separate from `EOSConfig`.

In _future_ implementations that will be upcoming, this value and field member will be more appropriately placed. The reason I chose to include this addition to the existing config implementation when the brunt of this PR is to define `RuntimeConfig` is because it made the implicit conversion operator that I defined for that struct complete in that the full set of runtime configuration values are represented within, and `EOSConfig` can be converted to and from it effectively.

Assets/Plugins/Source/Core/RuntimeConfig.cs
paulhazen: Which of these values pertain to bootstrap configuration?

Assets/Plugins/Source/Core/EOSConfig.cs
andrew-hirata-playeveryware: Adding the IntegratedPlatformManagmentFlags to `EOSConfig` now also means that users will have config files with the flags both in their base EOSConfig and in their platform specific flags in the future, if we do a release between now and when that future implementation is done. In other words, we would be doing a field add, then needing to deprecate and migrate the information to the appropriate place in future releases, as apposed to just creating the field in a new place.

Assets/Plugins/Source/Core/RuntimeConfig.cs
andrew-hirata-playeveryware: It's a little bit easier to answer it the other way around: 

These values are _not_ boostrap configuration, i.e. they're not needed for the EOS SDK bootstrap.
```
    public readonly bool AlwaysSendInputToOverlay;
    public readonly float InitialButtonDelayForOverlay;
    public readonly float RepeatButtonDelayForOverlay;
    public readonly bool SendInputDirectlyToSDK;
```

andrew-hirata-playeveryware (changes requested):

Assets/Plugins/Source/Core/EOSConfig.cs
paulhazen: If any user has the field defined in the other platform file, it isn't being collected and applied by anything anywhere, so it's absence cannot possibly be missed.

Similarly, an extra field in the JSON does not cause an error to be thrown if the corresponding field member does not exist in the object that is being deserialized.

Assets/Plugins/Source/Core/RuntimeConfig.cs
paulhazen: Got it. Just to clarify - these ARE currently defined within EOSConfig, so you are suggesting that we redefine these to not be set at runtime? That change seems out of scope with the intent of the PR.

andrew-hirata-playeveryware: I'm saying for the new runtime config file, it would be nice to have the 'bootstrap' values separate from the 'other' values for the runtime config.

paulhazen: Per conversation "offline" and in person, the definition of bootstrap in this context refers to the configuration values that are utilized during the setup / initialization of the plugin, not settings that are _required_ components of making the EOS plugin work.

Assets/Plugins/Source/Core/EOSConfig.cs
paulhazen: I have updated the PR to remove this, putting it back within `PlatformConfig`.

Assets/Plugins/Source/Core/RuntimeConfig.cs
paulhazen: I removed these values from `RuntimeConfig`.

andrew-hirata-playeveryware (approved):

PR #799, opened 2024-07-24
fix: Prevent placeholder `.meta` files from being deleted.

paulhazen:
In the current state of the project, because the contents of `com.playeveryware.eos` are represented in the Packages directory within the Unity Editor, Unity handles the presence of unassociated `.meta` files by deleting them. 

To bridge the gap between our old package creation process and the direction we are heading in, I added two empty `.gitignore` files to the corresponding directories that the placeholder `.meta` files are meant to represent. This prevents Unity from deleting them.


andrew-hirata-playeveryware (approved):

arthur740212 (approved):

WispyMouse (approved):
I wonder what is meant to be under these folders, and what the gitignore should hit. But since I don't see it in the repo or my local, I'll take your word for it on this being good to keep!

PR #800, opened 2024-07-24
feature(network): Implementing taskNetworkTimeoutSeconds

WispyMouse:
Implements "TaskNetworkTimeoutSeconds". To validate;

- Use the EOS Configuration pane to set it to a value, either an extreme partial-second or very long time
- Build the game
- Launch a sample with network operations, such as Player Data Storage
- Authorize
- Validate that networking works as expected
- Disconnect entirely from the internet
- Wait until the EOS SDK realizes that the internet has been disconnected
- Attempt a network operation
- Observe that it fails after the timeout you've specified, perhaps then automatically retrying at the same interval

https://dev.epicgames.com/docs/game-services/eos-platform-interface#creating-the-platform-interface


WispyMouse:
What unit tests would be appropriate here? I can validate the "to double" function _maybe_, but that also might not be accessible on the Unity layer. I can validate that the setting sets properly and gets applied, but I'm not really sure how to best simulate network connectivity issues with the plugin for making sure the timeout works as expected.

Maybe: Find a sample that does networking fudging, and observe that the operation fails after a specific amount of time?

paulhazen (approved):
Initially, I had some concerns regarding the way this handles the rendering of the editor window UI fields for changing the value, but upon inspection I was reminded that the newer sleeker way of doing things had not yet been brought to this area of the code, so it makes sense to do things the same way they are currently being done.

All that to say, these changes look good, and you can ignore the concerns I voiced offline earlier.

Assets/Plugins/Source/Editor/EditorWindows/EOSSettingsWindow.cs
```diff
@@ -282,6 +282,32 @@ private void OnDefaultGUI()
...
+
+            if (taskNetworkTimeoutSecondsAsString.Length != 0)
+            {
+                try
+                {
+                    mainEOSConfigFile.taskNetworkTimeoutSeconds = Convert.ToDouble(taskNetworkTimeoutSecondsAsString);
+                }
+                catch
+                {
+
```
paulhazen: Probably be good to throw in a `Debug.LogWarning` statement here.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/dllmain.cpp
```diff
@@ -153,6 +153,7 @@ struct EOSConfig
     uint64_t flags = 0;
 
     uint32_t tickBudgetInMilliseconds = 0;
+    double taskNetworkTimeoutSeconds = 0;
```
paulhazen: Should probably be `0.0`

andrew-hirata-playeveryware (approved):

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/dllmain.cpp
WispyMouse: Changed! Interestingly the compiler already treated this as `0.0`, I can see from hovering over it.

PR #801, opened 2024-07-24
fix(builds): Fixing changed function name on build

WispyMouse:
When making a build, this was failing because the function wasn't found. Pointed it at its new home.


arthur740212 (approved):

paulhazen (approved):
Good catch.

PR #802, opened 2024-07-25
tests(networking): Test for TaskNetworkTimeoutSeconds

WispyMouse:
This adds a test for the TaskNetworkTimeoutSeconds.

I expect this to be controversial and in need of discussion!

- Made the login method of EOSTestBase static, so non-inheritors can call it
- The test class we made tears down and sets up the entire EOS SDK each time it is run, for each test
- In order to get the EOSConfig up to where it needs to be, added parameters that override the test

I would really like suggestions on how to better manage the EOSConfig needing customization while creating a new platform. It's definitely not hitting all of the places. Maybe there's something we could add to the Config base class that lets us register an overriding EOSConfig to use?

### DIFF
--- a/Assets/Plugins/Source/Core/EOSManager.cs
+++ b/Assets/Plugins/Source/Core/EOSManager.cs
@@ -145,6 +145,11 @@ namespace PlayEveryWare.EpicOnlineServices
         //cached log levels for retrieving later
         private static Dictionary<LogCategory, LogLevel> logLevels;
 
+        /// <summary>
+        /// If provided, this configuration file will be used when the Platform Interface is created.
+        /// </summary>
+        public EOSConfig OverridingEOSConfigForPlatformCreate { get; set; } = null;
+
         enum EOSState
         {
             NotStarted,
@@ -427,9 +432,20 @@ namespace PlayEveryWare.EpicOnlineServices
             }
 
             //-------------------------------------------------------------------------
-            private Result InitializePlatformInterface()
+            private Result InitializePlatformInterface(EOSConfig overridingConfig = null)
             {
-                EOSConfig configData = Config.Get<EOSConfig>();
+                EOSConfig configData;
+                
+                if (overridingConfig == null)
+                {
+                    configData = Config.Get<EOSConfig>();
+                }
+                else
+                {
+                    print("InitializePlatformInterface: overridingConfig provided, using this for platform create");
+                    configData = overridingConfig;
+                }
+                
                 IPlatformSpecifics platformSpecifics = EOSManagerPlatformSpecificsSingleton.Instance;
 
                 print("InitializePlatformInterface: platformSpecifics.GetType() = " + platformSpecifics.GetType());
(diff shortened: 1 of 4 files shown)


Assets/Tests/PlayMode/EOSConfigurationTests.cs
```diff
@@ -0,0 +1,257 @@
...
+    using Epic.OnlineServices.TitleStorage;
+    using Epic.OnlineServices.Platform;
+    using PlayEveryWare.EpicOnlineServices.Tests;
+    using static PlayEveryWare.EpicOnlineServices.EOSManager;
+
+    /// <summary>
+    /// Tests for values set in <see cref="EOSConfig"/> and how they interact with the EOS SDK.
+    /// This test repeatedly sets up and tears down the EOS Libraries, so it doesn't inherit from the <see cref="EOSTestBase"/>.
+    /// </summary>
+    public class EOSConfigurationTests
```
paulhazen, line 42: This class should inherit from `EOSTestBase` - as that class contains a lot of the functionality that is common to all tests (for example, this is why the `TestCategories` class was made to be `protected`, because it should only be utilized within classes that extend from `EOSTestBase`).

Assets/Tests/Common/EOSTestBase.cs
```diff
@@ -77,20 +95,23 @@ public WaitUntilDone(float timeout, Func<bool> predicate)
...
         /// Initialize the EOSManager once before attempting to log in.
         /// </summary>
         [OneTimeSetUp]
         public void SetupScene()
         {
             eosObject = new GameObject();
-            var eosManager = eosObject.AddComponent<EOSManager>();
-            EOSManager.Instance.Init(eosManager);
+            EOSManager = eosObject.AddComponent<EOSManager>();
+            EOSManager.Instance.Init(EOSManager);
```
paulhazen, line 108: This should not be inside the base class `EOSTestBase` because there are unit tests that extend from `EOSTestBase` that do not require the presence of an `EOSManager` instance, and requiring them to nonetheless make an instance is incorrect.

Assets/Plugins/Source/Core/EOSManager.cs
```diff
@@ -145,6 +145,11 @@ public partial class EOSManager : MonoBehaviour, IEOSCoroutineOwner
         //cached log levels for retrieving later
         private static Dictionary<LogCategory, LogLevel> logLevels;
 
+        /// <summary>
+        /// If provided, this configuration file will be used when the Platform Interface is created.
+        /// </summary>
+        public EOSConfig OverridingEOSConfigForPlatformCreate { get; set; } = null;
```
paulhazen, line 151: Since there is parallel work being done regarding fixing the override functionality of the configuration values, I think that this change is at odds with changes that are already in our development pipeline.

However, if this is the _only_ way to make the `feature/timeout` branch function properly, that would be understandable - but even if that's the case I think the function that accomplishes making it function properly should be done with a more narrowly described function.

paulhazen (changes requested):

Assets/Tests/PlayMode/EOSConfigurationTests.cs
WispyMouse: I disagree here. EOSTestBase authenticates and logs in to Epic automatically. I _don't_ want that to happen, I need the specific state management in order to load and unload the SDK.

Assets/Tests/Common/EOSTestBase.cs
WispyMouse: This was already happening, though. This is the [OneTimeSetUp]. All that is changed here is I'm assigning a variable to a place it can be found.

Assets/Plugins/Source/Core/EOSManager.cs
WispyMouse: I'm comfortable waiting until the development pipeline is along and committed to the point where I can use that override. Do you have any estimate on how much work needs to be done until that is available?

WispyMouse:
Noting here: The plan is to hold off on merging this over until after the rework and retooling on tests has been made. This will be a testbed for iterating on the design of tests.

WispyMouse:
Just to remind myself: One of the things to change was the way that custom test config files can be passed to the initialization.

arthur740212 (approved):

andrew-hirata-playeveryware (approved):

WispyMouse:
While the custom config might become available at some point, I think it makes sense to close this PR for now. We'll revisit the ideas present here when the time is more appropriate.

PR #803, opened 2024-07-25
chore(macOS): Putting MacOS Binaries in Git LFS

WispyMouse:
This puts three files in to Git LFS as binaries.

Are these all of the MacOS binaries that need to be in LFS? If there is more, this is a good time to add them to LFS. Also does the inclusion in to Git LFS look correctly configured?


arthur740212 (approved):
Yep, lfs looks right.
And yes, there are two libs by us, and one from eos

Assets/Plugins/macOS/.gitattributes
```diff
@@ -1 +1,3 @@
-libEOSSDK-Mac-Shipping.dylib filter=lfs diff=lfs merge=lfs binary
\ No newline at end of file
+libEOSSDK-Mac-Shipping.dylib filter=lfs diff=lfs merge=lfs binary
+libDynamicLibraryLoaderHelper.dylib filter=lfs diff=lfs merge=lfs binary
+MicrophoneUtility_macos.dylib filter=lfs diff=lfs merge=lfs binary
```
paulhazen: It's not necessary to rectify this - so not a critique, but in case it was something not previously known, the contents of this file could be simplified to be the following:
```gitattributes
*.dylib filter=lfs diff=lfs merge=lfs binary
```

paulhazen (approved):
Looks good. Please feel free to resolve the one conversation I opened as it was merely a comment.

Assets/Plugins/macOS/.gitattributes
WispyMouse: I've had trouble getting LFS to add things when assigning like this. Now that it's in LFS, maybe it'll work? I'll go change this.

PR #808, opened 2024-07-27
fix: Consolidate test functionality of client sessions tests.

paulhazen:


andrew-hirata-playeveryware (approved):

Assets/Tests/Common/EOSTestBase.cs
paulhazen, line 1: Please squash the two commits into one - for the sake of greater clarity.

paulhazen: Done! Nice catch @paulhazen

Assets/Tests/PlayMode/ClientSessionTests.cs
```diff
@@ -117,19 +115,31 @@ public IEnumerator FindByBucketIdAndJoin()
             sessionSearchHandle.Find(ref findOptions, null, (ref SessionSearchFindCallbackInfo data) => { searchResult = data; });
 
             yield return new WaitUntil(() => searchResult != null);
```
WispyMouse, line 117: Not changed in scope; I wonder if WaitUntil should have a timeout here, so it doesn't roll infinitely.

WispyMouse (changes requested):
Running the test twice currently always hangs Unity. I think the TearDown in EOSTestBase isn't implemented correctly. Either this needs to be determined as being out of the scope of this PR or needs to be fixed, since we can't really use the tests as is.

paulhazen:
> Running the test twice currently always hangs Unity. I think the TearDown in EOSTestBase isn't implemented correctly. Either this needs to be determined as being out of the scope of this PR or needs to be fixed, since we can't really use the tests as is.

Yeah, I noticed this as well. Out of scope as far as I can tell, as I experienced similar behavior with the tests in the first place.

WispyMouse (approved):
Changing to approve. A fix for the bug I identified is in this PR: https://github.com/PlayEveryWare/eos_plugin_for_unity/pull/811

PR #809, opened 2024-07-27
Fix Custom Mapping Enum Tests

paulhazen:
Consolidate test functionality of custom mapped enums to a base class, and add verbose messaging when tests fail.

> [!IMPORTANT]
> What's great about this PR is that running the tests for the custom mapped enums actually revealed some gaps in their implementation! Real-time example of the value of testing :)


andrew-hirata-playeveryware (approved):

Assets/Tests/PlayMode/Utility/Extensions/AuthScopeFlagsExtensionsTests.cs
```diff
@@ -20,16 +20,13 @@
  * SOFTWARE.
  */
 
-using NUnit.Framework;
-
-namespace PlayEveryWare.EpicOnlineServices.Tests
+namespace PlayEveryWare.EpicOnlineServices.Tests.Utility.Extensions
 {
+    using NUnit.Framework;
```
arthur740212, line 25: What is NUnit.Framework for?

WispyMouse (approved):
Looks good to me.

I wonder if there's a way to set it up so that a parent class (like CustomMappedEnumTestBase) has a test function, which is inherited by the child (like AuthScopeFlagsExtensionsTests), that uses the configuration type (like AuthScopeFlags) as part of the inheriting line, which would automagically create tests for the enum with minimal lines of text needed to define those additional tests.

Assets/Tests/PlayMode/Utility/Extensions/AuthScopeFlagsExtensionsTests.cs
paulhazen: It is the unit test framework used within Unity.

PR #810, opened 2024-07-29
feat: Add support for conversion from `EOSConfig` to `RuntimeConfig`

paulhazen:
This PR introduces use of the already included `Unity.Plastic.Newtonsoft.Json` package. This _does not actually introduce a dependency for the project_ but makes use of the Unity package version of Newtonsoft that is included already in the project. (Prior attempts to introduce Newtonsoft failed because we were trying to introduce the library independent of the Unity package system).

This PR also moves the `JsonUtility` implementation away from using `UnityEngine.JsonUtility` under the covers, and instead makes use of the aforementioned `Unity.Plastic.Newtonsoft.Json` library for implementation, adding where necessary the custom Json converters for data types that are needed to accomplish serialization of the following types:

- `System.Guid`
- `System.Version`
- `Epic.OnlineServices.Platform.ClientCredentials`


Assets/Plugins/Source/Core/RuntimeConfig.cs
```diff
@@ -0,0 +1,332 @@
...
+                       Math.Abs(InitialButtonDelayForOverlay - other.InitialButtonDelayForOverlay) < FloatEqualityTolerance &&
+                       Math.Abs(RepeatButtonDelayForOverlay - other.RepeatButtonDelayForOverlay) < FloatEqualityTolerance &&
+                       SendInputDirectlyToSDK == other.SendInputDirectlyToSDK;
+            }
+            return false;
+        }
+
+        public override int GetHashCode()
+        {
+            HashCode hash = new();
```
WispyMouse: oh WHAT. I'm using this from now on when making HashCodes, I was super tired of making up baloney on the spot. Had no idea you could just do this!

WispyMouse (approved):
So far this seems good to me. There are other places JsonUtility are still used, but we'll consider that outside the scope of this PR.

paulhazen:
> So far this seems good to me. There are other places JsonUtility are still used, but we'll consider that outside the scope of this PR.

All places that use `JsonUtility` should be using _our_ version of the static class. Are you finding examples to the contrary?

WispyMouse:
That's what I had intended by what I said; BUT I went through the code again and can't find any examples of this within our own codebase. It seems like everything that references JsonUtility is now referencing our version, except for some unity internal stuff that we don't have control over. So I rescind the previous statement, this seems like it comprehensively hits all JsonUtility usages.

PR #811, opened 2024-07-29
fix(tests): OnShutdown during TearDown

WispyMouse:
In tests, during the one time tear down of the parent EOSTestBase class, explicitly shut down all of EOS.

Before this change, the Editor would hang 100% of the time when attempting a second test run after the first.


andrew-hirata-playeveryware (approved):

paulhazen (approved):
Awesome! This has been bugging me for a while, but I was too lazy to look into it. Shame on me, kudos to you!

PR #813, opened 2024-07-30
audit(login): Manager Re-Authorization Audit

WispyMouse:
This PR seeks to address the refreshing of authentication tokens for our sample's manager classes: Discord, Steam (App and Session), OpenID

Before this change, these platforms would not ever invalidate their token, even on a login failure. With this change each of the managers will now attempt to login, on a specific error invalidate the token, then retry once. The Discord approach now includes refresh tokens, which should now be used when reauthing rather than dropping the auth.

I've only gotten Steam App to re-acquire the token, because it is the only one that expires within a few days. The rest of the tokens theoretically re-auth appropriately, but they haven't hit that scenario in my testing yet. I also haven't run OpenID at all. This will require additional testing input.

image

This changelist includes something I'm considering entirely optional to this effort, which is a scene for testing the authentication and an associated script. It's how I've been trying to test the auth. It isn't visually in line with our other testing scenes. I'd be comfortable cutting it entirely from this changelist, if so desired.

My plan is to continue running this until the reproduction fix scenario is reproduced. This is scheduled to happen only after a week from now.


Assets/Scripts/StandardSamples/Discord/DiscordManager.cs
```diff
@@ -121,13 +123,54 @@ private class TokenReader
         {
             public string access_token;
             public float expires_in;
+            public string refresh_token;
         }
 
         private string cachedToken = null;
+        private string refresh_token = null;
```
paulhazen: This should be `camelCase` instead of `snake_case`.

Assets/Scripts/StandardSamples/Discord/DiscordManager.cs
```diff
@@ -121,13 +123,54 @@ private class TokenReader
...
         private DateTime tokenExpiration;
 
         private event Action<string> onTokenReceived;
 
+        public void StartConnectLoginWithDiscord(EOSManager.OnConnectLoginCallback onLoginCallback)
+        {
+            StartConnectLoginWithDiscord(onLoginCallback, false);
+        }
+
+        private void StartConnectLoginWithDiscord(EOSManager.OnConnectLoginCallback onLoginCallback, bool isRetry = false)
```
paulhazen: Out of curiosity (and depending on the answer this may be a critique), is there a reason to use the suffix `WithDiscord`, given the context that these functions are defined within the `DiscordManager`? Isn't `WithDiscord` therefore an extraneous descriptor? My thought process is that if this is the case, then reducing the function name to exclude the suffix might make it helpfully align with existing interfaces.

Assets/Scripts/StandardSamples/Discord/DiscordManager.cs
```diff
@@ -121,13 +123,54 @@ private class TokenReader
...
+
+                // Attempt to login to Discord
+                EOSManager.Instance.StartConnectLoginWithOptions(ExternalCredentialType.DiscordAccessToken, returnedToken, onloginCallback: (LoginCallbackInfo callbackInfo) =>
+                {
+                    if (!isRetry && callbackInfo.ResultCode == Result.ConnectExternalTokenValidationFailed)
+                    {
+                        // If the auth failed for exactly this reason, then try logging in again, doing the whole process again to acquire a new token
+                        cachedToken = null;
+                        StartConnectLoginWithDiscord(onLoginCallback, true);
+                    }
```
paulhazen: As a general principle, when I have a bit of code that needs to "retry" a connection, I like to structure it in such a way that the number of retries is variable, so that in the future if more than one retry is called for, that's easy to scale.

In my experience, this makes for cleaner, clearer code. Additionally, implementation of that "retrying" ought to be accomplished via a helper method that makes it explicit that a retry will be attempted.

This is more a thought than an explicit critique; I think it's an idea worth considering, but not a concept upon which I would ask for changes to be made.

Assets/Scripts/StandardSamples/Discord/DiscordManager.cs
```diff
@@ -157,8 +200,11 @@ private void GetTokenAsyncComplete(Task<TokenReader> tokenTask)
             if (tokenContent != null)
             {
                 cachedToken = tokenContent.access_token;
-                //in a real application the refresh token would be stored as well and used to refresh the auth token
+
                 tokenExpiration = DateTime.Now + TimeSpan.FromSeconds(tokenContent.expires_in-60);
+                
+                // Set the refresh token so next grant attempt can try to use it
+                refresh_token = tokenContent.refresh_token;
```
paulhazen: What happens if multiple calls are made and the `refresh_token` value is overwritten? Is that a scenario that could cause problems, or is the nature of it such that it wouldn't be a problem?

Assets/Scripts/StandardSamples/Discord/DiscordManager.cs
```diff
@@ -202,6 +248,32 @@ private async Task<TokenReader> GetTokenAsync()
             }
 
             //use received auth code to request auth token
+            const string grant_type = "grant_type";
```
paulhazen: Should be `camelCase`. From context, I can see that you got the variable name from the altered code, wherein the original value was `"grant_type"`. I still believe it should be changed, but I can see how the name was determined.

Assets/Scripts/StandardSamples/OpenID/OpenIdTokenManager.cs
```diff
@@ -20,6 +20,8 @@
 * SOFTWARE.
 */
 
+using Epic.OnlineServices;
+using Epic.OnlineServices.Connect;
 using System;
 using System.Collections;
 using System.Collections.Generic;
```
paulhazen: You had previously (and rightly in my mind) expressed a preference for placing `using` statements within `namespace` code blocks. This would be a good opportunity to make that change at least in this file.

Assets/Scripts/StandardSamples/OpenID/OpenIdTokenManager.cs
```diff
@@ -28,6 +30,7 @@
 
 namespace PlayEveryWare.EpicOnlineServices.Samples.OpenId
 {
+    using static PlayEveryWare.EpicOnlineServices.EOSManager;
```
paulhazen, line 32: +1 for `using static`, I was previously unaware of or had forgotten its benefits, you make me write better code! 😃

Assets/Scripts/StandardSamples/Steam/SteamManager.cs
```diff
@@ -474,14 +500,29 @@ public void StartConnectLoginWithSteamAppTicket(EOSManager.OnConnectLoginCallbac
                     // Now add to the callback a step wherein, if it fails from this point, the cached encryptedAppTicket will be invalidated
                     EOSManager.OnConnectLoginCallback cacheInvalidationOnFailureCallback = (Epic.OnlineServices.Connect.LoginCallbackInfo callbackInfo) =>
                     {
-                        if (callbackInfo.ResultCode != Epic.OnlineServices.Result.Success)
+                        if (!isReattempt && callbackInfo.ResultCode == Epic.OnlineServices.Result.ConnectExternalTokenValidationFailed)
+                        {
+                            Debug.Log($"Connect Login failed: SteamManager successfully retrieved an app ticket from Steam, but the provided app ticket was invalid for logging in to Epic Online Services. Epic identified the ticket as not being valid. Re-attempting to acquire ticket and try logging in again. The cached app ticket in `{nameof(encryptedAppTicket)}` will now be invalidated.");
```
paulhazen, line 509: Nice log statement. Plenty verbose 👍

Assets/Scripts/StandardSamples/UI/Common/UIDebugLog.cs
```diff
@@ -77,7 +77,11 @@ private void Start()
             logLevelMenuItems = new List<UIDebugLogLevelMenuItem>();
             BuildLogLevelMenu();
             ignoreLogLevelChange = false;
-            LogLevelScrollView.gameObject.SetActive(false);
+
+            if (LogLevelScrollView != null)
+            { 
+                LogLevelScrollView.gameObject.SetActive(false);
```
paulhazen: Would this be a good case to use `null` propagation? 
```csharp
LogLevelScrollView?.gameObject.SetActive(false);
```
I know that in certain areas of Unity, use of `null` propagation is discouraged, but I never understood why or what the borders of that recommendation are.

Assets/Scripts/StandardSamples/UI/Common/UIDebugLog.cs
```diff
@@ -341,9 +350,12 @@ private void Update()
...
             // FPS
-            deltaTime_FPS += (Time.deltaTime - deltaTime_FPS) * 0.1f;
-            float fps = 1.0f / deltaTime_FPS;
-            FPSValue.text = Mathf.Ceil(fps).ToString();
+            if (FPSValue != null)
+            {
+                deltaTime_FPS += (Time.deltaTime - deltaTime_FPS) * 0.1f;
+                float fps = 1.0f / deltaTime_FPS;
+                FPSValue.text = Mathf.Ceil(fps).ToString();
+            }
```
paulhazen: Would be good to check to prevent accidental divide-by-zero. I would favor this version personally:
```csharp
if (FPSValue != null)
{
    deltaTime_FPS += (Time.deltaTime - deltaTime_FPS) * 0.1f;
    float fps = (deltaTime_FPS != 0) ? 1.0f / deltaTime_FPS : float.PositiveInfinity;
    FPSValue.text = Mathf.Ceil(fps).ToString();
}
```
A clearer implementation that avoids the inclusion of a ternary statement would be:
```csharp
if (FPSValue != null)
{
    deltaTime_FPS += (Time.deltaTime - deltaTime_FPS) * 0.1f;
    
    // Prevent division by zero by checking if deltaTime_FPS is not zero
    if (deltaTime_FPS != 0)
    {
        float fps = 1.0f / deltaTime_FPS;
        FPSValue.text = Mathf.Ceil(fps).ToString();
    }
    else
    {
        FPSValue.text = "Inf"; // Or handle the case as needed
    }
}
```

_Either way_, what is the purpose of multiplying `(Time.deltaTime - deltaTime_FPS)` by `0.1f`?

paulhazen (changes requested):
Only items that I fervently believe warrant change are instances where `snake_case` is used where `camelCase` should be, other comments are more subjective in nature and might be topics for conversation.

Assets/Scripts/StandardSamples/Discord/DiscordManager.cs
WispyMouse: This is resolved as you suggested, changing the name of this to refreshToken.

Assets/Scripts/StandardSamples/Discord/DiscordManager.cs
WispyMouse: Given the location of this function, there is no need for it to be "WithDiscord". Resolving by removing as suggested.

Assets/Scripts/StandardSamples/Discord/DiscordManager.cs
WispyMouse: I find this agreeable. These boolean retry arguments are being traded for ints that count the number of retries, and a const int inside the function that gates how many times to retry.

Assets/Scripts/StandardSamples/Discord/DiscordManager.cs
WispyMouse: Noting that this is likely a problem, but we consider it outside the scope of this PR. I'll note this for later in ClickUp.

Assets/Scripts/StandardSamples/Discord/DiscordManager.cs
WispyMouse: Resolved by changing as you suggested. I made this "grantTypeString". Do you think this name doesn't need the -String suffix? I want to make sure its purpose is clear, to have one fewer magic string.

Assets/Scripts/StandardSamples/OpenID/OpenIdTokenManager.cs
WispyMouse: Resolved by moving the statements into the namespace.

Assets/Scripts/StandardSamples/UI/Common/UIDebugLog.cs
WispyMouse: Resolved as you suggested; this is a clear time to use the null propagation technique.

Assets/Scripts/StandardSamples/UI/Common/UIDebugLog.cs
WispyMouse: Resolved as recommended, though I'm using "NaN" instead of infinite.

I have.. no idea why this is multiplied by .1f. Honestly this bit of code doesn't make sense to me.

WispyMouse:
Remaining thing to think on: The location of the test script, and the location of the test scene.

Assets/Scripts/StandardSamples/Discord/DiscordManager.cs
paulhazen: I like the practice of having one fewer magic strings. I think it could go either way. I think there is some conflict with code style guideline stuff to do this - but in these cases I will opt to use the `Str` suffix as shorthand, just to indicate that it's a string without adding the full name. Could really go either way. Certainly `String` as a suffix is clearer.

paulhazen (approved):
Marking as approved per the requested changes being made. Great work Mint! 
I left one conversation unresolved, just so that it did not go unnoticed, but once it is read feel free to mark it as resolved in order to merge.

andrew-hirata-playeveryware (approved):

PR #814, opened 2024-07-31
tests(sessions): First batch of tests

WispyMouse:
A set of tests for the EOS Manager. All of these tests run to completion correctly on my local. I would love feedback on the naming, structure, and overall everything about these tests. The tests included here are only tests that can run with a single user, and does not include a lot of tests that should exist in relation to joining other people's Sessions.

### DIFF
--- a/Assets/Scripts/EOSSessionsManager.cs
+++ b/Assets/Scripts/EOSSessionsManager.cs
@@ -204,6 +204,34 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
         }
     }
 
+    public struct SessionsManagerCreateSessionCallbackInfo
+    {
+        public string SessionToCreateName;
+        public Result ResultCode;
+
+        public SessionsManagerCreateSessionCallbackInfo(string sessionToCreateName, Result resultCode)
+        {
+            SessionToCreateName = sessionToCreateName;
+            ResultCode = resultCode;
+        }
+    }
+
+    public delegate void SessionsManagerCreateSessionCallback(SessionsManagerCreateSessionCallbackInfo info);
+
+    public struct SessionsManagerDestroySessionCallbackInfo
+    {
+        public string SessionToDestroyName;
+        public Result ResultCode;
+
+        public SessionsManagerDestroySessionCallbackInfo(string sessionToDestroyName, Result resultCode)
+        {
+            SessionToDestroyName = sessionToDestroyName;
+            ResultCode = resultCode;
+        }
+    }
+
+    public delegate void SessionsManagerDestroySessionCallback(SessionsManagerDestroySessionCallbackInfo info);
+
     /// <summary>
     /// Class represents a single Session attribute
     /// </summary>
@@ -508,26 +536,6 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
 
         // UI Parameterized Callbacks
 
-        /// <summary>
-        /// This event reports when a Session has been attempted to be created.
-        /// The first argument to the event is the local name of the Session.
-        /// The second argument to the event is the Result code of the attempt.
-        /// Running does not guarantee that a Session was successfully created.
-        /// Get local Sessions by name with <see cref="TryGetSession(string, out Session)"/>.
-        /// </summary>
-        public UnityEvent<string, Result> UIOnSessionCreated;
-
-        /// <summary>
-        /// This event reports when a Session has been attempted to be 'destroyed', or left.
-        /// The first argument to the event is the local name of the Session.
-        /// The second argument to the event is the Result code of the attempt.
-        /// Running does not guarantee that a Session was successfully destroyed..
-        /// If the Result is Success, it is likely that the local Session has been destroyed,
-        /// and is no longer available locally. Otherwise, get local Sessions by name
-        /// using <see cref="TryGetSession(string, out Session)"/>.
-        /// </summary>
-        public UnityEvent<string, Result> UIOnSessionDestroyed;
-
         // TODO: Because the following four UI Action Queues are never set,
         // they should either be wired up fully, or removed.
         Queue<Action> UIOnSessionModified;
@@ -721,8 +729,6 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
 
             userLoggedIn = false;
 
-            UIOnSessionCreated = new();
-            UIOnSessionDestroyed = new();
             UIOnSessionRefresh = null;
 
             UIOnSessionModified = new Queue<Action>();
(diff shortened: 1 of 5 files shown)


Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -814,9 +818,37 @@ public void OnLoggedOut()
         [Obsolete]
         private void OnShutDown()
         {
-            DestroyAllSessions();
+            LeaveAllSessions();
```
paulhazen: I note that you're adding a comment to a method marked as `[Obsolete]`. It feels like either this should not be done, or the `[Obsolete]` attribute should be removed.

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -814,9 +818,37 @@ public void OnLoggedOut()
...
             UnsubscribeFromGameInvites();
             UnsubscribeToSessionMessageConnectionRequests();
+
+            ResetActionsAndLocalVariables();
+        }
+
+        private void ResetActionsAndLocalVariables()
```
paulhazen: I believe that a more appropriate name for this function might be `Initialize`. `ResetActionsAndLocalVariables` can be confusing if it doesn't reset _all_ of them or doesn't otherwise specify which ones are being reset. The term `Init` or `Initialize` makes it more ambiguous, but more accurate as a description IMO.

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -979,23 +1013,32 @@ private static void OnSetPresenceCompleteCallback(ref SetPresenceCallbackInfo da
         /// <param name="callback">Callback to add to <see cref="UIOnSessionCreated"/>. Invoked in <see cref="OnUpdateSessionCompleteCallback_ForCreate"/>.</param>
         /// <returns>True if the configuration and attempt to create a Session succeed. Does not indicate that the Session was created successfully in Epic Online Services,
         /// only that the program was able to attempt its creation and without running in to an error.</returns>
-        public bool CreateSession(Session session, bool presence = false, Action callback = null)
```
paulhazen, line 982: Since you removed the `callback` parameter, you should also remove the corresponding comment that references it in the comments above.

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -979,23 +1013,32 @@ private static void OnSetPresenceCompleteCallback(ref SetPresenceCallbackInfo da
...
+        public bool CreateSession(Session session, bool presence = false)
         {
             if (session == null)
             {
                 Debug.LogError($"{nameof(EOSSessionsManager)} ({nameof(CreateSession)}): parameter 'session' cannot be null!");
+                UIOnSessionCreated.Invoke(null, Result.InvalidParameters);
```
paulhazen: Instead of invoking this, does it make sense to _not_ invoke it, because the session was not in fact created? If it does make sense for it to be invoked, I would suggest that it might need to be renamed to reflect the fact that it is called whenever an attempt is made to create a session.

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -3426,6 +3420,39 @@ private void HandleRegistrationChangeMessage(string sessionId, string userIdCSV)
             UnregisterPlayers(session.Name, usersToRegister.ToArray());
         }
 
+        private void InformOnDestroy(string sessionName)
+        {
+            Session localSession;
+
+            if (!TryGetSession(sessionName, out localSession) || localSession.ActiveSession == null)
+            {
+                Debug.LogError($"{nameof(EOSSessionsManager)} ({nameof(OnDestroySessionCompleteCallback)}): Could not find local Session and associated ActiveSession, so could not inform Owner/members of destruction.");
```
paulhazen, line 3459: Seems like this ought to be a warning, since it's possible that there are no owners or members to notify, which while potentially concerning, is not an error.

Assets/Scripts/EOSSessionsManager.cs
```diff
@@ -3426,6 +3420,39 @@ private void HandleRegistrationChangeMessage(string sessionId, string userIdCSV)
             UnregisterPlayers(session.Name, usersToRegister.ToArray());
         }
 
+        private void InformOnDestroy(string sessionName)
+        {
+            Session localSession;
+
+            if (!TryGetSession(sessionName, out localSession) || localSession.ActiveSession == null)
```
paulhazen: You can make this cleaner by inlining the definition:

```csharp
if (!TryGetSession(sessionName, out Session localSession) || localSession.ActiveSession == null)
```

Assets/Tests/PlayMode/EOSSessionsManagerTests.cs
```diff
@@ -0,0 +1,654 @@
...
+        {
+            float startTime = Time.realtimeSinceStartup;
+            float endTime = startTime + secondsToWait;
+
+            while (Time.realtimeSinceStartup < endTime)
+            {
+                ManagerInstance.Update();
+                yield return new WaitForEndOfFrame();
+            }
+        }
```
paulhazen: This function does not appear to be referenced anywhere.

Assets/Tests/PlayMode/EOSSessionsManagerTests.cs
```diff
@@ -0,0 +1,654 @@
...
+                ManagerInstance.OnLoggedOut();
+            }
+        }
+
+        /// <summary>
+        /// Essential test of the most basic functions of the Manager.
+        /// Creates a Session, observes that it was created successfully, leaves the Session.
+        /// </summary>
+        [UnityTest]
+        public IEnumerator CreateSessionLeaveSession()
```
paulhazen: Looking at several of the tests defined in this class, I am seeing that there is a not-insignificant overlap between the functionality of each test; particularly the "CreateSession" portion of the test. Could we perhaps extract the portion of each test that is responsible for session creation so that we can consolidate that into a central location for better maintainability?

paulhazen (changes requested):
Scattered notes about a few things can be found in the comments. 

I also wanted to address some naming-convention stuff regarding Unit Tests. We have not codified these yet, but I will be doing so shortly in what will be considered a "living document." The main points that come to mind and pertain to these changes are the following:

1. Use a consistent naming pattern to make it easier to understand and organize tests.

    *Common patterns include:*
    - `MethodName_StateUnderTest_ExpectedBehavior`
    - `When_StateUnderTest_ExpectedBehavior`
    - `Operation_ExpectedResult`

2. Avoid redundancy
    - Test names should not repeat the class name or method name if it is already clear from the context.
    - Focus on the unique aspect of the test scenario

Assets/Scripts/EOSSessionsManager.cs
andrew-hirata-playeveryware: Hmm I don't know if this is Obsolete as much as it's broken behavior: I really believe we should be correctly leaving all sessions and unsubscribing if possible. However it looks like it's been broken for a quite a while.

Assets/Scripts/EOSSessionsManager.cs
andrew-hirata-playeveryware: I agree: a method called `Init` would be a good place to do actually variable initialization, then have a new method called `Reset`, with both the constructor and the `Reset` method calling the `Init` method.

Doing this means that the code stays readable because it's clear that it's resetting things when `Reset` is called, and in the constructor, it's clear that it's Initializing because one can read `Init` being called.

Assets/Scripts/EOSSessionsManager.cs
andrew-hirata-playeveryware: What is the rational behind removing the callback parameter?

Assets/Scripts/EOSSessionsManager.cs
paulhazen: Per conversations, another suggestion might be to use the function name `Reset()`. More specifically - rename _this_ function to `Init`, or something like it, and then have a `Reset()` function that calls it subsequently when you need to 'clear' the values.

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: This is attempted to be resolved by removing OnShutDown, and removing the reference to it in a comment.

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: This is resolved by being implemented as suggested. There is now a `Reset` which calls `Init`, and the `ctor` calls `Init`.

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: Per our conversation, I want to be running the callback/event in case of a failure, as it's the only way to be informed that the creation attempt failed.

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: This will be resolved by using an EOS-style Delegate + Struct callback.

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: This is resolved by demoting it to warning. I'm also demoting the above message.

Assets/Scripts/EOSSessionsManager.cs
WispyMouse: This is resolved by being implemented as you recommended.

WispyMouse: This pattern is also being implemented in two other places.

Assets/Tests/PlayMode/EOSSessionsManagerTests.cs
WispyMouse: Resolved by removing this function. It can be made again if it matters to.

Assets/Tests/PlayMode/EOSSessionsManagerTests.cs
WispyMouse: How do you think we should go about doing this? The `CreateSession` utility function feels like it's already doing this as best as can be done.

WispyMouse:
WRT Paul's change requested comment, can you review the newly renamed test methods?

Assets/Tests/PlayMode/EOSSessionsManagerTests.cs
paulhazen: I agree - I failed to notice that. I think there is still room for greater reduction of duplicated code - but not so significant that it ought to warrant a delay in approval. Put another way: the value of the tests as written outweigh the value of over-engineering a solution to reduce the code footprint to its smallest possible form.

paulhazen (approved):

PR #821, opened 2024-08-01
fix(android): Remove config data parameter from configuresystemoptions

WispyMouse:
This parameter was removed, so when you go to build for android it was complaining about having nothing to overwrite. This seems like the right fix.


arthur740212 (approved):
Yes, this is the fix.
It is likely that other platforms will have this problem too.

andrew-hirata-playeveryware (approved):

PR #822, opened 2024-08-01
`SteamConfig` relocation (from both a code and UI organization perspective)

paulhazen:
This PR addresses the confusion around configuration values for Steam existing alongside platforms like Windows, and macOS. 

These settings are now more appropriately placed within the "EOS Plugin Settings," window. Due to how large the contents of the "EOS Plugin Settings" window have grown to be (particularly when values are imported from Steamworks.NET), that window has been redesigned to feature collapsed sections for each area of configuration.

A notable task that will need to be added to our proverbial docket is to update the documentation to reflect these changes (in places where screenshots of the aforementioned window are featured).

image

This is part of a continuing effort to clean up the config system to more clearly handle platform-per-platform overriding of values, which in turn is part of a continuing effort to provide clear instructions to users when they have an invalid or suspicious set of configuration values that may be causing them problems.


Assets/Plugins/Source/Editor/ConfigEditors/ConfigEditor.cs
```diff
@@ -38,26 +39,101 @@ namespace PlayEveryWare.EpicOnlineServices.Editor
...
+        /// <summary>
+        /// Action that is called when the config editor is expanded.
+        /// </summary>
+        public Action<ConfigEditor<T>> OnExpanded;
+
+        /// <summary>
+        /// Used to animate the expansion and collapse of the config editor if
+        /// doing so is enabled.
+        /// </summary>
+        public AnimBool _animExpanded;
```
WispyMouse: As this is public, should they be named something else? Or is this particular name important?

Assets/Plugins/Source/Editor/ConfigEditors/ConfigEditor.cs
```diff
@@ -148,6 +224,12 @@ protected void RenderConfigFields()
                         case ConfigFieldType.Uint:
                             field.FieldInfo.SetValue(config, GUIEditorUtility.RenderInputField(field.FieldDetails, (uint)field.FieldInfo.GetValue(config), labelWidth));
                             break;
+                        case ConfigFieldType.Button:
+                            if (GUILayout.Button(field.FieldDetails.Label))
+                            {
+                                ((Action)field.FieldInfo.GetValue(config))();
```
WispyMouse: It surprises me that all of the other things call FieldInfo.SetValue, with a FieldInfo.Getvalue to do the thing. I can't tell if this is a problem here, or if it should be calling some kind of SetValue.

Assets/Plugins/Source/Editor/ConfigEditors/ConfigEditor.cs
```diff
@@ -189,5 +283,9 @@ public async Task RenderAsync()
             RenderContents();
         }
 
+        public void Dispose()
```
WispyMouse, line 287: Should the Action assigned to OnExpanded be unset here, so that GC can get it? Maybe Actions don't work that way.

Assets/Plugins/Source/Editor/ConfigEditors/IConfigEditor.cs
```diff
@@ -22,17 +22,28 @@
...
         /// <summary>
         /// Returns the human-readable labelText for the section of configuration.
         /// </summary>
         /// <returns>String representing the section this config controls.</returns>
         string GetLabelText();
 
+        /// <summary>
+        /// Affect the expansion of the ConfigEditor.
+        /// </summary>
+        public void Expand();
```
WispyMouse, line 40: nit: It feels odd to mark these as public, but not the others. As far as I understand for interface, `public` is always implied.

WispyMouse: This comment feels like maybe it should be, "expands the ConfigEditor"? What does it mean to "affect the expansion"?

Assets/Plugins/Source/Editor/Configs/ConfigFieldAttribute.cs
```diff
@@ -65,12 +65,17 @@ public enum ConfigFieldType
         /// A list of strings.
         /// </summary>
         TextList,
+
+        Button,
```
WispyMouse, line 76: Can this be given a comment?

WispyMouse:
The "Update from Steamworks.NET" button isn't functioning. I have Steamworks.NET pulled down and included properly. Gonna try making a build to see if this is being applied right.


EDIT: Ah well of course it wasn't working, I wasn't targeting Windows at the time. Disregard.

Assets/Plugins/Source/Editor/Configs/SteamConfig.cs
```diff
@@ -25,23 +25,61 @@
...
+    using Editor.Utility;
+
     [Serializable]
+    [ConfigGroup("Steam Configuration")]
+    // TODO: Make SteamConfig derive from EditorConfig, and update the native code
+    //       to properly reference the correct file where appropriate.
     public class SteamConfig : Config
     {
+        [ConfigField("Steam Flags", ConfigFieldType.TextList)]
+        [Obsolete] // This is marked as obsolete because it is actually not utilized anywhere.
```
WispyMouse: I want to huddle about this and validate it.

Assets/Plugins/Source/Editor/Configs/SteamConfig.cs
```diff
@@ -25,23 +25,61 @@
...
+        [ConfigField("Steamworks SDK Minor Version", ConfigFieldType.Uint)]
         public uint steamSDKMinorVersion;
 
+        [ConfigField("Steamworks Interface Versions", ConfigFieldType.TextList)]
         public List<string> steamApiInterfaceVersionsArray;
 
+        #endregion
+
+        [ButtonField("Update from Steamworks.NET")]
+        public Action UpdateFromSteamworksNET;
```
WispyMouse, line 58: Can this be assigned to a function, or do we have to do this pattern of setting the Action in the constructor?

Assets/Plugins/Source/Editor/Configs/SteamConfig.cs
```diff
@@ -25,23 +25,61 @@
...
+            UpdateFromSteamworksNET = () =>
+            {
+                string steamworksVersion = SteamworksUtility.GetSteamworksVersion();
+
+                if (Version.TryParse(steamworksVersion, out Version version))
+                {
+                    _ = SafeTranslatorUtility.TryConvert(version.Major, out steamSDKMajorVersion);
+                    _ = SafeTranslatorUtility.TryConvert(version.Minor, out steamSDKMinorVersion);
+
+                    steamApiInterfaceVersionsArray = SteamworksUtility.GetSteamInterfaceVersions();
```
WispyMouse: Since this doesn't require the Version (although it really should have one) can it be outside this if statement block?

WispyMouse:
In EOSSettingsWindow, should there be any remaining references to `steamEOSConfigFile`? It *seems* like it is going to overwrite the Steam configuration with an empty file every time you save the EOS Settings. (the window that Steam configuration isn't on!)

Assets/Plugins/Source/Editor/Utility/GUIEditorUtility.cs
```diff
@@ -22,10 +22,14 @@
 
 namespace PlayEveryWare.EpicOnlineServices.Editor.Utility
 {
+    using Microsoft.SqlServer.Server;
```
WispyMouse: What is this using for?

Assets/Plugins/Source/Editor/Utility/GUIEditorUtility.cs
```diff
@@ -226,6 +230,66 @@ public static void AssigningULongToStringField(string label, ref string value, f
...
+            }
+            catch (OverflowException)
+            {
+            }
+            finally
+            {
+                EditorGUILayout.EndHorizontal();
+            }
+
+            EditorGUIUtility.labelWidth = originalLabelWidth;
```
WispyMouse, line 285: In the above if statement, this can hit return before getting to here. Is this a value that always must be set before exiting the function?

WispyMouse (changes requested):
I think there's a bug with the editor overwriting the Steam Config. Maybe we can also huddle about the [Obsolete] attribute that was added.

Assets/Plugins/Source/Editor/Utility/GUIEditorUtility.cs
paulhazen: You didn't know that the `GUIEditorUtility` is actually based on a massive `SQL` database? haha. No, silly auto using statement that my editor thought I needed based on something I errantly autocompleted from intellisense.

Assets/Plugins/Source/Editor/ConfigEditors/ConfigEditor.cs
paulhazen: Good catch. At some point while developing, I had this temporarily be public so I could more easily play with it. I neglected to subsequently return it to `private`. I've done that now, so marking this conversation as resolved.

Assets/Plugins/Source/Editor/ConfigEditors/ConfigEditor.cs
paulhazen: No, not in this case. The action is not considered a "resource" that needs freeing, it is a value that is assigned to the instance of the `ConfigEditor`, and can only be called with an instance reference.

Assets/Plugins/Source/Editor/Configs/ConfigFieldAttribute.cs
paulhazen: Yup! So added. Marking this conversation as resolved.

Assets/Plugins/Source/Editor/ConfigEditors/IConfigEditor.cs
paulhazen: Reasonable note on the comment - likewise the access modifier being specified. I default to setting the access modifier because of other languages in which you can have protected or private interface functions (at least I think that's why I do it - might just be I'm used to specifying it whenever I'm writing a function).

In any case, I removed the `public` access modifiers, and changed the comments in both the interface `IConfigEditor`, and in `ConfigEditor`.

Assets/Plugins/Source/Editor/ConfigEditors/ConfigEditor.cs
paulhazen: I should not be calling a `SetValue`. The purpose of marking a field member (specifically an action) in a `Config` as a `Button` type is to inject arbitrary functionality into the UI in reaction to the user clicking the button. Because `ConfigEditor` is a template class that takes as a template parameter a `Config` type, and because it thereafter generically deals with each config - rendering the config field members in an automatic manner, it cannot know what action to take when a button is clicked, so the information about what button to render and what action to subsequently invoke must be attached to the `Config` instance.

I have updated this section of code to make that clearer by requiring the field member assigned as a `Button` be an `Action` type.

Assets/Plugins/Source/Editor/Configs/SteamConfig.cs
paulhazen: See my [note](https://github.com/PlayEveryWare/eos_plugin_for_unity/pull/822#discussion_r1701052663) on your other comment, I think it helps explain how this works and why it works the way that it does.

Assets/Plugins/Source/Editor/Configs/SteamConfig.cs
paulhazen: Yep, good catch! I've brought it outside the conditional.

Assets/Plugins/Source/Editor/Utility/GUIEditorUtility.cs
paulhazen: I removed the using (along with a few others) that were unnecessary. Marking this conversation as resolved.

Assets/Plugins/Source/Editor/Utility/GUIEditorUtility.cs
paulhazen: Perhaps - however this is marked as an "added" function by Git / GitHub only because I moved the definition from where it previously was in the file to this location, so fixing this is beyond the scope of this PR.

Assets/Plugins/Source/Editor/Configs/SteamConfig.cs
paulhazen: In light of the fact that this is utilized in the native code, it ought _not_ be marked obsolete. It's possible that this is an indication that this is broken to some extent.

paulhazen: Removed the `[Obsolete]` attribute. Marking this conversation as resolved.

WispyMouse (approved):

arthur740212 (approved):

PR #823, opened 2024-08-02
Transition to `Unity.Plastic.Newtonsoft.Json`

paulhazen:
This PR introduces use of the already included `Unity.Plastic.Newtonsoft.Json` package. This _does not actually introduce a dependency for the project_ but makes use of the Unity package version of Newtonsoft that is included already in the project. (Prior attempts to introduce Newtonsoft failed because we were trying to introduce the library independent of the Unity package system).

This PR also moves the `JsonUtility` implementation away from using `UnityEngine.JsonUtility` under the covers, and instead makes use of the aforementioned `Unity.Plastic.Newtonsoft.Json` library for implementation.


com.playeveryware.eos/Runtime/Core/Utility/JsonUtility.cs
paulhazen, line 1: Can you please confirm that this works when the project (the plugin) is deployed as a package? The inclusion of Newtonsoft might not behave the way you think it does in that context, and we should be certain of that before moving forward with this change. It might be the netcode package for game object that utilizes this but check Package Manager to see what the dependency tree is.

paulhazen: I have confirmed that this functions when deployed as a package. Consequently, I am marking this conversation as resolved.

WispyMouse (approved):

PR #829, opened 2024-08-05
Fix warnings when Steamworks.NET is not setup

paulhazen:
Place const declaration within proper compiler conditional branch so that it does not trigger a warning about an unused variable.

Resolves an issue where the Unity Editor generates warnings when Steamworks.NET is not enabled.


WispyMouse (approved):
Thank you for addressing this warning. Approved.

andrew-hirata-playeveryware (approved):

PR #830, opened 2024-08-07
Fixes for the latest development branch

arthur740212:
fix(test,discord) : Disable Discord within StartConnectLogin for unsupported platforms
`AuthenticationExpirationTestManager` call `DiscordManager.StartConnectLogin` on all platforms, including unsupported ones.
This is to make it compile and return an error log if not supported

fix(eos,disable) : missing EOS_DISABLEs for newly added function calls and files

### DIFF
--- a/Assets/Scripts/StandardSamples/Discord/DiscordManager.cs
+++ b/Assets/Scripts/StandardSamples/Discord/DiscordManager.cs
@@ -41,6 +41,50 @@ namespace PlayEveryWare.EpicOnlineServices.Samples.Discord
     [DisallowMultipleComponent]
     public class DiscordManager : MonoBehaviour
     {
+        public void StartConnectLogin(EOSManager.OnConnectLoginCallback onLoginCallback)
+        {
+            StartConnectLogin(onLoginCallback, 0);
+        }
+
+        private void StartConnectLogin(EOSManager.OnConnectLoginCallback onLoginCallback, int retryAttemptNumber = 0)
+        {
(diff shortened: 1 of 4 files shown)


Assets/Scripts/StandardSamples/Discord/DiscordManager.cs
```diff
@@ -41,6 +41,52 @@ namespace PlayEveryWare.EpicOnlineServices.Samples.Discord
     [DisallowMultipleComponent]
     public class DiscordManager : MonoBehaviour
     {
+        public void StartConnectLogin(EOSManager.OnConnectLoginCallback onLoginCallback)
+        {
+#if !DISABLEDISCORD
```
WispyMouse: I think that this shouldn't be surrounded by this ifdef. Users can only call the `public` version here, which will call the `private` version, and the `private` version has a log about Discord not being supported. I think that when the `public` version is called, it should just call the inner `private` version, so that it will log that error out.

WispyMouse (changes requested):
See comment in regards to ifdef-ing the public version of the function.

Assets/Scripts/StandardSamples/Discord/DiscordManager.cs
arthur740212: You are totally right, fixing that now.

paulhazen (approved):

WispyMouse (approved):

andrew-hirata-playeveryware (approved):

PR #834, opened 2024-08-13
`SampleMenu` & `EOSService`

paulhazen:
This PR consolidates some implementation from scenes into the base class `SampleMenu`. This is done preparatory to a PR that will integrate changes from the recent sessions rework into the `EOSService` approach and introduces changes to that base class that make that transition more straight-forward.

#EOS-2023


Assets/Scripts/AchievementsService.cs
```diff
@@ -348,34 +353,46 @@ public Texture2D GetAchievementLockedIconTexture(string achievementId)
...
                 }
-                else
+
+                if (_downloadCache.TryGetValue(uri, out byte[] iconBytes) && null != iconBytes)
                 {
-                    // TODO: Deal with circumstance where icon bytes are not cached, or are null
+                    if (!textureFromBytes.LoadImage(iconBytes))
+                    {
+                        Debug.LogWarning($"Could not load achievement icon bytes.");
+                        tcs.SetResult(textureFromBytes);
```
WispyMouse: Should this set an exception, since there was some kind of error?

Assets/Scripts/AchievementsService.cs
```diff
@@ -348,34 +353,46 @@ public Texture2D GetAchievementLockedIconTexture(string achievementId)
...
-                    else
+                    GetAndCacheData(uri, data =>
                     {
-                        // TODO: Deal with circumstances where image could not be loaded for some reason
-                    }
-                    break;
+                        if (data.result != UnityWebRequest.Result.Success)
+                        {
+                            Debug.Log($"Could not download achievement icon: {data.result}.");
+                            tcs.SetResult(textureFromBytes);
```
WispyMouse: Should this set an exception, since it resulted in some kind of error?

Assets/Scripts/AchievementsService.cs
```diff
@@ -348,34 +353,46 @@ public Texture2D GetAchievementLockedIconTexture(string achievementId)
...
+
+                if (!_downloadCache.ContainsKey(uri))
                 {
-                    Texture2D textureFromBytes = new(2, 2);
-                    if (textureFromBytes.LoadImage(iconBytes))
-                    {
-                        iconTexture = textureFromBytes;
-                    }
-                    else
+                    GetAndCacheData(uri, data =>
```
WispyMouse, line 373: It feels like we should either exit out of the query now, or use an else to bridge to the next thing.

Currently if the download cache doesn't contain the key, it'll kick off GetAndCacheData. Then it'll see that the data is or isn't in the download cache. Then it'll set the result from bytes, even if that's still the default texture it made earlier.

Assets/Scripts/AchievementsService.cs
```diff
@@ -558,14 +583,27 @@ private async void GetAndCacheData(string uri)
                 await System.Threading.Tasks.Task.Yield();
             }
 
+            asyncOp.completed += operation =>
```
WispyMouse, line 597: Should it set the completed callback before waiting? Will this fire if the asyncOp isDone when you set the .completed?

WispyMouse: ~~Does there need to be any error handling in the situation where the operation finishes in an error, and doesn't have the result available?~~

Edit: Actually, you are handling that! Nevermind, good to keep.

Assets/Scripts/StandardSamples/Services/EOSService.cs
```diff
@@ -23,79 +23,88 @@
...
+        /// ProductUserId for the currently logged in user.
+        /// </param>
         /// <returns>
-        /// True if the user is authenticated using the Connect Interface, false
-        /// otherwise.
+        /// True if the ProductUserId was successfully retrieved and is valid,
+        /// false otherwise.
         /// </returns>
-        private static bool IsLoggedInWithConnectInterface()
+        protected bool TryGetProductUserId(out ProductUserId productUserId)
```
WispyMouse, line 85: I take this will mean a service should utilize this function, rather than fetching from `EOSManager.Instance.GetProductUserId()`?

Assets/Scripts/StandardSamples/UI/Achievements/UIAchievementsMenu.cs
```diff
@@ -201,33 +147,10 @@ private async void OnAchievementDataUpdated()
...
-                        if (Time.frameCount > iconGiveupFrame)
-                        {
-                            UnityEngine.Debug.LogWarning("Timeout : Failed to get icon");
-                            break;
-                        }
-                    }
-                    newButton.SetIconTexture(iconTex);
-                    i += 1;
-                    achievementListItems.Add(newButton);
+                    await AddAchievementButton(achievementData);
```
WispyMouse, line 153: Could this possibly be set up to handle getting all of these at the same time / await all, instead of awaiting one at a time?

Assets/Scripts/StandardSamples/UI/Achievements/UIAchievementsMenu.cs
```diff
@@ -239,6 +162,28 @@ private async void OnAchievementDataUpdated()
...
+
+            achievementListItems.Add(button);
+
+            bool unlocked = achievement.PlayerData.HasValue && achievement.PlayerData.Value.Progress >= 1;
+
+            Task<Texture2D> getIconTextureTask = unlocked
+                ? AchievementsService.Instance.GetAchievementUnlockedIconTexture(achievementId)
+                : AchievementsService.Instance.GetAchievementLockedIconTexture(achievementId);
+
+            var tex = await getIconTextureTask;
```
WispyMouse, line 182: This makes me think we should have a temporary "loading" icon. Perhaps whatever is already set in the prefab, which currently will look like a white square.

Assets/Scripts/StandardSamples/UI/Common/SampleMenu.cs
```diff
@@ -0,0 +1,316 @@
...
+    public abstract class SampleMenuWithFriends : SampleMenu
+    {
+        public enum FriendInteractionState
+        {
+            Hidden,
+            Disabled,
+            Enabled
+        }
+
+        protected SampleMenuWithFriends(bool startsHidden = true) : base(startsHidden) { }
```
WispyMouse: Feels like this should pass in a "true" to the optional parameter of the base class, although that'll be implicitly true anyways.

Assets/Scripts/StandardSamples/UI/Common/SampleMenu.cs
```diff
@@ -0,0 +1,316 @@
...
+        public virtual void Show()
+        {
+            Log($"Show() started");
+
+            // Don't do anything if already showing.
+            if (!Hidden) return;
+
+            if (null == UIParent)
+            {
+                Log($"UIParent for is null for some reason.");
```
WispyMouse: Should this return here?

Assets/Scripts/StandardSamples/UI/Common/SampleMenu.cs
```diff
@@ -0,0 +1,316 @@
...
+        /// Sets the focused control to UIFirstSelected, assuming it is both
+        /// defined and active in the hierarchy.
+        /// </summary>
+        private void SetSelected()
+        {
+            // Controller: Detect if nothing is selected and controller input detected, and set default
+            if (UIFirstSelected == null 
+                || UIFirstSelected.activeSelf == false
+                || UIFirstSelected.activeInHierarchy == false
+                || EventSystem.current == null || EventSystem.current.currentSelectedGameObject != null
```
WispyMouse: Perhaps we could check if the currentSelectedGameObject is not null, but also its GameObject is set to be inactive. Like if there's a menu with a button, and we navigate to this scene while it's still set.

Maybe Unity is smart about this and already manages unsetting the currentSelectedGameObject if it becomes hidden.

Assets/Scripts/StandardSamples/UI/Common/SampleMenu.cs
```diff
@@ -0,0 +1,316 @@
...
+                || !InputUtility.WasGamepadUsedLastFrame())
+            {
+                return;
+            }
+
+            // Controller
+            EventSystem.current.SetSelectedGameObject(UIFirstSelected);
+            Debug.Log("Nothing currently selected, default to UIFirstSelected: EventSystem.current.currentSelectedGameObject = " + EventSystem.current.currentSelectedGameObject);
```
WispyMouse: Can you pass the currentSelectedGameObject as a parameter to Debug.Log? It'll make it so that if someone clicks on the log, it'll select that game object in the hierarchy.

WispyMouse: Either that, and/or it should be using the below Log function.

Assets/Scripts/StandardSamples/UI/Common/SampleMenu.cs
```diff
@@ -0,0 +1,316 @@
...
+        /// </summary>
+        /// <param name="startsHidden">
+        /// Indicates whether the SampleMenu should be visible or hidden when it
+        /// launches. Most of them should be hidden.
+        /// </param>
+        /// <param name="requiresAuthentication">
+        /// Indicates whether the SampleMenu requires a user be authenticated in
+        /// order for it to be used. Most of them require authentication.
+        /// </param>
+        protected SampleMenu(bool startsHidden = true, bool requiresAuthentication = true)
```
WispyMouse, line 151: MonoBehaviours cannot have constructors. I think?

Assets/Scripts/StandardSamples/UI/Peer2Peer/UIPeer2PeerMenu.cs
```diff
@@ -328,13 +314,11 @@ public void ShowMenu()
 
         }
 
-        public void HideMenu()
+        public override void Hide()
```
WispyMouse, line 317: Consider adding base.Hide()

Assets/Scripts/StandardSamples/UI/Storage/Player/UIPlayerDataStorageMenu.cs
```diff
@@ -71,21 +60,17 @@ public class UIPlayerDataStorageMenu : MonoBehaviour, ISampleSceneUI
...
         private EOSPlayerDataStorageManager PlayerDataStorageManager;
         private PlayerDataInventory currentInventory = null;
 
-        private HashSet<string> fileNames;
-        private List<UIFileNameEntry> fileNameUIEntries;
+        private HashSet<string> fileNames = new();
+        private List<UIFileNameEntry> fileNameUIEntries = new();
 
-        private void Awake()
+        protected override void OnEnable()
```
WispyMouse, line 78: I'm unsure we'll run in to this scenario, but this behaviour is different. I believe `Awake` can only run once, but `OnEnable` will run every time it becomes enabled. If someone shows, then hides, then shows this, it will have added `fileNameUIEntries` twice with the same data.

WispyMouse:
Left some food for thought. This is in draft, will do a more detailed review when it is out of draft.

Assets/Scripts/StandardSamples/UI/Achievements/UIAchievementsMenu.cs
paulhazen: The reason it does them one at a time is because the downloading of the image bytes happens one at a time. That is - we don't want to implement it in such a way that it requires all the images to be loaded. This implementation lets them load asynchronously.

Assets/Scripts/StandardSamples/Services/EOSService.cs
paulhazen: I added this because in the process of refactoring things, I had more than one class that derived from `EOSService` doing the same thing. It's up to the individual developer to decide whether or not to utilize this, but this helper function seemed like a clean way to simplify things without changing the API (since changes to `EOSManager` constitute changing the API).

Assets/Scripts/StandardSamples/UI/Achievements/UIAchievementsMenu.cs
paulhazen: The way our current implementation works is that it has an empty white square as you mentioned. A "loading" icon couldn't hurt, but is outside the scope of the PR.

Assets/Scripts/StandardSamples/UI/Common/SampleMenu.cs
paulhazen: Good catch! Since changing it to be `true` is the same as not defining the constructor at all, I just removed it altogether. Makes a lot more sense.

Assets/Scripts/StandardSamples/UI/Common/SampleMenu.cs
paulhazen: While this is true (and something I just learned) the reason that this _does_ work is because it sets what are effectively readonly parameters. Unity will not honor or support `MonoBehaviour` implementing classes having a constructor with parameters, since `SampleMenu` is abstract, and implementing classes all set the variables themselves, the result is that the type constructed by Unity will be parameterless, and therefore _will_ be executed.

Assets/Scripts/StandardSamples/UI/Common/SampleMenu.cs
paulhazen: Nope. The `UIParent` parameter is something that is optionally set by a user from within Unity Editor's inspector window. It's not guaranteed that the `UIFirstSelected` field member is set, so even if it _is_ null, the show method should still continue to set the `gameObject` to be active.

Assets/Scripts/StandardSamples/UI/Common/SampleMenu.cs
paulhazen: I'm fairly certain that Unity is smart enough about this. Meaning that if a gameObject is set as inactive, then Unity knows to remove it as the `currentSelectedGameObject`. Of course, this is an assumption on my part. 

While this function appears to be new, it is actually transplanted as it existed in one of the other `Menu` classes. Most of the existing `Menu` classes implement a function similar to this one, but this implementation is a reflection of the implementation that I found to be most complete.

Assets/Scripts/StandardSamples/UI/Common/SampleMenu.cs
paulhazen: Yep, good catch. Fixed.

Assets/Scripts/StandardSamples/UI/Peer2Peer/UIPeer2PeerMenu.cs
paulhazen: Yup. Ficksed!

paulhazen: Yup. Ficksed!

Assets/Scripts/StandardSamples/UI/Storage/Player/UIPlayerDataStorageMenu.cs
paulhazen: You are correct. Not certain how this changed to `OnEnable`

paulhazen: Ficksed!

Assets/Scripts/AchievementsService.cs
paulhazen: Exceptions should typically be utilized in circumstances where execution cannot continue. In this case, execution can continue, because it is handled gracefully and will not cause a crash.

Assets/Scripts/AchievementsService.cs
paulhazen: Not all errors / warnings need to throw an exception. As the name implies - exceptions should be used in narrow circumstances (typically in circumstances where the execution of the program cannot continue). In this particular case, the `AchievementService` can continue to function properly even if downloading an individual icon texture fails. 


For posterity and to start conversations in places other than this PR, I have the following additional thoughts:

I admit there is some subjective determination here, and I am not certain I can specifically state where I would draw the line in all circumstances between when to:

- Log a warning.
- Log an error.
- Throw an exception.

Additionally, there is a question about when a log should be made. Should it be logged before the exception is thrown? Should it be thrown by the code that _catches_ the exception? Should different policies apply to warnings vs errors in that respect? These are all questions that are worth answering and considering - questions I don't have definitive answers for - questions for which we have not defined a standard practice within the team.

Assets/Scripts/StandardSamples/UI/CustomInvites/UICustomInvitesMenu.cs
```diff
@@ -40,22 +38,22 @@ public class UICustomInvitesMenu : UIFriendInteractionSource, ISampleSceneUI
 
         private EOSCustomInvitesManager CustomInvitesManager;
         private EOSFriendsManager FriendsManager;
-        private List<UICustomInviteEntry> PendingInviteEntries;
+        private List<UICustomInviteEntry> PendingInviteEntries = new();
         private bool PayloadSet = false;
 
-        private void Awake()
+        protected override void OnEnable()
```
paulhazen: This needs to go back to being `Awake()` because it should only happen once.

Assets/Scripts/AchievementsService.cs
WispyMouse: How does the caller know that this failed, and to not use the texture?

Assets/Scripts/AchievementsService.cs
WispyMouse: How does the caller know that this failed, and to not use the texture? (same question as above)

Assets/Scripts/AchievementsService.cs
WispyMouse: Only meant to resolve the conversation that was crossed out, still consider first comment in this chain

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -129,6 +129,10 @@ public partial class EOSManager : MonoBehaviour, IEOSCoroutineOwner
         /// <value>Stores instances of feature managers</value>
         private static Dictionary<Type, IEOSSubManager> s_subManagers = new Dictionary<Type, IEOSSubManager>();
 
+        private static event OnConnectLoginCallback OnConnectLogin;
+        private static event OnAuthLoginCallback OnAuthLogin;
+        private static event OnAuthLogoutCallback OnAuthLogout;
```
paulhazen: Wasn't this introduced in another PR?

Assets/Scripts/AchievementsService.cs
paulhazen: Through conversations during peer review - we should change this so that when the image fails to load, a `null` texture is returned and a warning is logged.

Assets/Scripts/AchievementsService.cs
paulhazen: Same approach as the earlier comment. Marking this conversation as resolved since it's a duplicate-ish thing.

Assets/Scripts/AchievementsService.cs
```diff
@@ -558,14 +583,27 @@ private async void GetAndCacheData(string uri)
                 await System.Threading.Tasks.Task.Yield();
             }
```
paulhazen: This while loop should be removed, as the handling of the `completed` event obviates the need for it.

Assets/Scripts/AchievementsService.cs
paulhazen: Resolved in the form of a comment that precedes this one. Consequently, marking conversation resolved.

Assets/Scripts/AchievementsService.cs
paulhazen: Made changes consistent with aforementioned determination, therefore marking this conversation as resolved.

Assets/Scripts/StandardSamples/UI/Achievements/UIAchievementsMenu.cs
WispyMouse: Added #EOS-2024 for this.

Assets/Scripts/AchievementsService.cs
paulhazen: Excellent point! I have updated the implementation and now I believe things are handled more appropriately.

Assets/Scripts/StandardSamples/Services/AchievementsService.cs
```diff
@@ -348,34 +353,62 @@ public Texture2D GetAchievementLockedIconTexture(string achievementId)
...
+                    TaskCompletionSource<byte[]> downloadTcs = new();
+
+                    GetAndCacheData(uri, data =>
                     {
-                        iconTexture = textureFromBytes;
-                    }
-                    else
+                        if (data.result == UnityWebRequest.Result.Success)
+                        {
+                            downloadTcs.SetResult(data.data);
```
WispyMouse, line 377: Does the task completion source end when it has its data set? Or will the below code block of logging a warning and then setting the result still run?

Assets/Scripts/StandardSamples/UI/Common/SampleMenu.cs
WispyMouse: Study shows that an object being inactive does not unset itself. We discussed checking for its activeness if it is set.

andrew-hirata-playeveryware (approved):

Assets/Scripts/StandardSamples/Services/AchievementsService.cs
paulhazen: SetResult does not end the task completion source.

WispyMouse (approved):

PR #835, opened 2024-08-13
Change `EOSManager` to use events instead of lists of delegate instances.

paulhazen:
This PR introduces a change to `EOSManager` that implements the authentication callbacks by using `events`. 

There is no change in the API for EOSManager - the only changes made are to internal implementation of the handling of delegates.

### DIFF
--- a/com.playeveryware.eos/Runtime/Core/EOSManager.cs
+++ b/com.playeveryware.eos/Runtime/Core/EOSManager.cs
@@ -110,6 +110,10 @@ namespace PlayEveryWare.EpicOnlineServices
 
         public delegate void OnConnectLoginCallback(Epic.OnlineServices.Connect.LoginCallbackInfo loginCallbackInfo);
 
+        protected static event OnAuthLoginCallback OnAuthLogin;
+        protected static event OnAuthLogoutCallback OnAuthLogout;
+        protected static event OnConnectLoginCallback OnConnectLogin;
+
         public delegate void OnCreateConnectUserCallback(CreateUserCallbackInfo createUserCallbackInfo);
 
         public delegate void OnConnectLinkExternalAccountCallback(LinkAccountCallbackInfo linkAccountCallbackInfo);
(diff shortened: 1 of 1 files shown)


com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -378,19 +373,19 @@ public T GetOrCreateManager<T>() where T : IEOSSubManager, new()
                     manager = new T();
                     s_subManagers.Add(type, manager);
 
-                    if (manager is IEOSOnConnectLogin)
+                    if (manager is IEOSOnConnectLogin connectLogin)
```
arthur740212, line 376: Is this syntax a new feature?

arthur740212 (commented):
It looks like the change should do the same thing.
Wondering what the intention of the old version was for, and if the new version supports that.

Another question, does OnAuthLogin?.Invoke only call the first callback or all added listeners?

paulhazen:
> It looks like the change should do the same thing. Wondering what the intention of the old version was for, and if the new version supports that.
> 
> Another question, does OnAuthLogin?.Invoke only call the first callback or all added listeners?

It calls all of them - that's the intent behind C#'s `event` keyword. It basically does what we were doing manually, but it does it faster and better.

com.playeveryware.eos/Runtime/Core/EOSManager.cs
paulhazen: Not terribly new - been there for several years if I'm not mistaken.

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -378,19 +373,19 @@ public T GetOrCreateManager<T>() where T : IEOSSubManager, new()
...
                     {
-                        AddConnectLoginListener(manager as IEOSOnConnectLogin);
+                        OnConnectLogin += connectLogin.OnConnectLogin;
```
WispyMouse, line 378: nit: Would it make any sense to call `AddConnectLoginListener` here instead of adding directly to the event? Maybe in case in the future there's some extra code there. Probably fine as is.

WispyMouse:
EOSManager also has `sZ_onApplicationShutdownCallbacks`. Could we eventify this?

WispyMouse (approved):
I left a couple of possible suggestions. This is a clean and nice change, and I'll try to make use of events in the same way later. For some reason I had always thought this required some sophisticated EventListener, or to use UnityEvent, but here you show how easy it is to use events.

com.playeveryware.eos/Runtime/Core/EOSManager.cs
paulhazen: Not terribly - there isn't anything else that we do with it other than to add it to a list of delegates. This basically replaces our home-grown implementation with a system that does what we were doing much more smoothly.

paulhazen:
> EOSManager also has `sZ_onApplicationShutdownCallbacks`. Could we eventify this?

That's less straightforward to do, because there is a function named what I would change the name of the event to. Doing so seemed like it was out of scope for the time being - but since all these events pertain to authentication it felt appropriate to group them.

arthur740212 (approved):

PR #836, opened 2024-08-13
refactor(playerstorage): Starting on PlayerDataStorageService

WispyMouse:
This is the first part of the PlayerStorage rework, which is service-ifiying the PlayerDataStorageManager.

Refactor and rework to the actual sample UX will come next. For now, please review this approach to implementing services.

Note that this does not merge in to development, it merges in to a merge branch. There were multiple commits, but I squashed them to help excise an unintended file change.


Assets/Scripts/EOSTransferInProgress.cs
```diff
@@ -21,6 +21,7 @@
 */
 
 using System.Collections.Generic;
+using System.Threading.Tasks;
```
andrew-hirata-playeveryware, line 24: Why was this added?

andrew-hirata-playeveryware (approved):

Assets/Scripts/EOSPlayerDataStorageTransferTask.cs
```diff
@@ -0,0 +1,53 @@
...
+        }
+
+        public EOSPlayerDataStorageTransferTask(TaskCompletionSource<byte[]> innerTask, byte[] data) : base(innerTask, data)
+        {
+        }
+
+        public EOSPlayerDataStorageTransferTask(Result resultCode, Exception failureException = null) : base(resultCode, failureException)
+        {
+        }
+    }
```
paulhazen, line 52: I don't think it makes sense to add this class - particularly because it does not add functionality. There is no way for instance that `TransferRequest` can be accessed anywhere but externally, and having these constructors just pass through to the base task seems like an unnecessary step.

Assets/Scripts/EOSTransferTask.cs
```diff
@@ -0,0 +1,305 @@
...
+        /// When an EOSTransferTask is made, it either is an immediate failure or an inner task should be set.
+        /// This keeps a reference to the entire task completion source so that it can complete it when done.
+        /// </summary>
+        public TaskCompletionSource<byte[]> InnerTaskCompletionSource;
+
+        /// <summary>
+        /// Delegate signature for <see cref="OnProgressUpdated"/>.
+        /// </summary>
+        /// <param name="progress">The current percentage completion of the task, by number of bytes to work on.</param>
+        public delegate void OnProgressUpdatedDelegate(float progress);
```
paulhazen, line 56: Convention would dictate that this be named `ProgressUpdatedEventHandler`

Assets/Scripts/EOSTransferTask.cs
```diff
@@ -0,0 +1,305 @@
...
+        /// If this task is a download task, this is where the data will be put when the task is complete.
+        /// If this is an upload task, this is where the data should be put.
+        /// </summary>
+        public byte[] Data { get; protected set; } = new byte[] { };
+
+        /// <summary>
+        /// If this task has run to completion, or encountered an error, this ResultCode should be set.
+        /// You can determine that an EOSTransferTask is complete if it has any result code.
+        /// </summary>
+        public Result? ResultCode;
```
paulhazen, line 73: You should explain in your comments why you are using a `nullable` value, and what it means when this field member is `null`.

Assets/Scripts/EOSTransferTask.cs
```diff
@@ -0,0 +1,305 @@
...
+
+        /// <summary>
+        /// Inner backing for the total size of the transfer.
+        /// </summary>
+        private uint transferSize = 0;
+
+        /// <summary>
+        /// Indexes the current byte count that has been processed.
+        /// </summary>
+        private uint currentIndex = 0;
```
paulhazen, line 89: Non-public field members should be prefixed with `_`.

Assets/Scripts/EOSTransferTask.cs
```diff
@@ -0,0 +1,305 @@
...
+        /// Constructor for upload scenarios.
+        /// </summary>
+        /// <param name="innerTask">A task completion source that can be used to indicate when the task is completed.</param>
+        /// <param name="data">The data to upload, in byte[] form.</param>
+        public EOSTransferTask(TaskCompletionSource<byte[]> innerTask, byte[] data)
+        {
+            InnerTaskCompletionSource = innerTask;
+            Data = data;
+            TotalSize = (uint)data.Length;
+        }
```
paulhazen, line 131: Since the intent behind `EOSTransferTask` is to have it be agnostic to whether the operation is an upload or a download, the difference in implementation between those two things should be reflected in the deriving classes, not in the base implementation. 

`EOSTransferTask` should contain functionality that is common to both scenarios and exclude any implementation that is unique to any.

Assets/Scripts/EOSTransferTask.cs
```diff
@@ -0,0 +1,305 @@
...
+
+            if (failureException != null)
+            {
+                InnerTaskCompletionSource.SetException(failureException);
+            }
+            else
+            {
+                InnerTaskCompletionSource.SetCanceled();
+            }
+        }
```
paulhazen, line 151: It is unclear to me when this would be utilized - it seems like if you know that the failure is immediate there is no need to generate the task. Is there additional context that I am missing?

Assets/Scripts/EOSTransferTask.cs
```diff
@@ -0,0 +1,305 @@
...
+
+        /// <summary>
+        /// Indicates if the operation is complete and ready to read the results out of.
+        /// </summary>
+        /// <returns>True if the operatoin is complete.</returns>
+        public bool IsDone()
```
paulhazen, line 157: +1 on renaming this, it's much better than before when it was `Done()`.

Assets/Scripts/EOSTransferTask.cs
```diff
@@ -0,0 +1,305 @@
...
+            return ResultCode.HasValue || (TotalSize > 0 && TotalSize == currentIndex);
+        }
+
+        /// <summary>
+        /// Internal processor for incoming file work.
+        /// This should be assigned to <see cref="ReadFileOptions.ReadFileDataCallback"/>.
+        /// </summary>
+        /// <param name="data">Data from the EOS SDK about the incoming chunk of data.</param>
+        /// <returns>A command to either stop or continue reading the incoming data.</returns>
+        internal ReadResult OnFileDataReceived(ref ReadFileDataCallbackInfo data)
```
paulhazen, line 168: Why is this marked as `internal`?

Assets/Scripts/EOSTransferTask.cs
```diff
@@ -0,0 +1,305 @@
...
+            if (data.ResultCode != Result.Success)
+            {
+                ResultCode = data.ResultCode;
+                InnerTaskCompletionSource.SetException(new Exception($"{nameof(EOSTransferTask)} ({nameof(OnFileSent)}): Could not send data. Result code {data.ResultCode}"));
+                return;
+            }
+
+            ResultCode = Result.Success;
+            InnerTaskCompletionSource.SetResult(Data);
+        }
```
paulhazen, line 291: I think these function names need to be improved. It's not clear to me what the difference is between `OnFileSent` and `OnFileDataSent`, since (without looking at their respective implementations) they _mean_ the same thing. We should strive to have our method names be clearer, and not require a developer to look at the implementation (or even read the comments) to understand the distinction between the two.

Assets/Scripts/StandardSamples/Services/PlayerDataStorageService.cs
```diff
@@ -0,0 +1,468 @@
...
+    /// </summary>
+    public class PlayerDataStorageService : EOSService
+    {
+        private const uint MAX_CHUNK_SIZE = 4096;
+
+        /// <summary>
+        /// The EOS SDK can only handle one incoming or outgoing request at a time for several functions.
+        /// This semaphore can be used to gate calls, waiting until they're ready to be processed.
+        /// </summary>
+        private SemaphoreSlim _playerDataStorageSemaphore = new SemaphoreSlim(1);
```
paulhazen, line 53: If _the EOS SDK can only handle one incoming or outgoing request at a time_, then this semaphore needs to be more widely accessible to other parts of the code. An example that comes immediately to mind is that you cannot have a TitleStorage request active at the same time as a PlayerStorage request - so _at least those two_ need a way to share this semaphore.

Assets/Scripts/StandardSamples/Services/PlayerDataStorageService.cs
```diff
@@ -0,0 +1,468 @@
...
+                                tcs.SetException(new Exception($"{nameof(PlayerDataStorageService)} ({nameof(DownloadFile)}): CopyFileMetadataAtIndex returned error. Result code {result}"));
+                                return;
+                            }
+                            else if (fileMetadata != null)
+                            {
+                                if (!string.IsNullOrEmpty(fileMetadata?.Filename))
+                                {
+                                    fileNames.Add(fileMetadata?.Filename);
+                                }
+                            }
```
paulhazen, line 174: For the sake of clarity, I would excise this code and put it into a `private` method that can be designed statically that is _just responsible_ for doing the copying of file metadata.

Assets/Scripts/StandardSamples/Services/PlayerDataStorageService.cs
```diff
@@ -0,0 +1,468 @@
...
+        /// <returns>An awaitable task containing either the result or an exception.</returns>
+        public async Task<byte[]> DownloadFileAsync(ProductUserId userId, string fileName)
+        {
+            EOSPlayerDataStorageTransferTask transferTask = DownloadFile(userId, fileName);
+
+            await transferTask.InnerTaskCompletionSource.Task;
+
+            if (transferTask.InnerTaskCompletionSource.Task.IsFaulted)
+            {
+                return await Task<byte[]>.FromException<byte[]>(new AggregateException($"{nameof(PlayerDataStorageService)} ({nameof(DownloadFileAsync)}): Failed to download file. Result code {transferTask.ResultCode}", transferTask.InnerTaskCompletionSource.Task.Exception));
```
paulhazen, line 206: Why are you generating a new task here? If `IsFaulted` is true, that should be just as "catchable" as creating a new one. It's not clear to me what is being gained by this approach.

Assets/Scripts/StandardSamples/Services/PlayerDataStorageService.cs
```diff
@@ -0,0 +1,468 @@
...
+
+                    // If no valid handle is returned, then error out
+                    if (downloadTask.TransferRequest == null)
+                    {
+                        downloadTask.ResultCode = Result.UnexpectedError;
+                        tcs.SetException(new Exception($"{nameof(PlayerDataStorageService)} ({nameof(DownloadFile)}): Bad handle returned for reading file, can not download file."));
+                        _playerDataStorageSemaphore.Release();
+                        return;
+                    }
+                });
```
paulhazen, line 297: This code is doing three discrete things: 

1. Querying the file
2. Copying the file meta data
3. Reading the file

I think that these three tasks should be split into individual methods. This will make the code a lot easier to read and consequently to debug.

Assets/Scripts/StandardSamples/Services/PlayerDataStorageService.cs
```diff
@@ -0,0 +1,468 @@
...
+        /// </summary>
+        /// <param name="fileName">The file name to delete.</param>
+        /// <returns>An awaitable task. Contains an exception if an error occurs.</returns>
+        public Task DeleteFile(string fileName)
+        {
+            TaskCompletionSource<Result> tcs = new();
+
+            ProductUserId localProductUserId = EOSManager.Instance.GetProductUserId();
+
+            if (localProductUserId == null || !localProductUserId.IsValid())
```
paulhazen, line 394: Another good use-case for null propagation.

Assets/Scripts/StandardSamples/UI/Storage/Player/UIPlayerDataStorageMenu.cs
```diff
@@ -75,8 +75,8 @@ public class UIPlayerDataStorageMenu : MonoBehaviour, ISampleSceneUI
         public GameObject UIFirstSelected;
 
         private string currentSelectedFile = string.Empty;
+        private Dictionary<string, string> localFileNameToContents = new Dictionary<string, string>();
```
paulhazen, line 78: If I'm not mistaken, this data already exists within the Service class. Is there a reason to keep an independent copy of that data outside of it? Seems like it opens the possibility for there to be an inconsistency between the two.

Assets/Tests/PlayMode/PlayerDataStorageServiceTests.cs
```diff
@@ -0,0 +1,259 @@
...
+    using System;
+    using UnityEngine.Events;
+    using System.Text;
+    using System.Collections.Generic;
+    using System.Text.RegularExpressions;
+    using System.Threading.Tasks;
+
+    public class PlayerDataStorageServiceTests : EOSTestBase
+    {
+        static int[] NumberOfFilesToMakeFor_UploadedFiles_CanBeQueried = { 1, 3, 10 };
```
paulhazen, line 42: Commenting would be good for this - it's not clear what this field member does or what meaning its value has.

Assets/Tests/PlayMode/PlayerDataStorageServiceTests.cs
```diff
@@ -0,0 +1,259 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Assuming there are no files in storage for the test user, querying for
+        /// files should result in a NotFound error.
+        /// 
+        /// TODO: The Services paradigm doesn't allow for returning of messages.
+        /// So the expected behavior is no exception thrown, but do return an empty list.
+        /// </summary>
```
paulhazen, line 95: Can you clarify what you mean by this exactly?

paulhazen (changes requested):

WispyMouse:
This has been closed because it was implemented in another way in another PR. The tests for it have been put in to a branch, that it is not at this time yet in PR.

PR #839, opened 2024-08-14
Create check-for-package-difference.yml

WispyMouse:
#EOS-1718


arthur740212 (approved):

arthur740212:
Looks like it is working. It gives an error about the package version not being changed.
And indeed it wasn't :D
Though is there a way to only enable this check on stable?

WispyMouse:
@arthur740212 It only checks stable. It just so happens that this is trying to merge into stable.

andrew-hirata-playeveryware (approved):

PR #843, opened 2024-08-16
Remove com.playeveryware.eos.asmdef as it is not used

EllieTellie:
Hi,

When  you open the Unity project you can see this warning:
`Assembly for Assembly Definition File 'Assets/Plugins/com.playeveryware.eos.asmdef' will not be compiled, because it has no scripts associated with it.`

Because the asmdef does not have any c# scripts associated it is essentially irrelevant and just displays warnings when scripts get recompiled.
This should mean it is safe to delete unless of course you intend to add scripts to this .asmdef in the future.

For reference all the c# files in the Directory are referencing com.playeveryware.eos.core.asmdef instead.

Let me know what you think.


andrew-hirata-playeveryware:
Hello @EllieTellie,

Thank you for the PR. This is kind of a 'funny' issue that our plugin has had in the past, where due to some code structure refactors we accidentally but slowly moved all the code out of the main assembly, i.e.  com.playeveryware.eos. We can't delete that one because it's supposed to be the fixed one that one may use to reference our project as a dependency, so the resolution to that warning is to move at least one file back into the com.playeveryware.eos assembly. 

That being said, this was believed to be resolved in version `3.3.0` of the plugin. Which version of the plugin, and which version of Unity, are you using when you saw this issue?

EllieTellie:
> Hello @EllieTellie,
> 
> Thank you for the PR. This is kind of a 'funny' issue that our plugin has had in the past, where due to some code structure refactors we accidentally but slowly moved all the code out of the main assembly, i.e. com.playeveryware.eos. We can't delete that one because it's supposed to be the fixed one that one may use to reference our project as a dependency, so the resolution to that warning is to move at least one file back into the com.playeveryware.eos assembly.
> 
> That being said, this was believed to be resolved in version `3.3.0` of the plugin. Which version of the plugin, and which version of Unity, are you using when you saw this issue?

Just check out 'development' and open the console with Warnings enabled and you should be able to see that message. I think that is version 3.3.1

andrew-hirata-playeveryware:
I have made a [new PR](https://github.com/PlayEveryWare/eos_plugin_for_unity/pull/860) that adds a dummy file that resolves this issue till we can do the actual solution of either: 

* re-split the files shared by editor and runtime into the core asm and the 'main' asm
* move everything into the main asm

PR #844, opened 2024-08-16
Ignore .gitattributes files when exporting package files.

EllieTellie:
Simple fix to ignore .gitattributes files when exporting plugin.

Ideally the package does not contain .gitattributes because there is no guarantee whoever imports this package is even using git or may not have git lfs.

I was going to add to this that the package json files did have support for a blacklist, but that field seemed never to be referenced. I suppose that might be a cleaner fix for blacklisted files, but sometimes simple might be best.

### DIFF
--- a/Assets/Plugins/Source/Editor/Utility/PackageFileUtility.cs
+++ b/Assets/Plugins/Source/Editor/Utility/PackageFileUtility.cs
@@ -209,7 +209,11 @@ namespace PlayEveryWare.EpicOnlineServices.Utility
 
                 FileInfo destInfo = new(destPath);
 
-                if (file.originalSrcDestPair.dest.Contains('~') && !file.originalSrcDestPair.dest.Contains("Samples~"))
+                if (src.Name == ".gitattributes") // Ignore .gitattributes as these should not be in the package
+                {
+                    continue;
+                }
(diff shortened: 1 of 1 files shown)


WispyMouse:
Ahoy @EllieTellie ! Thank you for opening this pull request. I wanted to talk about this suggestion for a bit.

The .gitattributes being exported as a package helps with one of our user scenarios, which is people pulling down the UPM Package that [we release in this other repository](https://github.com/PlayEveryWare/eos_plugin_for_unity_upm). If this PR were to be used exactly as-is, the UPM repository's release would have some problems with large binary files and unnecessary checked in items.

Is the presence of the .gitattributes causing any problems in your environment? Or is this a philosophical sort of fix, for a cleaner repository?

EllieTellie:
I just felt it is uncommon for packages to distribute these in packages as git lfs is a user choice, but if it is intended just leave it as-is.

EllieTellie:
I suppose it could be limited to just .tgz export but I'm not too fuzzed about it. Just wanted to contribute.

paulhazen:
Marking this PR as closed, since it has been discussed as something that we will not be making a change to. 

Many thanks to @EllieTellie for their enthusiastic interest in making a contribution to the project, and for the insights they brought.

PR #856, opened 2024-08-20
Update: EOSLobbyManager, EOSTransport, EOSEACLobbyManager: Enhanced functionality of existing Callbacks, Added additional Voice Channel related functions and changed accessibility of Lookup functions related to Client Identities

Xeynt:
Class: EOSLobbyManager.cs
Affected Class: EOSEACLobbyManager.cs
Add new LobbyChangedEvents params to OnLobbyChanged Callbacks so all Clients will now be capable of identifying what triggered a LobbyChanged callback.

Added new MuteLocalMember and DeafenLocalMember functions which do not rely on CurrentLobby.Members List. The existing ToggleMute function relies on the CurrentLobby.Members list to be populated and will not work before that update is received. MuteLocalMember and DeafenLocalMember can be execute as soon as the Lobby is created.

Added new MuteMember function that will mute a Target ProductUserId. It is similar to the existing ToggleMute function just that this one isn't a toggle and will set Mute status based of the specified value.
____________________________________________________________________________________________________________
Class: EOSTransport.cs
Modified Accessibility (Changed to Public) of the following functions: 
public ProductUserId GetUserId(ulong clientId)
public ulong GetClientId(ProductUserId userId)
The above 2 functions are useful for other custom classes to build upon, especially for Host Migration where we need to be able to lookup the corresponding ProductUserID and ClientID.


Xeynt:
Resubmitted this PR here, squashed my changes into one commit: https://github.com/PlayEveryWare/eos_plugin_for_unity/pull/862

arthur740212:
Hello. BlockParticipants is in the RTCInterface, which differs from the RTCAudio interface.
We do not have an example that uses BlockParticipants in the sample code, but you should be able to use it directly from RTCInterface.
Hope this helps with your project :)

Xeynt:
> Hello. BlockParticipants is in the RTCInterface, which differs from the RTCAudio interface. We do not have an example that uses BlockParticipants in the sample code, but you should be able to use it directly from RTCInterface. Hope this helps with your project :)

Yes that indeed helped! I've found it and am proceeding to implement it. Thank you.

One question I've is, is it recommended to use the EOSLobbyManager or to create my own? Since EOSLobbyManager is classified by your team as Samples, I guess there won't be any benefits to be directly using it as a means to communicate with my own LobbyManager? By benefits, I mean if this plugin receives more new features and updates, I can directly use them since I'm using the EOSLobbyManager. 

Additionally if I create my own EOSLobbyManager, I would still have to do it for all the other classes which references EOSLobbyManager, which can become a pain.

Though in this case, I would have to push in the BlockParticipant code into the development branch if I don't want to have to manually discard changes and re-apply them every time I pull for updates or download the plugin from Tar ball which I'm currently doing.

PR #861, opened 2024-08-22
Integrate 1.16.3 35276460

arthur740212:
Includes :
EOS SDK updated to 1.16.3 CL 35276460
Fixes to pew modified SDK source.
Updated SDK import tool

#EOS-2036


com.playeveryware.eos/Runtime/EOS_SDK/Generated/ContinuanceToken.cs
```diff
@@ -50,9 +50,10 @@ public Result ToString(out Utf8String outBuffer)
 			// Call the EOS_ContinuanceToken_ToString function will a null buffer address to determine the correct buffer length.
 			// Discard the result of this call because it is not relevant. 
 			_ = Bindings.EOS_ContinuanceToken_ToString(InnerHandle, System.IntPtr.Zero, ref inOutBufferLength);
-			// PEW: End modify
-			
-			var funcResult = Bindings.EOS_ContinuanceToken_ToString(InnerHandle, outBufferAddress, ref inOutBufferLength);
+            outBufferAddress = Helper.AddAllocation(inOutBufferLength);
```
paulhazen: Since this line is new - can you clarify why it was added? It seems identical to line 39, and it's not clear to me whether this was something that _should have_ been in the previous version or not.

Assets/Plugins/Source/Editor/EditorWindows/InstallEOSZipWindow.cs
```diff
@@ -249,7 +249,7 @@ protected override void RenderWindow()
                             // This should be the correct directory
                             var projectDir = FileUtility.GetProjectPath();
                             // TODO: Async not tested here.
-                            PackageFileUtility.CopyFilesToDirectory(projectDir, fileResults).Wait();
+                            PackageFileUtility.CopyFilesToDirectory(projectDir, fileResults);
```
paulhazen: Changing this line causes a warning to be generated in the Unity Editor. This can be fixed by either discarding the return value like this:
```csharp
_ = PackageFileUtility.CopyFilesToDirectory(projectDir, fileResults);
```
or by adding `async` to the function signature, and changing this line to this:
```csharp
await PackageFileUtility.CopyFilesToDirectory(projectDir, fileResults);
```

paulhazen (changes requested):

etc/PlatformSpecificAssets/EOS/Android/static-stdc++/libs.meta
paulhazen, line 1: Leaving this comment here - but it pertains to all the files that seem to have been deleted from the `libs` directory. Are we certain that the files in that directory can be deleted?

andrew-hirata-playeveryware: Yes: the shared object files (`.so`) are also in the `.aar` files, so having them committed separately is redundant.

com.playeveryware.eos/Runtime/EOS_SDK/Generated/ContinuanceToken.cs
andrew-hirata-playeveryware: It should have been in the previous version. But more specifically, line 39 should have been moved down to where this line is at now.

andrew-hirata-playeveryware (changes requested):

com.playeveryware.eos/Runtime/EOS_SDK/Generated/ContinuanceToken.cs
arthur740212: My thought is that since it is our edit to the SDK code, I would like to keep them in a block, and don't touch EOSSDK code if possible, it will be slightly easier to look at whenever an update happens

paulhazen: This is true - but there is an earlier `Helper.AddAllocation` call, on line 39 which needs to be removed if this one is being kept.

andrew-hirata-playeveryware (approved):

Assets/Plugins/Source/Editor/EditorWindows/InstallEOSZipWindow.cs
paulhazen: This has been fixed, so I am marking the conversation as resolved.

paulhazen (approved):

Issue #859, opened 2024-08-21
EOSTransport ClientID does not match NGO ClientID


Ermelious:
**Describe the bug**
Because EOSTransport ClientID does not match NGO ClientID especially the ones stored in the dictionary, calling some NGO functions such as NetworkManager.Singleton.ConnectedClients[clientId] does not work when using the EOSTransport ClientToUserID lookup dictionary to get the ID from the member's ProductUserID. This ClientID mismatches when the same Lobby Member re-enters the NGO server the 2nd time. 

**To Reproduce**
1. Set the following in any network script (Monitor the outcome when the same Lobby Member re-enters the Server multiple times. The ClientID in EOSTransport will no longer match the NGO NetworkClient's ID):
```c#
        NetworkManager.Singleton.OnConnectionEvent += (NetworkManager networkManager, ConnectionEventData connectionEventData) => 
        {
            bool isLocalClient = connectionEventData.ClientId == NetworkManager.Singleton.LocalClientId;
            NetworkManager.Singleton.ConnectedClients.ContainsKey(eosTransport.GetClientId(productUserId))
        };
```

**Expected behavior**
Is the following possible? The ClientID should match or use the NGO's ClientID instead of having EOSTransport generate its own ClientID.
This seems to be adding unnecessary complexity to build upon as both EOSTransport and NGO will have 2 separate ClientIDs. 
Though perhaps I've missed out the function in this SDK (perhaps there is an existing way to lookup ProductUserID to the NGO's ClientID? Kindly advice.

I've tried looking up EOSTransport and EOSTransportManager, can't seem to find a list where it stores all connections with their respective NGO NetworkClients that matches the ClientIDs and ProductUserIDs.

**Screenshots**
Just assumptions here, this is the portion of the EOSTransport.cs code that is assigning a separate client ID entirely disregarding the NGO Client ID. Is it possible to change it so that EOSTransport just takes the ClientID assigned by NGO or perhaps change NGO's client id so that it matches the one assigned by EOSTransport?
![image](https://github.com/user-attachments/assets/6de5fb28-de29-4ba6-a4e7-3f66da1223c8)

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Editor
 - Unity Version: 6000
 - Plugin Version: Latest

andrew-hirata-playeveryware:
Hello @Ermelious 
It might be possible to unify the numbers, but fundamentally they're not the same concept.

[...]

PR #862, opened 2024-08-23
Update: EOSLobbyManager, EOSTransport, EOSEACLobbyManager: Enhanced functionality of existing Callbacks, Added additional Voice Channel related functions and changed accessibility of Lookup functions related to Client Identities

Xeynt:
Class: EOSLobbyManager.cs
Affected Class: EOSEACLobbyManager.cs
Add new LobbyChangedEvents params to OnLobbyChanged Callbacks so all Clients will now be capable of identifying what triggered a LobbyChanged callback.

Added new MuteLocalMember and DeafenLocalMember functions which do not rely on CurrentLobby.Members List. The existing ToggleMute function relies on the CurrentLobby.Members list to be populated and will not work before that update is received. MuteLocalMember and DeafenLocalMember can be execute as soon as the Lobby is created.

Added new MuteMember function that will mute a Target ProductUserId. It is similar to the existing ToggleMute function just that this one isn't a toggle and will set Mute status based of the specified value.
_______________________________________________________________________________________________________

Class: EOSTransport.cs
Added UserIdExist and ClientIdExist Functions
Modified Accessibility (Changed to Public) of the following functions: public ProductUserId GetUserId(ulong clientId)
public ulong GetClientId(ProductUserId userId)
The above 2 functions are useful for other custom classes to build upon, especially for Host Migration where we need to be able to lookup the corresponding ProductUserID and ClientID.

### DIFF
--- a/Assets/Scripts/EOSEACLobbyManager.cs
+++ b/Assets/Scripts/EOSEACLobbyManager.cs
@@ -233,7 +233,7 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
             }
         }
 
-        private void OnLobbyChanged()
+        private void OnLobbyChanged(LobbyChangedEvent lobbyChangedEvent)
         {
             string previousLobbyId = CurrentLobbyId;
             var currentLobby = LobbyManager.GetCurrentLobby();
--- a/Assets/Scripts/EOSLobbyManager.cs
+++ b/Assets/Scripts/EOSLobbyManager.cs
@@ -32,6 +32,8 @@ using Epic.OnlineServices.RTCAudio;
 
 namespace PlayEveryWare.EpicOnlineServices.Samples
 {
+    public enum LobbyChangedEvent { Create, Join, Leave, Kicked }
+
     /// <summary>
     /// Class represents all Lobby properties
     /// </summary>
@@ -466,7 +468,7 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
         public delegate void OnMemberUpdateCallback(string LobbyId, ProductUserId MemberId);
 
         private List<OnMemberUpdateCallback> MemberUpdateCallbacks;
-        private List<Action> LobbyChangeCallbacks;
+        private List<Action<LobbyChangedEvent>> LobbyChangeCallbacks;
         private List<Action> LobbyUpdateCallbacks;
 
         private EOSUserInfoManager UserInfoManager;
@@ -490,7 +492,7 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
             LobbySearchCallback = null;
 
             MemberUpdateCallbacks = new List<OnMemberUpdateCallback>();
-            LobbyChangeCallbacks = new List<Action>();
+            LobbyChangeCallbacks = new List<Action<LobbyChangedEvent>>();
             LobbyUpdateCallbacks = new List<Action>();
         }
 
@@ -1331,11 +1333,11 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
 
                 LobbyCreatedCallback?.Invoke(Result.Success);
 
-                OnCurrentLobbyChanged();
+                OnCurrentLobbyChanged(LobbyChangedEvent.Create);
             }
         }
 
-        private void OnCurrentLobbyChanged()
+        private void OnCurrentLobbyChanged(LobbyChangedEvent lobbyChangedEvent)
         {
             if (CurrentLobby.IsValid())
             {
@@ -1343,7 +1345,7 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
             }
             foreach (var callback in LobbyChangeCallbacks)
             {
-                callback?.Invoke();
+                callback?.Invoke(lobbyChangedEvent);
             }
         }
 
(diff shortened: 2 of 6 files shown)


Assets/Scripts/P2PNetcodeSample/Networking/EOSTransport.cs
```diff
@@ -544,8 +548,12 @@ private ProductUserId GetUserId(ulong clientId)
...
         }
 
-        // Returns the ClientId corresponding to a given ProductUserId
-        private ulong GetClientId(ProductUserId userId)
+        /// <summary>
+        /// Returns the ClientId corresponding to a given ProductUserId
+        /// </summary>
+        /// <param name="userId">ProductUserId to use to search for the corresponding ClientId</param>
+        /// <returns></returns>
+        public ulong GetClientId(ProductUserId userId)
```
andrew-hirata-playeveryware, line 556: As mentioned in [#859](https://github.com/PlayEveryWare/eos_plugin_for_unity/issues/859#issuecomment-2307539253), making this public would be problematic because the client ids are not the same as the NetworkManager and would probably add additional confusion.

Assets/Scripts/P2PNetcodeSample/Networking/EOSTransport.cs
```diff
@@ -561,6 +569,44 @@ private ulong GetClientId(ProductUserId userId)
                 return UserIdToClientId[userId];
             }
         }
+
+        /// <summary>
+        /// Check if the specified ClientID exists
+        /// </summary>
+        /// <param name="clientId">The ClientId to search for</param>
+        /// <returns></returns>
+        public bool ClientIDExist(ulong clientId)
```
andrew-hirata-playeveryware, line 578: In a similar way, this clientID mentioned would only work with the transportIDs that are internal to this class, so making this public would imply that the client IDS are the same as the rest of the client IDs in the NetworkManager.

WispyMouse (approved):
Tentative approval from me. The code all seems good, and I have no notes on changing what you have here.

Discussing with the team how to best validate that everything functions properly.

Assets/Scripts/EOSLobbyManager.cs
```diff
@@ -1590,6 +1592,102 @@ public void ToggleMute(ProductUserId targetUserId, OnLobbyCallback ToggleMuteCom
...
+                rtcAudioHandle.UpdateReceiving(ref recevingOptions, MuteMemberCompleted, OnRTCRoomUpdateReceivingCompleted);
+            }
+        }
+
+        /// <summary>
+        /// Set the Mute Status of the Local Member
+        /// </summary>
+        /// <param name="micOn">Active status of the Mic</param>
+        /// <param name="MuteLocalMemberCompleted">Callback when MuteLocalMember is completed</param>
+        public void MuteLocalMember(bool micOn, OnLobbyCallback MuteLocalMemberCompleted)
```
WispyMouse: Perhaps there could be a different name here. It took me a moment to realize that "Local Member" meant "the user that is currently running this". Either we should change the name, or we should add more descriptive comments/documentation.

WispyMouse: Since we have this, I think any time it's determined that the target of the mute is the local user, we should call this function. Right now, the logic of this function is implemented twice in above functions, I think it should just call this instead.

Assets/Scripts/EOSLobbyManager.cs
```diff
@@ -1590,6 +1592,102 @@ public void ToggleMute(ProductUserId targetUserId, OnLobbyCallback ToggleMuteCom
             }
         }
 
+        /// <summary>
+        /// Mute a Member in the Lobby. This only affects the Local Member, other remote Members in the Lobby will still be able to hear them.
+        /// </summary>
+        /// <param name="isMicOn">Active status of the mic</param>
+        /// <param name="targetProductUserId">The Product User ID of the target member to mute</param>
+        /// <param name="MuteMemberCompleted">Callback when MuteMember is completed</param>
+        public void MuteMember(bool isMicOn, ProductUserId targetProductUserId, OnLobbyCallback MuteMemberCompleted)
```
WispyMouse: I feel like the existing ToggleMute should be changed to determine the target status of the mute operation, and then call this function. I think this repeats enough of the logic that we should certainly not have both.

WispyMouse:
@Ermelious : Would you be okay with me pushing commits to your merge branch with some changes that I think would improve this? Wanting to get your go-ahead, since this appears to be coming from another repository that forked this one, and I'd have to change code on your repository.

The specific things I would change are;
- More detailed comment/documentation on the nature of some of the functions
- Utilizing the function you've made in our sample that allows for muting
- Changing the existing mute function to call the function you've made

Xeynt:
> @Ermelious : Would you be okay with me pushing commits to your merge branch with some changes that I think would improve this? Wanting to get your go-ahead, since this appears to be coming from another repository that forked this one, and I'd have to change code on your repository.
> 
> The specific things I would change are;
> 
> * More detailed comment/documentation on the nature of some of the functions
> * Utilizing the function you've made in our sample that allows for muting
> * Changing the existing mute function to call the function you've made

@WispyMouse Sure! Please go ahead. If there's any access issues feel free to let me know. I think I've already clicked on the option to allow the original contributors to access the forked repo.

Assets/Scripts/EOSEACLobbyManager.cs
```diff
@@ -233,7 +233,7 @@ private void OnLobbyUpdated()
             }
         }
 
-        private void OnLobbyChanged()
+        private void OnLobbyChanged(LobbyChangedEvent lobbyChangedEvent)
```
arthur740212, line 236: It could be discussed whether to pass an argument here or to have separate events for each type.
The OnLobbyChanged by its name handles lobby changes regardless how the lobby was changed.
It feels indirect to have, for instance, OnKicked to call OnCurrentLobbyChanged then back to a kicked state handler.
Let us know your thoughts, we will toss it around the team.

WispyMouse:
Okay, made some adjustments. Would love people to review as it is, now.

I tested it by making two clients and deafening myself on either to make sure that works, as well as muting users. Could use more rigorous testing but should be solid?

Xeynt:
I'm not too sure why my direct responses are marked as "pending" and invisible by Git so will response here instead:

@andrew-hirata-playeveryware This is related to my response to your response: https://github.com/PlayEveryWare/eos_plugin_for_unity/issues/859#issuecomment-2307539253
Can we check, how do we use that DisconnectRemoteClient function if we don't have access to the EOSTransport's ClientId/ProductUserId lookup dictionary? More information in the original response linked above.
Kindly advise.

Hi @paulhazen , based on the discussions in this issue (which is related to the functions Andrew has highlighted and an earlier discussion regarding Disconnecting a client): https://github.com/PlayEveryWare/eos_plugin_for_unity/issues/859#issuecomment-2307539253
It seems like DisconnectRemoteClient which relies on the ClientID assigned by EOSTransport is not the function that we should be using to disconnect another client since as mentioned by @andrew-hirata-playeveryware the ClientID/ProductUserID lookup in EOSTransport is supposed to be non-accessible and private.

If that is the case, then is there a way to Disconnect a client out of the Server from the Host? Especially since NetworkManager.Singleton.Disconnect function is not working in this plugin as discussed earlier in another issue (which is now closed). All the links to the relevant issues are in the above linked issue.

Xeynt:
@arthur740212 
A separate event for each type sounds great as well. The initial thoughts for that LobbyChangedEvent enum is for the client affected by the event to identify what triggered it. E.g Kicked by the Host. (Because while kicking a player from the Server, we could not find an existing functionality/callback to subscribe to for this, except for the OnLobbyChanged though there wasn't any way to know what triggered that callback. With that enum, it solved our issues and gave us a way to be able to know exactly if it was caused by a "Kicked" or not.)
```c#
        eosAssets.lobbyManager.AddNotifyLobbyChange((LobbyChangedEvent lobbyChangedEvent) =>
        {
            // If kicked by the Host, display a Prompt to inform this Client that they've been kicked by the host.
            if (lobbyChangedEvent == LobbyChangedEvent.Kicked)
            {
                assets.promptPanel.DisplayPrompt(string.Format(GlobalLobbyConstants.PROMPTPANEL_DISPLAY_KICKED_FROM_LOBBY_MESSAGE, cache.mostRecentLobbyOwnerDisplayName));
                callbacks.OnLeaveLobbyCompleted?.Invoke();
            }
        });
```
If these new set of separate events that you've proposed would give the clients the ability to know what caused the Lobby to changed e.g when Create, Kicked, Join, Leave etc Or when either one of those event occurs for the remote client, on the remote client, so that we can subscribe for the purpose of e.g Displaying a prompt to tell the player the event occurred, (in the above example, we are displaying a prompt to inform the player of the unfortunate event of being kicked), then please go ahead with that if you/the team thinks it is the preferred approach. If that is the approach the team decides, I'll subscribe to those instead of what I've proposed here.

Additionally correct me if I'm wrong but from my understanding, OnLobbyChanged should only be invoked on the client that triggered that event, is this correct? E.g if another client joins, the OnLobbyChanged callback will only invoke on the client that joins the lobby and not on the rest of the clients and host. If this is correct, then having an additional param in the callbacks for ClientID (that triggered the event) would not be necessary since we would know it is always the client themselves that executes the OnLobbyChanged callbacks and not on the other clients and host. And for Kicked, it should also be a definite that the Host is the one that triggered that callback. (So we wouldn't need an extra ClientID param to lookup to the user that kicked this Client, we can just assume the Host's ID is the one.)

WispyMouse:
@Ermelious Ahoy! Something we discussed internally is that this pull request has three separate concerns; adding deafen and muting, add a way to know what the lobby changed event kind is, and the surfacing of these Ids. @andrew-hirata-playeveryware has identified that the surfacing of the ids is not going to be accepted, but the other two changes could be.

What I'm going to do is cherry pick your changes into two separate change lists, and make pull requests for those into this repository's development branch. The changes will still be identified on GitHub as being made by you, and then I'll apply my changes on top of that.

Xeynt:
> @Ermelious Ahoy! Something we discussed internally is that this pull request has three separate concerns; adding deafen and muting, add a way to know what the lobby changed event kind is, and the surfacing of these Ids. @andrew-hirata-playeveryware has identified that the surfacing of the ids is not going to be accepted, but the other two changes could be.
> 
> What I'm going to do is cherry pick your changes into two separate change lists, and make pull requests for those into this repository's development branch. The changes will still be identified on GitHub as being made by you, and then I'll apply my changes on top of that.

@WispyMouse Yes I should have separated the PR, thanks for the assistance in the cherry-picking.

Noted that we are not supposed to use DisconnectRemoteClient (which requires EOSTransports ClientIDs). Was initially confused since the team had stated earlier that NetworkManager.Singleton.Disconnect is intended to not work in the SDK, hence the confusion with EOSTransport.DisconnectRemoteClient.

Will wait for the team's fixes for a proper way to Disconnect a Client from the Server, as the NetworkManager.Singleton.Disconnect(clientID) is not working in this SDK, it is impossible to disconnect a client from the host in this SDK at the moment. 

Hoping for the fix to land in this SDK soon as it is an important feature required in my projects, thanks!

WispyMouse (changes requested):
This should now be handled by #875 and #874 . This PR should not be merged in as is.

The work here is cherry-picked to the other PRs.

WispyMouse:
@Ermelious I'm going to close this PR, as we've decided not to go with surfacing the ClientID, as well as breaking out the work in to the other PRs. Thank you so much for your contributions, we'll work on getting those in soon!

Issue #850, opened 2024-08-18
Can't Retrieve Connect Interface on Android


kdserra:
**Describe the bug**
I upgraded from PlayEveryWare EOS 3.2.0 to 3.3.1 which introduced a bug where the plugin is unable to retrieve the EOS Connect Interface on Android.  It works perfectly fine in the Windows 10 Editor.

I did not forget to place EOSManager in scene, and I sucessfully retrieve the Singleton instance, just not the Connect Interface.

See video of the issue:
https://www.youtube.com/watch?v=BO-KQQ59xeM

**To Reproduce**
(Only tested on physical device Pixel 4, Android 13)

1. Create new Unity 2023.20f1 Project, Template: 3D BiRP
2. Import tarball: com.playeveryware.eos-3.3.1
3. Import Sample: "Standard Pack: EOS Plugin for Unity Samples"
4. Go to File -> Build Settings -> Switch to Android platform
5. Go to: Edit -> Project Settings -> Player -> Select Android -> Set Minimum API Level to "Android 6.0 'Marshmallow' (API level 23)"

6. After Import - Add the following script:

Demo.cs
```csharp
using UnityEngine;
using PlayEveryWare.EpicOnlineServices;
using TMPro;

public class Demo: MonoBehaviour
{
    [SerializeField] private TMP_Text _tmpText;
    
    private void Start()
    {
        var eosInstance = EOSManager.Instance;
        var connectInterface = eosInstance.GetEOSConnectInterface();
        _tmpText.text = "Retrieved EOS Connect Interface: " + (connectInterface != null);
    }
}
```

7. Attach Demo script to GameObject in scene
8. Assign _tmpText reference in scene to TMP Text
9. Save Scene
10. Run in editor, should show "Retrieved EOS Connect Interface: True" - worked fine.
11. Build for Android with Mono Scripting Backend
12. Run app on device
13. Notice text says "Retrieve EOS Connect: False".

**Expected behavior**
I expect the app to successfully retrieve the Connect Interface.
In the demo the text should say "Retrieve EOS Connect: True" on Android device.

**Actual behavior**
The app fails to retrieve the Connect Interface.
In the demo the text says "Retrieve EOS Connect: False" on Android device.

**Screenshots**
![1-exception](https://github.com/user-attachments/assets/718732e0-7b86-4e44-bc10-95ed7b703c6c)
[...]

PR #864, opened 2024-08-26
Narrow the usage of `System.IO` within non-editor code

paulhazen:
This PR restricts usage of the `System.IO` namespace to primarily be within the `StorageUtility` class (formerly named `FileUtility`). 

The class name was changed because restricting usage of the `System.IO` namespace introduced functionality to `FileUtility` that is not strictly speaking related to _files_. The new name reflects that change and the expanded scope of the class's functionality.

This PR _also_ changes the `StorageUtility` and some associated Utility classes to have the `internal` access modifier, so as to ensure that their usage is primarily restricted to contexts within the EOS Plugin.

> [!NOTE] 
> This is part of a preliminary effort to resolve Issue #850, as it is suspected that the `ConnectInterface` is not being properly instantiated due to a quirk of IO operations on Android devices.
 
#EOS-2048


com.playeveryware.eos/Runtime/Core/Utility/StorageUtility.cs
```diff
@@ -40,7 +51,7 @@ namespace PlayEveryWare.EpicOnlineServices.Utility
     /// <summary>
     /// Utility class used for a variety of File tasks.
     /// </summary>
-    public static class FileUtility
+    internal static class StorageUtility
```
paulhazen: `StorageUtility` is kind of _icky_, I think we should go with `FileSystemUtility`.

com.playeveryware.eos/Runtime/Core/Utility/StorageUtility.cs
```diff
@@ -20,17 +20,28 @@
  * SOFTWARE.
  */
 
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("com.playeveryware.eos-Editor")]
```
paulhazen: Confirm that this applies only to the classes defined in this file - not to the entire `PlayEveryWare.EpicOnlineServices.Utility`.

com.playeveryware.eos/Runtime/Core/Utility/StorageUtility.cs
paulhazen: It has been so renamed.

com.playeveryware.eos/Runtime/Core/DLLHandle.cs
```diff
@@ -22,252 +22,203 @@
...
+            if (null == libraryPath)
             {
-                return filesystemEntry;
+                return null;
             }
+
+            IntPtr libraryHandle = SystemDynamicLibrary.Instance.LoadLibraryAtPath(libraryPath);
+            print("Trying to load with entry " + libraryPath);
+
+            return libraryHandle != IntPtr.Zero ? new DLLHandle(libraryHandle) : null;
```
andrew-hirata-playeveryware: This change to remove the enumeration now won't throw an exception when it is unable to load a found dynamic library. In the previous code, it returns null if there is no matching dynamic library on disk, however if it finds a match on disk but `SystemDynamicLibrary.LoadLibraryAtPath` is unable to load it (i.e. it returns null), then it will throw as that is seen as an unrecoverable error.

The new code will return null in both cases.

andrew-hirata-playeveryware (changes requested):

com.playeveryware.eos/Runtime/Core/DLLHandle.cs
paulhazen: Good catch. I'm resolving that now.

andrew-hirata-playeveryware (approved):

Assets/Plugins/Android/Core/AndroidFileIOHelper.cs
```diff
@@ -23,35 +23,50 @@
...
+
+            while (!operation.isDone)
             {
-                request.timeout = 2; //seconds till timeout
-                request.SendWebRequest();
+                await Task.Yield();
+            }
 
-                //Wait till webRequest completed
-                while (!request.isDone) { }
```
WispyMouse, line 38: haha wow this is a terrifying bit of code. Glad you adjusted it.

Assets/Plugins/Android/Core/AndroidFileIOHelper.cs
```diff
@@ -23,35 +23,50 @@
...
+                                   $"UnityWebRequest for path " +
+                                   $"\"{filePath},\" failed with error code " +
+                                   $"'{request.result},' and error, " +
+                                   $"\"{request.error}.\"");
+                    break;
+                case UnityWebRequest.Result.Success:
+                    text = request.downloadHandler.text;
+                    break;
+                default:
+                    throw new ArgumentOutOfRangeException();
```
WispyMouse: I feel like we could log out the result/error here, assuming this can ever be hit.

Assets/Plugins/Source/Editor/EditorWindows/CheckDeploymentWindow.cs
```diff
@@ -101,7 +101,7 @@ protected override void RenderWindow()
 
         private void LoadTextFile(string filepath)
         {
-            content = FileUtility.ReadAllText(filepath);
+            content = FileSystemUtility.ReadAllText(filepath);
```
WispyMouse, line 104: Is it feasible to make this use the async version somehow?

com.playeveryware.eos/Runtime/Core/Config.cs
```diff
@@ -324,19 +303,51 @@ protected virtual async Task ReadAsync()
...
+        /// <summary>
+        /// Determines if the config file exists, and if it does not, and the
+        /// editor is running, then create the file.
+        /// </summary>
+        /// <param name="isAsync">
+        /// Indicates whether the operation should run synchronously or
+        /// asynchronously.
+        /// </param>
+        /// <returns>Task.</returns>
+        private async Task EnsureConfigFileExistsAsync(bool isAsync)
```
WispyMouse: I feel weird about this... could we use a Task<bool> to return if it exists or not, instead of relying on the exception? I guess this is mostly for config code, which should always exist.

com.playeveryware.eos/Runtime/Core/Config.cs
```diff
@@ -324,19 +303,51 @@ protected virtual async Task ReadAsync()
...
-            JsonUtility.FromJsonOverwrite(_lastReadJsonString, this);
+        /// <summary>
+        /// Determines if the config file exists, and if it does not, and the
+        /// editor is running, then create the file.
+        /// </summary>
+        /// <param name="isAsync">
+        /// Indicates whether the operation should run synchronously or
+        /// asynchronously.
+        /// </param>
+        private void EnsureConfigFileExists(bool isAsync)
```
WispyMouse: It feels odd that this allows for `isAsync`, given that it always will be synchronous since it's getting the awaiter and result.

Assets/Plugins/Android/Core/AndroidFileIOHelper.cs
paulhazen: It cannot be hit right now - but is there because it completes the coverage of all possible cases within the `switch` statement.

Assets/Plugins/Source/Editor/EditorWindows/CheckDeploymentWindow.cs
paulhazen: Possibly - there are many areas where we are using synchronous code where asynchronous code would arguably make more sense - however doing that often requires a lot more refactoring to do it properly. This is why we maintain synchronous calls. I'm not as familiar with the `CheckDeploymentWindow` code, so I am not sure what would be required. For the scope of this PR however, the intent is only to change the name of the `FileUtility` to `FileSystemUtility` to reflect the expanded functionality of the class.

com.playeveryware.eos/Runtime/Core/DLLHandle.cs
paulhazen: I have resolved this issue, consequently I am marking this as resolved.

com.playeveryware.eos/Runtime/Core/Utility/FileSystemUtility.cs
```diff
@@ -20,6 +20,9 @@
  * SOFTWARE.
  */
 
+using System.Runtime.CompilerServices;
```
WispyMouse, line 23: Can this be moved under namespace, or does it need to be defined before the InternalsVisibleTo attribute piece?

com.playeveryware.eos/Runtime/Core/Utility/FileSystemUtility.cs
```diff
@@ -455,20 +465,31 @@ public static void ConvertDosToUnixLineEndings(string srcFilename, string destFi
 
         #endregion
 #endif
+
+        #region File Read Functionality
+
+        public static async Task<(bool Success, string Result)> TryReadAllTextAsync(string filePath)
```
WispyMouse, line 471: Hhhuh. Is this a Task that returns an anonymous function? I hadn't considered that as something possible. It's like a named Tuple... I feel like I'd like this more as a struct with these fields.

com.playeveryware.eos/Runtime/Core/Utility/FileSystemUtility.cs
```diff
@@ -478,15 +499,235 @@ public static string ReadAllText(string path)
...
+        /// <param name="filePath">The path to write to.</param>
+        /// <param name="content">The contents to write to the file.</param>
+        /// <param name="createDirectory">
+        /// Whether or not to create the directory structure represented by the
+        /// filepath.
+        /// </param>
+        public static void WriteFile(string filePath, string content, bool createDirectory = true)
+        {
+            // Use discard in-order to call async function without awaiting.
+            _ = WriteFileInternal(filePath, content, createDirectory);
```
WispyMouse: Does this actually run it synchronously? It looks like it just kicks off the function, but doesn't wait for it. Would calling this function promise that it'll be done by the time another function runs?

com.playeveryware.eos/Runtime/Core/Utility/FileSystemUtility.cs
```diff
@@ -478,15 +499,235 @@ public static string ReadAllText(string path)
...
+            FileInfo file = new(filePath);
+
+            // If the directory should be created, and the DirectoryInfo is not
+            // null, then create the directory.
+            if (createDirectory && null != file.Directory)
+            {
+                CreateDirectory(file.Directory);
+            }
+
+            await using StreamWriter writer = new(filePath);
```
WispyMouse, line 635: Interesting, I've never considered awaiting a StreamWriter. Learned something from this.

com.playeveryware.eos/Runtime/Core/Utility/FileSystemUtility.cs
```diff
@@ -478,15 +499,235 @@ public static string ReadAllText(string path)
...
+        /// </param>
+        private static void CreateDirectory(DirectoryInfo dInfo)
+        {
+            try
+            {
+                dInfo.Create();
+            }
+            catch(Exception ex) 
+            {
+                Debug.LogError($"Could not create directory " +
```
WispyMouse: Can we log the exception and pass it along?

WispyMouse: i.e. Debug.LogException

Assets/Plugins/Android/Core/AndroidFileIOHelper.cs
WispyMouse: Just in case, can you please add a log that prints out the resulting error?

Assets/Plugins/Source/Editor/EditorWindows/CheckDeploymentWindow.cs
WispyMouse: Valid. Room for improvement in another space, perhaps.

Assets/Plugins/Android/Core/AndroidFileIOHelper.cs
arthur740212: Yeah, I did that. It sure looks scary. The request has the timeout set to 2 though.
There are definitely better ways to achieve the same thing.

WispyMouse:
I'm getting Unity to hang by trying to open the EOS Configuration dropdown. Not the popup, the dropdown; trying to go to Tools > EOS results in Unity permanently hanging.

Assets/Plugins/Android/Core/AndroidFileIOHelper.cs
paulhazen: Yup :) just did.

com.playeveryware.eos/Runtime/Core/Utility/FileSystemUtility.cs
paulhazen: It needs to be defined before the `InternalsVisibleTo` attribute.

com.playeveryware.eos/Runtime/Core/Config.cs
paulhazen: Yeah re-reading these changes they are pretty wonky. I replaced them with a better approach, which is, when an "Async" function needs to be called from a non-async context, do the following:

```csharp
public async Task DoWorkAsync()
{
    /* Some kind of async calls using the "await" keyword */
}

public void DoWork()
{
    DoWorkAsync().GetAwaiter().GetResult()
}
```

And for circumstances where the result is not void:

```csharp
public async Task<int> DoWorkAsync()
{
    /* Some kind of async calls using the "await" keyword */
}

public int DoWork()
{
    return DoWorkAsync().GetAwaiter().GetResult()
}
```

I am going to mark the other comment you made regarding this as resolved, so we can continue the conversation in a single place (this conversation)

com.playeveryware.eos/Runtime/Core/Config.cs
paulhazen: Per [this comment](https://github.com/PlayEveryWare/eos_plugin_for_unity/pull/864#discussion_r1735448207), I am marking _this_ conversation as resolved.

com.playeveryware.eos/Runtime/Core/Utility/FileSystemUtility.cs
paulhazen: It's kind of a cool C#-y thing to do. 

It returns a `Tuple`, not an anonymous function. 

The equivalent function signature _without_ naming the tuple elements would be:

```csharp
public static async Task<(bool, string)> TryReadAllTextAsync(string filePath)
```

Naming the components helps avoid less than clear code when handling the return value which (without the tuple naming) would require the following syntax when utilizing the returned (and unnamed) `Tuple`:

```csharp
var result = await FileSystemUtility.TryReadAllTextAsync('SomeFile.txt');
bool didReadSucceed = result.item1;
string fileContents = result.item2;

if (didReadSucceed)
{
    /* ... */
}
```

com.playeveryware.eos/Runtime/Core/Utility/FileSystemUtility.cs
paulhazen: Good inference. You are correct that it is not guaranteed to be finished. Per an earlier comment, I've updated the implementation to do it properly.

com.playeveryware.eos/Runtime/Core/Utility/FileSystemUtility.cs
paulhazen: Yuppers! All ficksed.

paulhazen:
> I'm getting Unity to hang by trying to open the EOS Configuration dropdown. Not the popup, the dropdown; trying to go to Tools > EOS results in Unity permanently hanging.

Took me a while to hunt this down. I'm honestly not certain what ended up fixing it - but I was able to initially reproduce the bug, and after changing the "Sign DLLs" `MenuItem` "validate" function to use `Config.GetAsync<SigningConfig>()` instead of `Config.Get<SigningConfig>()`, the issue was resolved.

WispyMouse (approved):

PR #865, opened 2024-08-26
Unit Test Code Coverage Measurement Tools

paulhazen:
## Code Coverage

This PR enables Unity's Code Coverage utility. In concert with the Test Runner and our Unit Tests, it allows us to measure current code coverage and track historic code coverage. This is critical to understanding how good (or bad) our test coverage is in a data-driven manner that can help keep us accountable to our users through increased transparency.

<div align="center">

</div>

## Visibility & Historic Tracking

This PR adds `.gitignore` files, and automation scripts that ensure historic coverage data is retained in the form of the _last run coverage analysis for the day_. This means that when you run the Unit Tests locally, and you subsequently commit changes - there will be only two files changed. The first is an XML file that retains the minimal amount of historical coverage data, the second is a badge image `.svg` that can be embedded in our `README.md` file.

> [!NOTE]
> Historic data is not a part of the visuals that are committed to the repository. When Unit Tests are run, a report will be generated at `tools/coverage/results/Report/Index.html`. 
>
> While the report _itself_ is ignored by `git`, the _historic coverage data_ utilized by the Code Coverage utility _is_ tracked by `git` so that the report can display a graph showing the level of code coverage over time.
>
> Since there is not currently any historical information regarding coverage, this graph does not look interesting, and so is not included here.

## Future Plans

Over time, running of these Unit Tests can and will be done entirely through actions - but this will serve as a good stop-gap measure to enable us to start to track the coverage of our tests.

## Coverage Configuration

Tests are configured to only consider core plugin code, editor code, and sample code as lines to cover for testing purposes (the EOS SDK itself is intentionally excluded from code coverage consideration).

> [!WARNING]
> It might make more sense to the community for us to delay the inclusion of the test coverage badge until we are able to have a coverage of at least 20%. 
> Conversely, including it now can demonstrate the speed with which we increase coverage.

#EOS-2049


tools/coverage/historic/Consolidated_CoverageHistory.xml
andrew-hirata-playeveryware, line 1: I think this file should not be in git-lfs, as I don't believe it will get to the size that github won't let us store it as a regular file in the repo.

andrew-hirata-playeveryware (changes requested):

tools/coverage/historic/Consolidated_CoverageHistory.xml
paulhazen: I've removed the file from being tracked by Git LFS, and re-added it using vanilla git. Consequently, I'm marking this conversation as resolved.

Assets/Tests/Common/CodeCoverageHistoryCleaner.cs
```diff
@@ -0,0 +1,170 @@
...
+        public static void ExtractLineCoverage()
+        {
+            XDocument summaryDoc = XDocument.Load(CODE_COVERAGE_REPORT_SUMMARY_FILE);
+
+            var summaryElement = summaryDoc.Descendants("Summary").First();
+
+            ulong coveredLines = ulong.Parse(summaryElement.Descendants("Coveredlines").First()?.Value);
+            ulong coverableLines = ulong.Parse(summaryElement.Descendants("Coverablelines").First()?.Value);
+
+            double percentage = (double)coveredLines / coverableLines * 100;
```
WispyMouse, line 89: Is there ever a world where coverableLines can be 0?

Assets/Tests/Common/CodeCoverageHistoryCleaner.cs
```diff
@@ -0,0 +1,170 @@
...
+                // Save the contents of the consolidated history. Omit any 
+                // duplicate namespaces and disable formatting, since this file
+                // should only ever be read by the code coverage utility itself.
+                root.Save(outputFilePath, SaveOptions.DisableFormatting | SaveOptions.OmitDuplicateNamespaces);
+
+                Debug.Log($"Consolidated CoverageHistory saved to {outputFilePath}");
+            }
+            catch (Exception ex)
+            {
+                Debug.LogError($"CoverageHistoryCleaner failed: {ex.Message}");
```
WispyMouse: Consider Debug.LogException

tools/coverage/historic/.gitignore
```diff
@@ -0,0 +1,5 @@
+# Ignore all files in the historic directory.
+*
+# Except for _this_ .gitignore file, the consolidated coverage history file, and the gitattributes file
+!.gitignore
```
WispyMouse, line 4: not real criticism: This is hilarious to me. "Please ignore.. not me!"

WispyMouse:
I have a question about what this'll do moving forward. For every commit going forward, should we be manually running tests, observing code coverage results, and committing some subset of these new coverage files?

Should we be running *every* test? I just ran some subset of tests and it thought the code coverage was .7%, but that's probably because the subset I ran only affected that much of the code base.

So we'd commit;
- Cosonlidated_CoverageHistory.xml
- current_coverage.txt
- badge_linecoverage.svg

?

WispyMouse (approved):

Assets/Tests/Common/CodeCoverageHistoryCleaner.cs
paulhazen: Not really, but I added a check for that since doing so is good practice.

Assets/Tests/Common/CodeCoverageHistoryCleaner.cs
paulhazen: Good suggestion! I just changed it to that.

paulhazen:
> I have a question about what this'll do moving forward. For every commit going forward, should we be manually running tests, observing code coverage results, and committing some subset of these new coverage files?
> 
> Should we be running _every_ test? I just ran some subset of tests and it thought the code coverage was .7%, but that's probably because the subset I ran only affected that much of the code base.
> 
> So we'd commit;
> 
> * Cosonlidated_CoverageHistory.xml
> * current_coverage.txt
> * badge_linecoverage.svg
> 
> ?

Unclear at the current moment. There is no need to run tests regularly unless you want to increase the line coverage numbers. So probably only commit them when there is an increase. Future changes to this flow could get more nuanced.

andrew-hirata-playeveryware (approved):

PR #871, opened 2024-08-29
Namespace Cleanup (for Unit Test measurements)

paulhazen:
This PR is the result of a swath of changes that were made in _another branch_ during the process of expanding the set of Unit Tests that we have. The code coverage utility that is built into Unity organizes and displays the code coverage by summarizing the coverage at the namespace level.

Consequently, having our namespaces in a semi-inconsistent state makes it difficult to measure increases in code coverage.

Because the changes made in the other branch _also_ included additional unit tests, I didn't want to conflate the two tasks, so I separated the changes that were exclusive to the namespace organization into this PR. This will make the subsequent PR that adds to our Unit Tests much smaller and focused exclusively on introducing _new unit tests_, instead of optimizing the way those unit tests' coverage _is measured_.

> [!NOTE]
> There is _nothing functionally different_ about the changes that this PR introduces - each file should only show changes that reflect the relocation of `using` statements, the addition of namespaces to files that previously did not have them, and corrections to namespace usage that make things more consistent.

 #EOS-2054


Assets/Plugins/Source/Editor/Configs/SteamConfig.cs
```diff
@@ -20,14 +20,16 @@
 * SOFTWARE.
 */
 
-using System;
-using System.Collections.Generic;
 
-namespace PlayEveryWare.EpicOnlineServices
+
+namespace PlayEveryWare.EpicOnlineServices.Editor.Config
```
andrew-hirata-playeveryware: I believe this should stay in an non-editor namespace.
While it is currently the case that this config is only used in the Editor, it _should_ be used on Linux and macOS when running under Steam.

Assets/Plugins/Source/Editor/Build/PlatformSpecificBuilder.cs
```diff
@@ -31,10 +31,8 @@ namespace PlayEveryWare.EpicOnlineServices.Build
     using UnityEditor.Build.Reporting;
     using Debug = UnityEngine.Debug;
     using UnityEditor;
-    using PlayEveryWare.EpicOnlineServices.Editor;
     using System;
-    using System.Threading.Tasks;
-    using Utility;
+    using Config = EpicOnlineServices.Config;
```
WispyMouse, line 35: noncrit: Having both `using Config` and `using Config = EpicOnlineServices.Config` is odd at first sight, but apparently completely legal to do. Seems to work. PrebuildConfig depends on this using, so it is fine.

Assets/Scripts/StandardSamples/Core/EOSHostManager.cs
WispyMouse, line 1: possible noncrit: `EOSPSNManagerPS5.EnsurePS5Initialized` relies on something in this same namespace. I think it's safe to remove those other usings, because that class isn't in those namespaces. So this is probably good to go for removing these usings.

WispyMouse (approved):

com.playeveryware.eos/Runtime/Core/DLLHandle.cs
```diff
@@ -20,8 +20,6 @@
 * SOFTWARE.
 */
 
-﻿//#define ENABLE_DLLHANDLE_PRINT
```
andrew-hirata-playeveryware, line 23: This should go back, but perhaps with a comment explaining what it does.

com.playeveryware.eos/Runtime/Core/EnumeratorWrapper.cs
```diff
@@ -1,42 +0,0 @@
-/*
```
andrew-hirata-playeveryware: I might be mistaken, but this file delete doesn't seem relevant to the namespace cleanup.

andrew-hirata-playeveryware (changes requested):

com.playeveryware.eos/Runtime/Core/DLLHandle.cs
paulhazen: Done! :)

com.playeveryware.eos/Runtime/Core/EnumeratorWrapper.cs
paulhazen: Good call. I restored it.

Assets/Plugins/Source/Editor/Configs/SteamConfig.cs
paulhazen: Shouldn't it also be used in Windows (non-editor) when running under steam? If that is what is supposed to be happening, it may need to be fixed there because it's not referenced _anywhere_ in non-editor code. 

Regardless, I have restored it to the previous namespace for the time being.

andrew-hirata-playeveryware: On windows the steam config is read in the GFX plugin, so this code won't be hit when running on Windows standalone builds.

andrew-hirata-playeveryware (approved):

PR #874, opened 2024-08-29
feat(lobby): Indicate lobby changed event kind

WispyMouse:
This adds a way for subscribers of the lobby changing event to know what kind of event has occurred.

This is a pull request split from: https://github.com/PlayEveryWare/eos_plugin_for_unity/pull/862

There is consideration if we instead of having this enum, want to use four different events. That discussion should happen here.

Original work from @Ermelious

### DIFF
--- a/Assets/Scripts/EOSLobbyManager.cs
+++ b/Assets/Scripts/EOSLobbyManager.cs
@@ -30,7 +30,13 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
     using Epic.OnlineServices.RTC;
     using Epic.OnlineServices.RTCAudio;
 	
-    public enum LobbyChangedEvent { Create, Join, Leave, Kicked }
+    public enum LobbyChangeType
+    { 
+        Create, 
+        Join, 
+        Leave,
+        Kicked
+    }
 
     /// <summary>
     /// Class represents all Lobby properties
@@ -475,7 +481,16 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
         public delegate void OnMemberUpdateCallback(string LobbyId, ProductUserId MemberId);
 
         private List<OnMemberUpdateCallback> MemberUpdateCallbacks;
-        private List<Action<LobbyChangedEvent>> LobbyChangeCallbacks;
+
+
+        public delegate void LobbyChangedEventHandler(string lobbyId, LobbyChangeType typeOfChange);
+
+        /// <summary>
+        /// Event that is run whenever the local user's relationship to a Lobby has been changed.
+        /// Indicates the Lobby that the change relates to.
+        /// </summary>
+        public event LobbyChangedEventHandler OnLobbyChanged;
+
         private List<Action> LobbyUpdateCallbacks;
 
         private EOSUserInfoManager UserInfoManager;
@@ -499,7 +514,6 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
             LobbySearchCallback = null;
 
             MemberUpdateCallbacks = new List<OnMemberUpdateCallback>();
-            LobbyChangeCallbacks = new List<Action<LobbyChangedEvent>>();
             LobbyUpdateCallbacks = new List<Action>();
         }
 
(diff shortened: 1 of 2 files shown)


andrew-hirata-playeveryware (approved):

Assets/Scripts/EOSLobbyManager.cs
```diff
@@ -32,6 +32,8 @@
 
 namespace PlayEveryWare.EpicOnlineServices.Samples
 {
+    public enum LobbyChangedEvent { Create, Join, Leave, Kicked }
```
paulhazen: My suggestion would be to utilize the following delegate and event, updating the corresponding places within this PR that would need to reference it:

(these would be inside `EOSLobbyManager`, and comments have been excluded for the sake of brevity):

```csharp

public enum LobbyChangeType
{
    Create,
    Join,
    Leave,
    Kicked
}

public delegate void LobbyChangedEventHandler(string lobbyId, LobbyChangeType typeOfChange);

public event LobbyChangedEventHandler OnLobbyChanged;

**Review by paulhazen — CHANGES_REQUESTED** [review]

**Comment by WispyMouse** [comment]
We've thought about it as a team and had a couple of ideas for change;

- Use an Event instead of an Action, which fits a strategy our team is moving towards
- Use a Delegate as the event kind, storing the type of lobby change event as well as an identifier for the lobby

I'm going to do that work, push that up, and then continue review.

File: Assets/Scripts/EOSLobbyManager.cs (hunk 1 above)
Comment (WispyMouse) [reply Assets/Scripts/EOSLobbyManager.cs]: Thank you for the expressive example! I'll implement it basically as you have here.

I'm impressed by your use of the markdown for comments, I need to start utilizing this.

**Review by paulhazen — APPROVED** [review]
Changed my review to be an approval, since merging will require resolution of comments, and my concerns are articulated there.

File: Assets/Scripts/EOSLobbyManager.cs [hunk 2]
```diff
@@ -468,7 +474,16 @@ public class EOSLobbyManager : IEOSSubManager
...
-        private List<Action<LobbyChangedEvent>> LobbyChangeCallbacks;
+
+
+        public delegate void LobbyChangedEventHandler(string lobbyId, LobbyChangeType typeOfChange);
+
+        /// <summary>
+        /// Event that is run whenever the local user's relationship to a Lobby has been changed.
+        /// Indicates the Lobby that the change relates to.
+        /// </summary>
+        public event LobbyChangedEventHandler OnLobbyChanged;
```
paulhazen: This should be named, `LobbyChanged`. The fact that it is an event is the indicator that it is an action / event / verb. This allows the functions that _subscribe_ to it (ala the `+=` operator) to be named `OnLobbyChange` without confusion.

paulhazen: I should note that we are not doing this consistently throughout the project, but that this is the Correct™ way to do it from my understanding of industry conventions. @andrew-hirata-playeveryware, is this consistent with your perspective on the topic?

paulhazen (changes requested):
Noticed some changes that happened after the review that I either failed to notice earlier or are new. Minor thing to fix I believe.

Assets/Scripts/EOSLobbyManager.cs
paulhazen: I did just find this: https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members#names-of-events

There are other aspects of this we do not follow (and probably should), but the one regarding what to name the event is consistent with my recommendation I believe.

paulhazen: Here are three examples within our code base of us using different naming conventions:

https://github.com/PlayEveryWare/eos_plugin_for_unity/blob/7806e659f285992b38a17395a4495f56dcb8d078/Assets/Scripts/StandardSamples/Steam/SteamManager.cs#L71

https://github.com/PlayEveryWare/eos_plugin_for_unity/blob/7806e659f285992b38a17395a4495f56dcb8d078/Assets/Scripts/StandardSamples/Services/EOSService.cs#L47

https://github.com/PlayEveryWare/eos_plugin_for_unity/blob/7806e659f285992b38a17395a4495f56dcb8d078/Assets/Scripts/StandardSamples/Discord/DiscordManager.cs#L177

WispyMouse: For now, dropped `EventHandler` from the name.

paulhazen (approved):

WispyMouse:
Thank you for the contribution, @Ermelious ! This is going to be merged into `development` now, and will be available in a future release. We may decide to revisit this event structure later, but for now we have the information propagating the way you wanted.

Xeynt:
> Thank you for the contribution, @Ermelious ! This is going to be merged into `development` now, and will be available in a future release. We may decide to revisit this event structure later, but for now we have the information propagating the way you wanted.

Interesting, thanks for the update and info!

PR #875, opened 2024-08-29
feat(lobby): Mute, deafen, without Lobby Members

WispyMouse:
This adds the ability to deafen yourself, as well as a couple more convenient functions for muting target users.

This is a pull request split from: https://github.com/PlayEveryWare/eos_plugin_for_unity/pull/862

### DIFF
--- a/Assets/Scripts/EOSLobbyManager.cs
+++ b/Assets/Scripts/EOSLobbyManager.cs
@@ -1590,6 +1590,102 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
             }
         }
 
+        /// <summary>
+        /// Mute a Member in the Lobby. This only affects the Local Member, other remote Members in the Lobby will still be able to hear them.
+        /// </summary>
+        /// <param name="isMicOn">Active status of the mic</param>
+        /// <param name="targetProductUserId">The Product User ID of the target member to mute</param>
+        /// <param name="MuteMemberCompleted">Callback when MuteMember is completed</param>
+        public void MuteMember(bool isMicOn, ProductUserId targetProductUserId, OnLobbyCallback MuteMemberCompleted)
+        {
+            RTCInterface rtcHandle = EOSManager.Instance.GetEOSRTCInterface();
+            RTCAudioInterface rtcAudioHandle = rtcHandle.GetAudioInterface();
+
+            LobbyMember lobbyMember = GetCurrentLobby().Members.Find(x => x.ProductId == targetProductUserId);
+            if (lobbyMember != null)
+            {
+                // Do not allow multiple local mute toggles at the same time
+                if (lobbyMember.RTCState.MuteActionInProgress)
+                {
+                    Debug.LogWarningFormat("Lobbies (MuteMember): 'MuteActionInProgress' for productUserId {0}.", targetProductUserId);
+                    MuteMemberCompleted?.Invoke(Result.RequestInProgress);
+                    return;
+                }
+
+                // Set mute action as in progress
+                lobbyMember.RTCState.MuteActionInProgress = true;
+            }
+
+            // Check if muting ourselves vs other member
+            if (EOSManager.Instance.GetProductUserId() == targetProductUserId)
+            {
+                // Toggle our mute status
+                UpdateSendingOptions sendOptions = new UpdateSendingOptions()
+                {
+                    LocalUserId = EOSManager.Instance.GetProductUserId(),
+                    RoomName = GetCurrentLobby().RTCRoomName,
+                    AudioStatus = isMicOn ? RTCAudioStatus.Enabled : RTCAudioStatus.Disabled
+                };
+
+                Debug.LogFormat("Lobbies (MuteMember): Setting self audio output status to {0}", sendOptions.AudioStatus == RTCAudioStatus.Enabled ? "Unmuted" : "Muted");
+
+                rtcAudioHandle.UpdateSending(ref sendOptions, MuteMemberCompleted, OnRTCRoomUpdateSendingCompleted);
+            }
+            else
+            {
+                // Toggle mute for remote member (this is a local-only action and does not block the other user from receiving your audio stream)
+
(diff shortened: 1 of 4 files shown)


andrew-hirata-playeveryware (approved):

Assets/Scripts/EOSLobbyManager.cs
```diff
@@ -48,6 +48,12 @@ public class Lobby
         public bool AllowInvites = true;
         public bool? DisableHostMigration;
 
+        /// <summary>
+        /// Has the local user deafened themself?
+        /// This doesn't carry over between rooms.
```
paulhazen: nit: Change to "Indicates whether the local user has deafened themselves (this doesn't carry over between rooms)."

Assets/Scripts/EOSLobbyManager.cs
```diff
@@ -1529,65 +1535,137 @@ public void PressToTalk(KeyCode PTTKeyCode, OnLobbyCallback TogglePressToTalkCom
...
         {
+            // If the user can be found in CurrentLobby.Members, set this value to the opposite of the user's current mute value
+            // Otherwise, always attempt to mute
+            bool shouldAudioBeEnabled = false;
+
             RTCInterface rtcHandle = EOSManager.Instance.GetEOSRTCInterface();
             RTCAudioInterface rtcAudioHandle = rtcHandle.GetAudioInterface();
 
-            foreach(LobbyMember lobbyMember in CurrentLobby.Members)
+            LobbyMember lobbyMember = GetCurrentLobby().Members.Find(x => x.ProductId == targetUserId);
```
paulhazen: Might need to use null propagation to check to make sure that `Members` is not null. Not certain about that though.

Assets/Scripts/EOSLobbyManager.cs
```diff
@@ -1529,65 +1535,137 @@ public void PressToTalk(KeyCode PTTKeyCode, OnLobbyCallback TogglePressToTalkCom
...
+        /// If the local user mutes themself, other users won't be able to hear the local user.
+        /// </summary>
+        /// <param name="shouldAudioBeEnabled">
+        /// If true, audio "becomes enabled", unmuting the target.
+        /// If false, audio "becomes disabled", muting the target.
+        /// </param>
+        /// </param>
+        /// <param name="targetProductUserId">The Product User ID of the target member to mute</param>
+        /// <param name="MuteMemberCompleted">Callback when MuteMember is completed</param>
+        public void MuteMember(bool shouldAudioBeEnabled, ProductUserId targetProductUserId, OnLobbyCallback MuteMemberCompleted)
```
paulhazen: I think contextually, it makes sense to rename this function signature to `SetIsMemberMuted`, or something that is not just `Mute`, since this function can actually _unmute_ a member, so the name is misleading as-is.

Assets/Scripts/EOSLobbyManager.cs
```diff
@@ -1529,65 +1535,137 @@ public void PressToTalk(KeyCode PTTKeyCode, OnLobbyCallback TogglePressToTalkCom
...
+        /// <summary>
+        /// Sets the Deafen Status of the Local Member.
+        /// If <paramref name="shouldBeDefeaned"/> is true, then the user won't be able to hear other users.
+        /// </summary>
+        /// <param name="shouldBeDefeaned">
+        /// If true, the local user should become defeaned.
+        /// Otherwise, undeafen the user.
+        /// </param>
+        /// <param name="DeafenCompleted">Callback when DeafenLocalMember is completed</param>
+        public void DeafenLocalMember(bool shouldBeDefeaned, OnLobbyCallback DeafenCompleted)
```
paulhazen: Probably change the function name similarly to the `MuteMember` function, so that it indicates that the deafen state can be either set or un-set.

paulhazen (approved):
Approved, with the caveat of the comments that I've left that need resolving.

Assets/Scripts/EOSLobbyManager.cs
WispyMouse: Implemented as suggested. (plus line break)

Assets/Scripts/EOSLobbyManager.cs
WispyMouse: That's a good idea, and I'll add it here, but I'll note there's a _lot_ of places that should really be doing nullchecking. It's unsafe to just fire off half the operations in this class before getting in to a good state. To be fixed in time.

Assets/Scripts/EOSLobbyManager.cs
WispyMouse: Imma try `SetMemberMuteStatus`.

Assets/Scripts/EOSLobbyManager.cs
WispyMouse: Did some renaming.

Assets/Scripts/EOSLobbyManager.cs
```diff
@@ -1529,65 +1535,136 @@ public void PressToTalk(KeyCode PTTKeyCode, OnLobbyCallback TogglePressToTalkCom
...
+        /// If the local user mutes themself, other users won't be able to hear the local user.
+        /// </summary>
+        /// <param name="shouldUserBeMuted">
+        /// If true, audio "becomes disabled", muting the target.
+        /// If false, audio "becomes enabled", unmuting the target.
+        /// </param>
+        /// </param>
+        /// <param name="targetProductUserId">The Product User ID of the target member to mute</param>
+        /// <param name="MuteMemberCompleted">Callback when MuteMember is completed</param>
+        public void SetMemberMuteStatus(bool shouldUserBeMuted, ProductUserId targetProductUserId, OnLobbyCallback MuteMemberCompleted)
```
paulhazen, line 1582: In light of the renamed function, the `MuteMemberCompleted` parameter should likewise be renamed (first for making it lowercase, second for the name) as:

```csharp
onMemberMuteStatusSet
```

arthur740212 (approved):

PR #884, opened 2024-09-02
(fix): Move `EventSystemManager` to `Assets/Scripts/` from `com.playeveryware.eos/Runtime` 

paulhazen:
Moves `EventSystemManager` into the `Assets/Scripts/` directory from the `com.playeveryware.eos/Runtime` directory, adding a namespace around it that corresponds to its proper location.

This is done out of the continuing effort to introduce more unit tests and is opened as a solitary PR to distinguish the change from one that actually introduces new Unit Tests (which will be forthcoming).

#EOS-2054


arthur740212 (approved):

WispyMouse (approved):

andrew-hirata-playeveryware (dismissed):
The file move isn't needed, but the namespace / code changes are good!

paulhazen:
> The file move isn't needed, but the namespace / code changes are good!

Just to follow up on this, I did a grep search within our current file structure (the one in `development`) to count how many files contain the namespace `PlayEveryWare.EpicOnlineServices.Samples` (which we agreed is the namespace that should be applied to `EventSystemManager`:

First against the `com.playeveryware.eos/Runtime` directory _specifically_:
```bash
grep -hr "PlayEveryWare.EpicOnlineServices.Samples" ./com.playeveryware.eos/Runtime/ | wc -l
```

I then expanded the search to the entire `com.playeveryware.eos` directory:
```bash
grep -hr "PlayEveryWare.EpicOnlineServices.Samples" ./Assets/ | wc -l
```

The results for both grep commands is 0. 

However, running the command against `Assets/` yields a result of `157`, which _strongly_ suggests to me that moving the file into that directory is the correct decision. 

@andrew-hirata-playeveryware does that change your thought process on this at all, and if not, can you help me understand why not?

andrew-hirata-playeveryware:
@paulhazen it does not change my mind because I haven't finished moving the files from samples into the new UPM layout based directory so most of them are still under assets.

PR #885, opened 2024-09-02
(fix): Remove unreferenced files

paulhazen:
This PR removes one file and its corresponding `.meta` file that are not being used or referenced anywhere. 

_This is another PR in the continuing effort to prepare to add more Unit Tests and is done preparatory to that so that when it is introduced that PR accomplishes the singular task of increasing code coverage._

#EOS-2066


WispyMouse (approved):
Approved.

`Assets/EOS/` gets put in `Runtime/EOS` in the built package. The location of the other prefab at `Assets/Plugins/EOSManager.prefab` will get put in `Runtime`. The users should be able to find both.

My main concern is that anyone who pulls the updated plugin, I *think* since this isn't sample code this could affect any number of users who are using the prefab. Upon updating the plugin, would the missing prefab cause their game to possibly break?

andrew-hirata-playeveryware (changes requested):
This needs to update documentation to remove ref to the EOSManager, and the linked task needs to be updated with a new task that encapsulates the work.

paulhazen:
> This needs to update documentation to remove ref to the EOSManager, and the linked task needs to be updated with a new task that encapsulates the work.

I have narrowed the scope of the PR to only remove the file that was unreferenced by both code and documentation. The correct placement of the `EOSManager.prefab` file, and the removal of its duplicate can be relegated to another PR. 

I have also updated the linked task to reflect the work being done.

andrew-hirata-playeveryware (approved):

PR #886, opened 2024-09-02
fix: EOSManager DontDestroyOnLoad, removal of EOSHostManager

WispyMouse:
Users of the plugin are required to keep EOSManager around, so that it can run Update statements. This is achieved through either putting the EOSManager singleton on every scene, or by marking EOSManager as DontDestroyOnLoad.

This changeset seeks to address that by making the EOSManager - 

1) Notice if there's an existing, alive EOSManager instance. If so, mark self as not enabled, so that Update etc aren't run.
2) If there isn't one, then always try to initialize.
3) Mark the EOSManager's GameObject as DontDestroyOnLoad, so that it sticks around in other scenes.

This does some possibly contentious things. In no particular order, here are some things to consider about this PR:

- I've decided to get rid of EOSHostManager, a supporting class and also member of the Singletons object. The purpose of this before was only to initialize the EOSSingleton. I believe this is better handled by EOSManager.
- EOSManager used to have the boolean to InitializeOnAwake. Instead, this now attempts to always initialize. I believe the InitializeOnAwake was only useful when EOSManager relied on EOSHostManager, which it doesn't need to.
- This deletes an assemblydefinition at `com.playeveryware.eos.samples.core`, because there are now no associated scripts. It had nothing referencing it.

Could also use a review on naming conventions and commentary.

#EOS-2061


com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -138,6 +135,48 @@ public delegate void OnAuthLinkExternalAccountCallback(
...
+        /// </summary>
+        private static EOSManager existingManager = null;
+
+        /// <summary>
+        /// Answers the question of whether a new EOSManager should be set up.
+        /// There shouldn't be two active and enabled EOSManagers, this helps prevent accidentally having two.
+        /// There will never be duplicate EOSSingletons, but duplicate EOSManager MonoBehaviours would
+        /// make update ticks run twice when it should run once.
+        /// </summary>
+        private static bool isThereAnExistingRunningManager
```
paulhazen: As a C# Property, even if it has a non-public access modifier it should be Pascal Case, meaning it ought to be `IsThereAnExistingRunningManager`.

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -138,6 +135,48 @@ public delegate void OnAuthLinkExternalAccountCallback(
         //cached log levels for retrieving later
         private static Dictionary<LogCategory, LogLevel> logLevels;
 
+        /// <summary>
+        /// This stores a reference to an existing EOSManager.
+        /// This should only be used to determine if a new copy of EOSManager should become not enabled.
+        /// </summary>
+        private static EOSManager existingManager = null;
```
paulhazen: As a `private static` field member, this should be prefixed with `s_` to indicate that it is static, and that it is private. Other field members within this file only do not follow this convention because their existence predates that convention, and a task was never defined to retro-actively apply the convention.

com.playeveryware.eos/Runtime/Core/EOSManager.cs
paulhazen: While the above comment is a correct application of the naming convention - I think that this property (moreover the functionality that its existence provides) would be better accomplished by a better implementation of the Singleton pattern - since that is what this property helps to accomplish.

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -1799,10 +1838,24 @@ static public EOSSingleton Instance
         /// </summary>
         void Awake()
         {
-            if (InitializeOnAwake)
+            // If the backing field s_instance already exists, and it has indicated it is running,
+            // then this EOSManager shouldn't perform any operations.
+            // So at this time if that condition is met, disable this behavior so that it doesn't run Update, etc
+            if (isThereAnExistingRunningManager)
```
paulhazen: What does the lifecycle of objects look like for the scenario where the manager is not null and is not running. Is there a way to make the existing non-running instance running?

paulhazen (commented):
I am a _HUGE_ fan of deleting files. No notes on that.

I think that the changes you made to `EOSManager` are both insightful and valuable, but I think it can be better implemented by making use of the Singleton pattern. This is the pattern I prefer to make use of, and I think it can be applied here:

```c#
private static readonly Lazy<Foo> s_lazyInstance = new(() => new Foo());

public static Foo Instance
{
    get
    {
        return s_lazyInstance.Value;
    }
}

private Foo() { }
```

WispyMouse:
> I am a _HUGE_ fan of deleting files. No notes on that.
> 
> I think that the changes you made to `EOSManager` are both insightful and valuable, but I think it can be better implemented by making use of the Singleton pattern. This is the pattern I prefer to make use of, and I think it can be applied here:
> 
> ```cs
> private static readonly Lazy<Foo> s_lazyInstance = new(() => new Foo());
> 
> public static Foo Instance
> {
>     get
>     {
>         return s_lazyInstance.Value;
>     }
> }
> 
> private Foo() { }
> ```

What thing are you suggesting to make a lazy loaded singleton? We *can* do that for the EOSManager, by making it create a GameObject, attaching the script to it, and marking as DontDestroyOnLoad. Is that preferable to having users put in a prefab?

com.playeveryware.eos/Runtime/Core/EOSManager.cs
WispyMouse: Noted, will follow this convention.

com.playeveryware.eos/Runtime/Core/EOSManager.cs
WispyMouse: Noted on naming convention, as well as questioning the need for this lifecycle management.

com.playeveryware.eos/Runtime/Core/EOSManager.cs
WispyMouse: So I think now that I've gotten rid of the "initialize on awake", this isn't really possible.

Before, you could have a EOSManager that is alive, but marked to not initialize things. However, that's pretty silly.

WispyMouse:
My takeaway from @paulhazen 's comments is to try and make the EOSManager lazy loaded. I originally was against it because prefabs *could* contain useful information, *but* this prefab doesn't. We can easily make a GameObject and attach it.

com.playeveryware.eos/Runtime/Core/EOSManager.cs
paulhazen: Agreed on the silliness 😆 

Sounds good, and makes sense to me in context.

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -320,7 +351,7 @@ public bool IsOverlayOpenWithExclusiveInput()
 
             //-------------------------------------------------------------------------
             [Conditional("ENABLE_DEBUG_EOSMANAGER")]
-            static void print(string toPrint)
+            internal static void print(string toPrint)
```
paulhazen: +1 good idea to make this `internal`.

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -1781,6 +1812,10 @@ static public EOSSingleton Instance
         {
             get
             {
+                // The EOSSingleton object requires EOSManager to be around and alive in order to run its updates
+                // So when the Instance is asked for, make sure there's an EOSManager
+                _ = ManagerInstance;
```
paulhazen: It is not clear to me how this line of code accomplishes what the comment above it describes.

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -138,6 +135,40 @@ public delegate void OnAuthLinkExternalAccountCallback(
...
+            EOSSingleton.print($"{nameof(EOSManager)} ({nameof(s_lazyManagerInstance)}): Lazy loaded singleton instance is creating a GameObject with an EOSManager attached.");
+
+            // EOSManager is a MonoBehaviour, so create a GameObject to hold it
+            GameObject instanceGameObjectHolder = new GameObject();
+
+            // Add the component
+            EOSManager addedManager = instanceGameObjectHolder.AddComponent<EOSManager>();
+
+            return addedManager;
+        });
```
paulhazen: A few thoughts on this:

* The `Lazy<T>` field member when used as the instance for a singleton class should use the `readonly` modifier. This is because the "Lazy Object" never changes, but _the thing it encapsulates_ does.
* If the intent is to utilize a singleton pattern, the constructor should be a `private` function, not an inline lambda function like it is here.

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -138,6 +135,40 @@ public delegate void OnAuthLinkExternalAccountCallback(
...
+
+        public static EOSManager ManagerInstance
```
paulhazen: I understand why you chose to name this `ManagerInstance` instead of `Instance`, but I think doing so reinforces a bad pattern that we are currently using. 

`EOSSingleton` is a bad name for a variety of reasons:
* Singletons should not contain the word `Singleton`
* Most of the functionality within `EOSManager` exists inside `EOSSingleton`
* There can be multiple instances of `EOSSingleton` (that is, there is nothing preventing anyone from creating more instances of it)

Ostensibly - despite utilizing vocabulary often associated with the singleton pattern - _neither_ `EOSManager` nor `EOSSingleton` _are_ following the singleton pattern.

The Correct™ way for this to be implemented is to move all the functions within `EOSSingleton` into `EOSManager`, and simply have `EOSManager` use the singleton pattern. 

Now, that is clearly outside of the scope of this PR but beginning to implement `ManagerInstance` _reinforces_ that poor implementation and will make it difficult in the future to correct properly.

paulhazen (changes requested):
I think that the bulk of these changes can be accomplished without altering the pseudo-singleton pattern that `EOSManager` currently employs. That pattern _should_ be corrected in my mind, but I disagree with the approach taken here regarding _how_ it is being changed. I have included more detailed feedback in the form of individual comments.

com.playeveryware.eos/Runtime/Core/EOSManager.cs
WispyMouse: It accesses the `get` part of the static property, which will resolve it and create the ManagerInstance if it needs to.

com.playeveryware.eos/Runtime/Core/EOSManager.cs
WispyMouse: The goal for not being readonly was that the `set` part of the `ManagerInstance` needs to be able to set an instance, in case the user is continuing to use the Singletons prefab object that has this behaviour attached. If it's readonly, I can't do that trick, and I don't think that `Lazy` has other ways of setting an instance like so.

Noted on the suggestion about not using the lambda. So a static private method, then?

com.playeveryware.eos/Runtime/Core/EOSManager.cs
WispyMouse: We talked in Slack about this; Ultimately this is right, we don't want to entrench ourselves further in to bad patterns.

andrew-hirata-playeveryware (approved):

Assets/Scripts/EOSLobbyManager.cs
```diff
@@ -473,7 +481,16 @@ public class EOSLobbyManager : IEOSSubManager
...
-        private List<Action> LobbyChangeCallbacks;
+
+
+        public delegate void LobbyChanged(string lobbyId, LobbyChangeType typeOfChange);
+
+        /// <summary>
+        /// Event that is run whenever the local user's relationship to a Lobby has been changed.
+        /// Indicates the Lobby that the change relates to.
+        /// </summary>
+        public event LobbyChanged OnLobbyChanged;
```
paulhazen: My thought process on this has evolved since my last review. Primarily due to [this](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members#names-of-events) documentation provided by Microsoft that outlines naming conventions surrounding events and delegates in C#.

My updated understanding of the pattern (as described in the link above, and applied to this circumstance) is that the following should be done:

```c#
public class LobbyChangeEventArgs
{
    public string LobbyId { get; }
    public LobbyChangeType { get; }
    
    public LobbyChangeEventArgs(string lobbyId, LobbyChangeType changeType)
    {
        LobbyId = lobbyId;
        LobbyChangeType = changeType;
    }
}

public delegate void LobbyChangeEventHandler(object sender, LobbyChangeEventArgs e);

public event LobbyChangedEventHandler LobbyChanged;

// Made virtual in case a deriving class wants to override it.
// Use this function to trigger the event from inside EOSLobbyManager so that it is threadsafe
protected virtual void OnLobbyChanged(LobbyChangeEventArgs args)
{
    LobbyChangedEventHandler handler = LobbyChanged;
    handler?.Invoke(this, args);
}
```

Assets/Scripts/StandardSamples/UI/Friends/UIFriendsMenu.cs
```diff
@@ -34,6 +34,12 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
     using Epic.OnlineServices.Friends;
     using Epic.OnlineServices.Presence;
 
+    /// <summary>
+    /// Provides an implementation of requesting friends from the EOS SDK, arranging, and displaying the information.
+    /// [!NOTE]: Only friends that have consented to the currently running application will appear when queried as friends.
```
paulhazen: Is this `[!NOTE]` something widely recognized by intellisense parsers or other documentation generating tools? I am only familiar with it as a feature within GitHub Flavored Markdown (support for this is introduced via a blog post [here](https://github.blog/changelog/2023-12-14-new-markdown-extension-alerts-provide-distinctive-styling-for-significant-content/)). If it's not something you are aware of widespread support for, I suggest changing it to simply `NOTE:`.

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -320,7 +325,7 @@ public bool IsOverlayOpenWithExclusiveInput()
 
             //-------------------------------------------------------------------------
             [Conditional("ENABLE_DEBUG_EOSMANAGER")]
-            static void print(string toPrint)
+            internal static void print(string toPrint)
```
paulhazen: I believe we discussed in our PR meeting renaming this to `log`, but I'm fuzzy on the details of the conversation - my recollection could be mistaken.

com.playeveryware.eos/Runtime/Core/Utility/FileSystemUtility.cs
```diff
@@ -507,10 +507,7 @@ public static async Task<string> ReadAllTextAsync(string path)
             // appropriate.
             try
             {
-                Debug.Log("Starting async file read");
-                var result = await File.ReadAllTextAsync(path);
-                Debug.Log("Finished async file read");
-                return result;
+                return await File.ReadAllTextAsync(path);
```
paulhazen: This change should not be a part of this PR - since it was made in a different branch that has since been merged in. This might be something that will go away when you update the base branch from `development`. If it doesn't go away though, then this is curious.

paulhazen (changes requested):

Assets/Scripts/StandardSamples/UI/Friends/UIFriendsMenu.cs
WispyMouse: I know this to be Markdown, which is something you've used in comments before. But I had _thought_ this pattern was already used in the code base, which seems to not be the case.

I'm comfortable changing these all to `NOTE:`.

WispyMouse: Imma do that in another PR.

Assets/Scripts/EOSLobbyManager.cs
WispyMouse: This has been brought over in to a separate PR: #890

Assets/Scripts/StandardSamples/UI/Friends/UIFriendsMenu.cs
WispyMouse: This has been brought in to #891

com.playeveryware.eos/Runtime/Core/Utility/FileSystemUtility.cs
WispyMouse: It went away after more rebasing.

com.playeveryware.eos/Runtime/Core/EOSManager.cs
WispyMouse: Polling people for opinions on Slack; but this is separate from the PR, only brought in from rebasing.

paulhazen (approved):

PR #887, opened 2024-09-03
fix: EOSManager to use print for all logs

WispyMouse:
Moves all EOSManager `Debug.Log` calls under the `EOSSingleton.print` functions, which are conditionally excluded from builds. Made a warning/error version, and gave them the internal accessor.

Bonus; Can anyone think of a better name than `print`, etc? It turns out that's already a function name for `MonoBehaviours`, and I had originally accidentally used that instead of the thing I meant to.

#EOS-1785

### DIFF
--- a/com.playeveryware.eos/Runtime/Core/EOSManager.cs
+++ b/com.playeveryware.eos/Runtime/Core/EOSManager.cs
@@ -325,6 +325,18 @@ namespace PlayEveryWare.EpicOnlineServices
                 Debug.Log(toPrint);
             }
 
+            [Conditional("ENABLE_DEBUG_EOSMANAGER")]
+            static void printError(string toPrint)
+            {
+                Debug.LogError(toPrint);
+            }
+
+            [Conditional("ENABLE_DEBUG_EOSMANAGER")]
+            static void printWarning(string toPrint)
+            {
+                Debug.LogWarning(toPrint);
+            }
+
             //-------------------------------------------------------------------------
             public void AddConnectLoginListener(IEOSOnConnectLogin connectLogin)
             {
--- a/com.playeveryware.eos/Runtime/Core/EOSManager.cs
+++ b/com.playeveryware.eos/Runtime/Core/EOSManager.cs
@@ -320,21 +320,9 @@ namespace PlayEveryWare.EpicOnlineServices
 
             //-------------------------------------------------------------------------
             [Conditional("ENABLE_DEBUG_EOSMANAGER")]
-            internal static void print(string toPrint)
+            internal static void print(string toPrint, LogType type = LogType.Log)
             {
                 Debug.Log(toPrint);
-            }
-
-            [Conditional("ENABLE_DEBUG_EOSMANAGER")]
-            internal static void printError(string toPrint)
-            {
-                Debug.LogError(toPrint);
-            }
-
-            [Conditional("ENABLE_DEBUG_EOSMANAGER")]
-            internal static void printWarning(string toPrint)
-            {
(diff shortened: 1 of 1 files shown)


com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -320,11 +320,23 @@ public bool IsOverlayOpenWithExclusiveInput()
...
+            internal static void printError(string toPrint)
+            {
+                Debug.LogError(toPrint);
+            }
+
+            [Conditional("ENABLE_DEBUG_EOSMANAGER")]
+            internal static void printWarning(string toPrint)
+            {
+                Debug.LogWarning(toPrint);
+            }
```
paulhazen: I'm not a huge fan of having different functions like this - I know that's how Unity does it though (ie `Debug.Log`, `Debug.LogError`, `Debug.LogWarning`). But I tend to prefer something like this (the level may already be an `enum` defined somewhere):

```c#
internal enum LogLevel
{
    Info,
    Warning,
    Error
}

[Conditional("ENABLE_DEBUG_[InsertClassNameHere]")]
internal static void log(string message, LogLevel level = LogLevel.Info)
{
    /* ... */
}
```

WispyMouse: Yup, there's already a `LogType` which comes with `Error`, `Assert`, `Warning`, `Log`, and `Exception`. So here, you'd prefer logging and setting that `LogType` instead?

Do you reckon there should be `logWarning` that calls the first method with `LogType.Warning`, or just specify the `LogType` when calling this each time that it isn't the default value (which should probably be `Log`)?

paulhazen: I think having the `logWarning` function call the `log` function makes sense. I tend to still prefer the `log` function being the only thing, but having to add the `LogLevel level` parameter each time it's not a `Log` seems equally silly, so it's probably good to have `logWarning`.

paulhazen: As an aside: I think this change highlights an important but often overlooked distinction between "log" and "print," in that not all logs are printed, and not all printed things are logs. That's a worthwhile correction to make in many areas of the code, but on its own would be a sprawling task that doesn't have any immediate benefits.

WispyMouse: Implemented as suggested.

paulhazen (approved):

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -320,7 +320,7 @@ public bool IsOverlayOpenWithExclusiveInput()
 
             //-------------------------------------------------------------------------
             [Conditional("ENABLE_DEBUG_EOSMANAGER")]
-            static void print(string toPrint)
+            internal static void print(string toPrint, LogType type = LogType.Log)
             {
                 Debug.Log(toPrint);
```
andrew-hirata-playeveryware: I think we should also use the LogType that is passed in 🛩️ 😄 .

andrew-hirata-playeveryware (changes requested):

com.playeveryware.eos/Runtime/Core/EOSManager.cs
WispyMouse: Implemented as suggested; oops on forgetting in the first place!

andrew-hirata-playeveryware (approved):

PR #889, opened 2024-09-04
Add: login with the used option.

skeyll:
I've implemented the processing inside ConfigureConnectExpirationCallback. 

It uses the options used during a successful login to re-login when the notification is triggered. 

The process of refreshing access by calling the login is described here: https://dev.epicgames.com/docs/ja/api-ref/functions/eos-connect-add-notify-auth-expiration.

The reason for implementing a new login process internally, instead of simply calling public void StartConnectLoginWithOptions(Epic.OnlineServices.Connect.LoginOptions connectLoginOptions, OnConnectLoginCallback onloginCallback) again, is that the actions required during the initial login and for a user during gameplay are likely to differ. 

Regarding changes in the user's login status, even if the re-login fails here, the user will remain valid for another 10 minutes, so changes will be handled via ConfigureConnectStatusCallback.

#EOS-2001

### DIFF
--- a/Assets/Scripts/BuiltSceneTestScripts/AuthenticationExpirationTestManager.cs
+++ b/Assets/Scripts/BuiltSceneTestScripts/AuthenticationExpirationTestManager.cs
@@ -23,6 +23,8 @@
 namespace PlayEveryWare.EpicOnlineServices.Tests
 {
     using Epic.OnlineServices;
+    using Epic.OnlineServices.Auth;
+    using Epic.OnlineServices.Connect;
     using EpicOnlineServices;
     using Samples;
     using Samples.Steam;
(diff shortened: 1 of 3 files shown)


WispyMouse:
Ahoy @skeyll ! I wanted to acknowledge your contribution here, thanks for the pull request! I can definitely agree, there should be _something_ that happens by default when the connect authentication is about to expire.

I'd like more time to chew on this before giving a proper review or integrating these changes. There are details that need arranging (like handling fail state), and I need to consider if re-using the Credentials will succeed. I'll have to test it to be sure that, for example, that credentials used for connect can be submitted again without re-upping the refresh token or anything.

It could be that the fate of this is to have it call `StartConnectLoginWithEpicAccount` whenever the expiration is coming up... I'll get back to you soon!

paulhazen:
Oops! I made a "faux paus" here by merging the `development` branch into your branch - something that I often do when going through PR requests and after the `development` branch has changes merged into it. Hopefully I didn't cause any confusion - but apologies nonetheless.

WispyMouse:
@skeyll Heya! May I have your permission to change code on your repository, and push it to this PR?

skeyll:
Thank you for checking.
@paulhazen. I understand, no problem at all!

@WispyMouse, of course! Feel free to make any changes you'd like.

WispyMouse:
@skeyll Hmm, I can't seem to push to your branch. I think you might have to give me some kind of permission on your repo, can you please check that?

skeyll:
@WispyMouse I invite you as a collaborator now, sorry! Let me know if you still have any issues.

paulhazen:
Hi @skeyll, I added you as an "assignee" to this PR for a couple of reasons:

1. It helps keep you appraised of the progress as it is shepherded through our PR process.
2. Once merged - you will become a contributor of record on the project (adding you to the proverbial "credits" in the parlance of open source projects).

If you prefer not to be an assignee that is of course totally fine - I just wanted to give you context and offer to re-add you if you would like :)

Assets/Scripts/BuiltSceneTestScripts/AuthenticationExpirationTestManager.cs
```diff
@@ -57,20 +59,75 @@ public enum AuthenticationProviderToTest
...
         {
+            if (RunningTestCoroutine != null)
+            {
+                Debug.Log($"{nameof(AuthenticationExpirationTestManager)} ({nameof(SetAuthenticationProviderAndLogin)}): Test is already running, cannot run multiple tests in the same instance.");
+                return;
+            }
+
             RunningTest = toSet;
-            StartProbe();
+            StartAuthenticationExpirationProbe();
```
paulhazen, line 103: In an earlier change within this file, you moved away from using the word `Probe`, is this function's rename consistent with that change?

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -1176,11 +1176,16 @@ public void StartConnectLoginWithOptions(Epic.OnlineServices.Connect.LoginOption
                 connectInterface.Login(ref connectLoginOptions, null,
                     (ref Epic.OnlineServices.Connect.LoginCallbackInfo connectLoginData) =>
                     {
+                        if (connectLoginData.ResultCode != Result.Success)
+                        {
+                            print($"Connect login was not successful. ResultCode: {connectLoginData.ResultCode}", LogType.Error);
+                        }
```
paulhazen, line 1182: I recall an earlier conversation on this topic, but I do not recall the outcome: I believe that it's possible that a result code can be returned indicating that the user is already logged in (I believe the result code is something like `AlreadyConfigured`). _If that is the case_, then we need to change this conditional to be:

```c#
if (connectLoginData.ResultCode != Result.Success && connectLoginData.ResultCode != Result.AlreadyConfigured)
{
    /* ... */
}
```

paulhazen (commented):
From what I can tell this only adds tests, and modestly changes `EOSManager`. I am not seeing the changes that the title of this PR would imply exist. Is there context I am missing?

Assets/Scripts/BuiltSceneTestScripts/AuthenticationExpirationTestManager.cs
WispyMouse: There's now two separate tests, so I renamed this one to be more clear which test this applied to. `StartConnectTest` and `SetAuthenticationProviderAndLogin` are separate tests, and use separate inner functions.

com.playeveryware.eos/Runtime/Core/EOSManager.cs
WispyMouse: I'll create a situation that reproduces this message. The documentation mentions error codes, but not which are specifically used here.

WispyMouse:
@paulhazen The title is implying, I think, "login with the used option" meaning "login with the (previously) used option (when notified that connect expiration is happening soon)".

The changed function is `StartConnectLoginWithOptions`. The change is subscribing to the notification that connect is about to expire (this subscription existed before) by adding `StartConnectLoginWithOptions` again, using the previous log in information options. There's some change to the signatures of the functions to retain a reference to the options.

skeyll:
@paulhazen 
Oops, it seems I accidentally unassigned myself! There was no deeper intention behind it. If possible, please feel free to add me again.


As WispyMouse mentioned, this request implements the process of logging in again within ConfigureConnectExpirationCallback. 

While the notification already existed in StartConnectLoginWithOptions, the process was empty. Therefore, I made it so that the login process is called again within ConfigureConnectExpirationCallback, which is notified 10 minutes before the connection expires.

 At that time, instead of creating new options during re-login for the various login way, I pass only the options to the notify that were used when the login was successful . 

I made this request because there have been occasional posts in the issue about expiration occurring every hour.

andrew-hirata-playeveryware (approved):

paulhazen (approved):

PR #890, opened 2024-09-04
fix(lobby): Use LobbyChangeEventArgs for events

WispyMouse:
This PR implements a suggested change from @paulhazen in relation to the LobbyChange. Here's the original comment:

> My thought process on this has evolved since my last review. Primarily due to [this](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members#names-of-events) documentation provided by Microsoft that outlines naming conventions surrounding events and delegates in C#.
> 
> My updated understanding of the pattern (as described in the link above, and applied to this circumstance) is that the following should be done:
> 
> ```c#
> public class LobbyChangeEventArgs
> {
>     public string LobbyId { get; }
>     public LobbyChangeType { get; }
>     
>     public LobbyChangeEventArgs(string lobbyId, LobbyChangeType changeType)
>     {
>         LobbyId = lobbyId;
>         LobbyChangeType = changeType;
>     }
> }
> 
> public delegate void LobbyChangeEventHandler(object sender, LobbyChangeEventArgs e);
> 
> public event LobbyChangedEventHandler LobbyChanged;
> 
> // Made virtual in case a deriving class wants to override it.
> // Use this function to trigger the event from inside EOSLobbyManager so that it is threadsafe
> protected virtual void OnLobbyChanged(LobbyChangeEventArgs args)
> {
>     LobbyChangedEventHandler handler = LobbyChanged;
>     handler?.Invoke(this, args);
> }
> ```

This was implemented almost exactly as suggested, including position in the file. It'd make sense if we want to rearrange these, let me know.

My main thought is that now that the event is meant to be fired through OnLobbyChanged, we might want to reintroduce a public function for subscribing/unsubscribing, and make the event not public?

### DIFF
--- a/Assets/Scripts/EOSEACLobbyManager.cs
+++ b/Assets/Scripts/EOSEACLobbyManager.cs
@@ -62,7 +62,7 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
 
             if (AntiCheatManager.IsAntiCheatAvailable())
             {
-                LobbyManager.OnLobbyChanged += OnLobbyChanged;
+                LobbyManager.LobbyChanged += OnLobbyChanged;
                 LobbyManager.AddNotifyLobbyUpdate(OnLobbyUpdated);
                 LobbyManager.AddNotifyMemberUpdateReceived(OnMemberUpdated);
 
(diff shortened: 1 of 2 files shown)


arthur740212 (approved):

paulhazen (approved):
The helper function is only used inside the class because of how external classes can publicly subscribe to and unsubscribe from the event.

What it protects against is having a client subscribe to `LobbyChanged` while `LobbyChanged` is being invoked.

PR #892, opened 2024-09-04
Break up `IPlatformSpecifics.cs` contents into individual files

paulhazen:
In the course of implementing the config overrides, namespace changes were appropriate, but they were complicated by all the definitions being contained within the same file. They are separated here preparatory to making those changes.

#EOS-1982


WispyMouse (approved):
Seems cleanly the same. Ran it in the Editor and had no problems. All the namespaces seem right and preserved.

andrew-hirata-playeveryware (approved):

PR #893, opened 2024-09-04
String Extensions

paulhazen:
In the process of implementing config overrides, it made sense to relegate the conversion of strings to `ulong` with a default value to a `StringExtensions` class. This change is put into a separate PR so that the PR containing the changes for the attached task can be evaluated in isolation.

#EOS-1982

### DIFF
--- a/com.playeveryware.eos/Runtime/Core/Extensions/StringExtensions.cs
+++ b/com.playeveryware.eos/Runtime/Core/Extensions/StringExtensions.cs
@@ -24,7 +24,7 @@ namespace PlayEveryWare.EpicOnlineServices.Extensions
 {
     public static class StringExtensions
     {
-        public static ulong ToUInt64(this string value, ulong defaultValue = 0L)
+        public static ulong ToUlong(this string value, ulong defaultValue = 0L)
         {
             ulong returnValue = defaultValue;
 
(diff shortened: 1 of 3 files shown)


com.playeveryware.eos/Runtime/Core/Extensions/StringExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+namespace PlayEveryWare.EpicOnlineServices.Extensions
+{
+    public static class StringExtensions
+    {
+        public static ulong ToUInt64(this string value, ulong defaultValue = 0L)
```
WispyMouse: What motivated changing the name referencing `ULong` to `UInt64`? It's still returning the `ulong` data type.

WispyMouse (approved):
Seems good to me. I tried to look up if there was any difference between `ulong` and `uint64`, but it appears as though C# treats them effectively identically.

com.playeveryware.eos/Runtime/Core/Extensions/StringExtensions.cs
paulhazen: Let's change this to `Ulong` to be consistent with areas of the EOS SDK.

andrew-hirata-playeveryware (approved):

PR #895, opened 2024-09-05
fix: Relocate `ConfigFieldAttribute` and break into disparate components

paulhazen:
This PR relocates `ConfigFieldAttribute` into the Runtime directory and breaks them into separate files. This work is done preparatory to work being done on the config override system. Specifically, the `ConfigFieldAttribute` and related classes must be accessible from within the Runtime, no longer just within the editor.

#EOS-1982


paulhazen (commented):
We should change the `Meta` directory to `Attributes`.

andrew-hirata-playeveryware (approved):

paulhazen:
> We should change the `Meta` directory to `Attributes`.

This has been done.

com.playeveryware.eos/Runtime/Core/Config/Attributes/ConfigFieldAttribute.cs
```diff
@@ -0,0 +1,60 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+namespace PlayEveryWare.EpicOnlineServices
+{
+    using System;
+
+    [AttributeUsage(AttributeTargets.Field)]
+    public class ConfigFieldAttribute : Attribute
```
WispyMouse, line 28: Would there be benefit in making this abstract?

WispyMouse (approved):
No issues detected, everything seems to work in the editor.

com.playeveryware.eos/Runtime/Core/Config/Attributes/ConfigFieldAttribute.cs
paulhazen: No, because the ConfigFieldAttribute class itself is used in many circumstances (actually in most of them):

Here are some assorted examples from the `development` branch.

https://github.com/PlayEveryWare/eos_plugin_for_unity/blob/2a18ec7e3fddff1c35baf37195a9d7989be1c4bd/Assets/Plugins/Source/Editor/Configs/ToolsConfig.cs#L59

https://github.com/PlayEveryWare/eos_plugin_for_unity/blob/2a18ec7e3fddff1c35baf37195a9d7989be1c4bd/Assets/Plugins/Source/Editor/Configs/LibraryBuildConfig.cs#L36

https://github.com/PlayEveryWare/eos_plugin_for_unity/blob/2a18ec7e3fddff1c35baf37195a9d7989be1c4bd/Assets/Plugins/Source/Editor/Configs/SigningConfig.cs#L51

It's worth noting that C# will _specifically for attribute classes_ allow you to omit the `Attribute` component of the class name when decorating things with your custom attribute. This can be a point of confusion for people who are less familiar with attributes.

PR #897, opened 2024-09-05
Change `FileSystemUtility.GetProjectPath()` to not use `Application.dataPath`

paulhazen:
This PR addresses an issue currently existing within the `development` branch that causes all the unit tests that derive from `EOSTestBase` to fail. The cause of this was access of the `Application.dataPath` property from within a thread other than the "main" Unity Thread.

This PR sidesteps that limitation by using a different means by which to infer the location of the project directory.

Thanks to @WispyMouse for documenting this issue and helping us find a solution.

 #EOS-2084

### DIFF
--- a/com.playeveryware.eos/Runtime/Core/Utility/FileSystemUtility.cs
+++ b/com.playeveryware.eos/Runtime/Core/Utility/FileSystemUtility.cs
@@ -413,6 +413,7 @@ namespace PlayEveryWare.EpicOnlineServices.Utility
             }
         }
 
+#if UNITY_EDITOR
         /// <summary>
         /// Returns the root of the Unity project.
         /// </summary>
@@ -433,7 +434,7 @@ namespace PlayEveryWare.EpicOnlineServices.Utility
             throw new DirectoryNotFoundException("Unable to locate the Assets folder from the current directory.");
         }
 
-#if UNITY_EDITOR
+
         #region Line Ending Manipulations
 
         public static void ConvertDosToUnixLineEndings(string filename)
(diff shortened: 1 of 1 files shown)


com.playeveryware.eos/Runtime/Core/Utility/FileSystemUtility.cs
```diff
@@ -419,7 +419,18 @@ await Task.Run(() =>
         /// <returns>Fully-qualified file path to the root of the Unity project.</returns>
         public static string GetProjectPath()
         {
-            return Path.GetFullPath(Path.Combine(Application.dataPath, "../"));
+            // Assuming the current directory is within the project (e.g., in the Editor or during Play mode)
```
WispyMouse, line 423: This should have UNITY_EDITOR around it if it is only editor-time.

WispyMouse: Determine if a built UPM package will still resolve correctly here.

WispyMouse (changes requested):
Request change in regards to UNITY_EDITOR and for validation that the built UPM package works.

com.playeveryware.eos/Runtime/Core/Utility/FileSystemUtility.cs
paulhazen: I have moved the `UNITY_EDITOR` compile-time conditional and confirmed for certain that this functions properly when built as a UPM package.

WispyMouse (approved):

andrew-hirata-playeveryware (approved):

PR #898, opened 2024-09-05
fix(config,IO) : Add missing if statement

arthur740212:
Without the if statement, it throws an exception always.
EOS-2087

### DIFF
--- a/com.playeveryware.eos/Runtime/Core/Config/Config.cs
+++ b/com.playeveryware.eos/Runtime/Core/Config/Config.cs
@@ -290,18 +290,18 @@ namespace PlayEveryWare.EpicOnlineServices
         {
             bool fileExists = await FileSystemUtility.FileExistsAsync(FilePath);
 
-#if UNITY_EDITOR
             if (!fileExists)
             {
+#if UNITY_EDITOR
                 await WriteAsync();
-            }
 #else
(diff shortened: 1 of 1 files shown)


WispyMouse (approved):

paulhazen (changes requested):
I believe that a more correct implementation of this fix would be as follows. The following example maintains the singular conditional, but changes what happens if `!fileExists` evaluates to `false` based on the compile time conditional `UNITY_EDITOR` and provides a comment explaining the purpose of the compile time conditional.

This example would be to replace the contents of the file indicated from lines `293` to `307` as they currently exist in the `development` branch.
```c#
if (!fileExists)
{
    // This compile time conditional is here because if the file 
    // does not exist in the editor, then it should be written, 
    // whereas in a non-editor context the absence of a config file
    // warrants throwing an exception.
#if UNITY_EDITOR
    await WriteAsync();
#else
    // If the editor is not running, then the config file not
    // existing should throw an error.
    throw new FileNotFoundException(
        $"Config file \"{FilePath}\" does not exist.");
#endif
}
```

arthur740212:
Updated

paulhazen (approved):

andrew-hirata-playeveryware (approved):

PR #903, opened 2024-09-10
tests(PlayerDataStorage)

WispyMouse:
This absorbs all of the tests made in https://github.com/PlayEveryWare/eos_plugin_for_unity/pull/836 , brought over to the retooled Manager service thing.


Assets/Scripts/EOSPlayerDataStorageManager.cs
```diff
@@ -483,5 +483,11 @@ private void OnFileRemoved(ref DeleteFileCallbackInfo data)
 
             QueryFileList();
         }
+
+        public override void RemoveCallbacks()
+        {
+            base.RemoveCallbacks();
+            OnFileListUpdated = null;
+        }
```
paulhazen: A few things with respect to this:
- It does not appear that the `RemoveCallbacks` function (either the base implementation or the overridden one) is ever called.
- In general, `Action` and `event` field members do not need to be set to null - because the relationship between them and the "clients" who subscribe to them is such that no error should occur if the action goes away - the client just doesn't get called. The client doesn't store a reference to the callback that is at risk of causing a problem.

For these reasons, I believe that this function (and the function that it overrides) can be removed.

com.playeveryware.eos/Runtime/Core/EOS_SDK_Additions/PlayerDataStorageFileTransferRequestWrapper.cs
```diff
@@ -53,7 +53,8 @@ public override Result CancelRequest()
 
         public override void Release()
         {
-            _instance.Release();
+            _instance?.Release();
+            _instance = null;
```
paulhazen: I believe that you should rebase / merge `development` into this branch and this change will go away.

Assets/Tests/PlayMode/PlayerDataStorageManagerTests.cs
```diff
@@ -0,0 +1,326 @@
...
+
+            for (int fileIndex = 0; fileIndex < numberOfFilesToMake; fileIndex++)
+            {
+                // Generate a random name until we hit something that isn't already taken
+                // Having the same name would work as an overwrite
+                string randomName = "";
+                do
+                {
+                    randomName = UnityEngine.Random.Range(0, 9999).ToString();
+                } while (randomNames.Contains(randomName));
```
paulhazen: I think it would be far simpler to use `Guid::New()` (or whatever the signature is). That way it is clearer, far less likely to collide, and all the file names will be of the same length.

paulhazen (changes requested):
I believe that there are some minor things that need to be addressed, but overall excellent addition to our Unit Tests. 

Before merging this in, I think that you should re-run the code coverage tool so we can see what the increased coverage looks like.

com.playeveryware.eos/Runtime/Core/EOS_SDK_Additions/PlayerDataStorageFileTransferRequestWrapper.cs
WispyMouse: Merged it in, then rebased. This is out of the PR now.

Assets/Scripts/EOSPlayerDataStorageManager.cs
WispyMouse: Oh! That's a relief, okay. I had always thought Actions needed to be cleaned up in this way.

These are supposed to be called during the teardown of the test. Looks like it is there. Will remove that part of the teardown, and remove this function.

Assets/Tests/PlayMode/PlayerDataStorageManagerTests.cs
WispyMouse: Sure, that's really simple and easy to do. GUID collisions should never happen.

paulhazen (approved):

andrew-hirata-playeveryware (approved):

arthur740212 (approved):

PR #905, opened 2024-09-10
More Unit Tests

paulhazen:
This PR introduces unit tests that expand code coverage. 

#EOS-2094


andrew-hirata-playeveryware (approved):

Assets/Tests/PlayMode/Extensions/FileInfoExtensionsTests.cs
WispyMouse, line 1: This file is currently a PlayMode test. Since it doesn't rely on any play mode features, could this be moved to be an EditMode test? Or does being a PlayMode test expose it in a way that lends towards later being able to do automated test runs?

Assets/Tests/PlayMode/Extensions/ListExtensionsTests.cs
```diff
@@ -0,0 +1,111 @@
...
+        /// </summary>
+        [Test]
+        public void Shuffle_ChangesOrderOfElements()
+        {
+            // Arrange
+            List<int> originalList = Enumerable.Range(1, 100).ToList();
+            List<int> shuffledList = new(originalList);
+
+            // Act
+            shuffledList.Shuffle();
```
WispyMouse, line 44: funny nit: This could *theoretically* result in a list that shuffles itself perfectly back in to place. HOWEVER, Wolfram Alpha says this is "1 in 100!", which means `1` in `93326215443944152681699238856266700490715968264381621468592963895217599993229915608941463976156518286253697920827223758251185210916864000000000000000000000000` chance of that happening.

Assets/Tests/PlayMode/Extensions/ListExtensionsTests.cs
```diff
@@ -0,0 +1,111 @@
...
+
+            // Assert
+            Assert.AreNotEqual(originalList, shuffledList,
```
WispyMouse, line 47: not review: I wonder if this works as intended? Does this just compare the pointer to the originalList and the shuffledList?

So I made a tiny test;

```c#
        [Test]
        public void AreEqualTest()
        {
            List<int> originalList = Enumerable.Range(1, 10).ToList();
            List<int> otherList = Enumerable.Range(1, 10).ToList();
            Assert.AreNotEqual(originalList, otherList);
        }
```

This test failed; the two arrays are equal, even though they're different pointers. So there's probably some equality operator being run. So you have it correct here!

ProjectSettings/Packages/com.unity.testtools.codecoverage/Settings.json
```diff
@@ -6,7 +6,7 @@
             {
                 "type": "System.String, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089",
                 "key": "IncludeAssemblies",
-                "value": "{\"m_Value\":\"com.playeveryware.eos,com.playeveryware.eos.core,com.playeveryware.eos.samples,com.playeveryware.eos.samples.appleauth,com.playeveryware.eos.samples.appleauth.editor,com.playeveryware.eos.samples.core,com.playeveryware.eos.samples.facepunch-steamworks,com.playeveryware.eos.samples.networking,com.playeveryware.eos.samples.oculus,com.playeveryware.eos.samples.postprocessing,com.playeveryware.eos.samples.steam,com.playeveryware.eos-Editor,com.playeveryware.eos-Editor.steam.utility\"}"
+                "value": "{\"m_Value\":\"com.playeveryware.eos,com.playeveryware.eos.core,com.playeveryware.eos-Editor,com.playeveryware.eos-Editor.steam.utility\"}"
```
WispyMouse, line 9: Should we have code coverage for the samples namespace? I was making tests for the Manager/Services.

WispyMouse (approved):
Looks good. Only thought it to consider moving these to EditMode instead of PlayMode, although there's potential flaws with that. For example, can we even run both PlayMode and EditMode tests in the same breath in a way that code coverage understands?

ProjectSettings/Packages/com.unity.testtools.codecoverage/Settings.json
paulhazen: Eventually we should, but the priority is the non-sample code.

Assets/Tests/PlayMode/Extensions/FileInfoExtensionsTests.cs
paulhazen: Moving it to EditMode might make sense. I'll look into that as I write unit test documentation.

PR #906, opened 2024-09-11
Expansion of config attribute usage

paulhazen:
Preparatory to introducing the ability to validate configuration values and override them in a clear manner through the use of overrides on a per-platform basis, this PR adds (as-of-yet unreferenced) `ConfigFieldAttribute`s to each of the fields within `EOSConfig`.

Adjacent to these changes, several unreferenced editor-centric and unused functions were removed from the utility `GUIEditorUtility`, while a new one for `double` was added:

- `AssigningUintField`
- `AssigningULongField`
- `AssigningPath`

#EOS-1982

### DIFF
--- a/Assets/Plugins/Source/Editor/ConfigEditors/ConfigEditor.cs
+++ b/Assets/Plugins/Source/Editor/ConfigEditors/ConfigEditor.cs
@@ -221,7 +221,7 @@ namespace PlayEveryWare.EpicOnlineServices.Editor
                 float labelWidth = GetMaximumLabelWidth(fieldGroup);
 
                 // If there is a label for the field group, then display it.
-                if (_groupLabels?.Length > fieldGroup.Key)
+                if (0 >= fieldGroup.Key && _groupLabels?.Length > fieldGroup.Key)
                 {
                     GUILayout.Label(_groupLabels[fieldGroup.Key], EditorStyles.boldLabel);
                 }
--- a/Assets/Plugins/Source/Editor/ConfigEditors/ConfigEditor.cs
+++ b/Assets/Plugins/Source/Editor/ConfigEditors/ConfigEditor.cs
@@ -32,8 +32,6 @@ namespace PlayEveryWare.EpicOnlineServices.Editor
     using UnityEngine.Events;
     using Utility;
     using Task = System.Threading.Tasks.Task;
-    using Config = EpicOnlineServices.Config;
-
 
     /// <summary>
     /// Contains implementations of IConfigEditor that are common to all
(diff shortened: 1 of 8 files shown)


com.playeveryware.eos/Runtime/Core/Config/EOSConfig.cs
```diff
@@ -200,12 +327,14 @@ protected EOSConfig() : base("EpicOnlineServicesConfig.json") { }
...
+            ConfigFieldType.Flag, 
+            "Workaround to force send input without any delay. If " +
+            "true, the native plugin will always send input receive directly " +
+            "to the SDK.", 4)]
         public bool hackForceSendInputDirectlyToSDK;
 
-        /// <summary>
-        /// Set to 'true' if the application is a dedicated game server.
-        /// </summary>
-        public bool isServer;
```
paulhazen, line 208: while this _looks_ to be removed, it was in-fact **moved** to an earlier place within the class definition.

Assets/Plugins/Source/Editor/Utility/GUIEditorUtility.cs
```diff
@@ -96,94 +96,6 @@ public static void AssigningTextField(string label, ref string value, float labe
...
-                }
-            }
-
-            if (horizontalLayout)
-            {
-                EditorGUILayout.EndHorizontal();
-            }
-        }
-
-        private delegate T InputRenderDelegate<T>(string label, T value, float labelWidth, string tooltip);
```
paulhazen, line 125: This was relocated to another place within the class definition that makes more sense.

com.playeveryware.eos/Runtime/Core/Config/RuntimeConfig.cs
```diff
@@ -20,7 +20,7 @@
  * SOFTWARE.
  */
 
-//#define EOS_RUNTIME_NEW_CONFIG_SYSTEM
+#define EOS_RUNTIME_NEW_CONFIG_SYSTEM
```
paulhazen: This should be commented out.

arthur740212 (approved):

Assets/Plugins/Source/Editor/ConfigEditors/ConfigEditor.cs
```diff
@@ -214,11 +216,16 @@ private static float GetMaximumLabelWidth(IEnumerable<(FieldInfo, ConfigFieldAtt
...
         protected void RenderConfigFields()
         {
-            var fieldGroups = GetFieldsByGroup();
-            foreach (var fieldGroup in fieldGroups)
+            foreach (var fieldGroup in GetFieldsByGroup())
             {
                 float labelWidth = GetMaximumLabelWidth(fieldGroup);
 
+                // If there is a label for the field group, then display it.
+                if (_groupLabels?.Length > fieldGroup.Key)
```
WispyMouse: silly nit that you don't really need to address; if fieldGroup.Key is negative, this'd break. Sometimes people put "-1" in Unity ordering things to automatically sort them.

paulhazen: Excellent point! This actually could have caused a problem, because the default value if the group number is not set is actually `-1`, so I've updated the implementation to include that in the conditional.

WispyMouse (approved):

PR #907, opened 2024-09-11
fix: Rename managers that have been refactored into services

paulhazen:
This PR completes the transition of the recently refactored storage managers from being "managers" to being "services," by affecting the following changes:

- Rename `EOSPlayerDataStorageManager` to `PlayerDataStorageService`
- Rename `EOSTitleStorageManager` to `TitleStorageService`
- Rename `DataService` to `StorageService` (so as to match the suffixes of the two aforementioned classes)
- Move `TitleStorageService` and `PlayerDataStorageService` into the appropriate `Services` subdirectory, where the other "service" classes reside.
- In keeping with the patterns of use for existing "Service" classes, implement the singleton pattern for both `PlayerDataStorageService`, and `TitleStorageService`.

The reason that this PR is a draft is twofold:
1. It should be done _after_ a pending [PR](https://github.com/PlayEveryWare/eos_plugin_for_unity/pull/903) that adds Unit Testing for some of these classes (and after that PR is ultimately merged, these renames should be applied to those unit tests as well)
2. The "tasking" for this needs to be established. Ideally, this would have been a component of the task that has already been marked as complete that refactors the storage samples, but it was overlooked.


WispyMouse:
The other existing Service has a lazy-loaded Singleton accessed through `Instance`, but these renamed managers still use the GetOrCreateManager. Is there a specific rule for which to use in which case?

paulhazen:
> The other existing Service has a lazy-loaded Singleton accessed through `Instance`, but these renamed managers still use the GetOrCreateManager. Is there a specific rule for which to use in which case?

Good point. I suppose the transition is not quite complete. I will update this PR to include implementation of the singleton pattern for each of these, provided there is no objection from the team in doing so.

Assets/Tests/PlayMode/PlayerDataStorageManagerTests.cs
```diff
@@ -39,13 +39,9 @@ namespace PlayEveryWare.EpicOnlineServices.Tests.IntegrationTests
 
     public class PlayerDataStorageManagerTests : EOSTestBase
     {
-        EOSPlayerDataStorageManager playerDataStorageManager;
-
         [UnitySetUp]
         public IEnumerator UnitySetUp()
         {
-            playerDataStorageManager = new EOSPlayerDataStorageManager();
```
WispyMouse, line 47: We have the singleton, so we don't need to have this new-ing anymore. This comment is to observe that we do lose something, and gain something. Both of which are, there are no guarantees that the Manager will be entirely reset between runs.

This can be useful because if a test randomly breaks because of this new situation, then that's evidence we should fix something.

Assets/Scripts/StandardSamples/UI/Storage/Title/UITitleStorageMenu.cs
```diff
@@ -46,7 +46,7 @@ public class UITitleStorageMenu : SampleMenu
         public GameObject UIFileNameEntryPrefab;
 
         public Text FileContent;
-        private EOSTitleStorageManager TitleStorageManager;
+        //private TitleStorageService TitleStorageManager;
```
WispyMouse: Can this be removed now that it isn't needed?

Assets/Scripts/StandardSamples/UI/Storage/Title/UITitleStorageMenu.cs
```diff
@@ -64,13 +64,12 @@ protected override void Awake()
...
         private void Start()
         {
-            TitleStorageManager = EOSManager.Instance.GetOrCreateManager<EOSTitleStorageManager>();
         }
 
         protected override void OnDestroy()
         {
             base.OnDestroy();
-            EOSManager.Instance.RemoveManager<EOSTitleStorageManager>();
+            EOSManager.Instance.RemoveManager<TitleStorageService>();
```
WispyMouse: Should this be removed through this way, now that we use Singletons?

Assets/Scripts/StandardSamples/UI/Storage/Player/UIPlayerDataStorageMenu.cs
```diff
@@ -89,12 +87,12 @@ private void Start()
         protected override void OnDestroy()
         {
             base.OnDestroy();
-            EOSManager.Instance.RemoveManager<EOSPlayerDataStorageManager>();
+            EOSManager.Instance.RemoveManager<PlayerDataStorageService>();
```
WispyMouse: Can this be removed now that we're using Singleton?

WispyMouse (approved):
Approved. Can we remove the .RemoveManager calls?

Assets/Scripts/StandardSamples/UI/Storage/Player/UIPlayerDataStorageMenu.cs
paulhazen: Yep, done.

Assets/Scripts/StandardSamples/UI/Storage/Title/UITitleStorageMenu.cs
paulhazen: Yep, done.

Assets/Scripts/StandardSamples/UI/Storage/Title/UITitleStorageMenu.cs
paulhazen: Yup, done.

Assets/Tests/PlayMode/PlayerDataStorageManagerTests.cs
paulhazen: This is a good point - but it is worth noting that the Services will _always_ be reset whenever the state of authentication changes - which should be the same thing. That is to say - the "state" of a service is not usually reset _except_ in circumstances where the currently authenticated user becomes no longer authenticated.

WispyMouse: It certainly should! It's just possible now to miss some reset thing. Which the test would find, and we would fix. So seems good to me!

andrew-hirata-playeveryware (approved):

PR #908, opened 2024-09-12
test: Auth/Connect while already logged in produces expected results

WispyMouse:
Paul had a suspicion that #889 might need to handle a user scenario where the system is already authenticated. To attempt to determine how this works, we made these two very simple tests.

Auth login, then try to Auth login. The result should be success.

Connect login, then try to Connect login. The result should be success.

Added a reference to the -Editor assembly in PlayMode tests so that it can access UnitTestConfig. If this isn't appropriate, we can try to address that.

#EOS-2113


andrew-hirata-playeveryware (approved):

paulhazen (approved):
Looks good to me!

PR #912, opened 2024-09-17
Improve Unit Test Organization

paulhazen:
This is the first of several PRs aimed at providing documentation that is consistent with our approach to Unit Tests. Some of this larger effort will be (as this PR is) making the Unit Tests conform to a consistent approach that documentation can be applied to.

This PR:
- Organizes existing tests into appropriate subdirectories 
- Renames `PlayerDataStorageManagerTests.cs` to `PlayerDataStorageServiceTests.cs`
- Removes unnecessary using statements.

#EOS-1903


WispyMouse (approved):
Clean and makes sense. Feels like an easy to follow foldering schema.

andrew-hirata-playeveryware (approved):

PR #913, opened 2024-09-17
feat(config): Customizable Overlay Config Button Combination

WispyMouse:
Adds the ability to configure the overlay button in the EOS Config. Currently does not allow for per-platform specific implementation. Needs testing on valid platforms.

#EOS-1592


com.playeveryware.eos/Runtime/Core/Config/Attributes/ConfigFieldAttribute.cs
```diff
@@ -49,6 +49,13 @@ public class ConfigFieldAttribute : Attribute
         /// </summary>
         public ConfigFieldType FieldType { get; }
 
+        /// <summary>
+        /// If the FieldType is ConfigFieldType.Enum, this value is used to 
+        /// determine the type of enum, so that its possible values can be
+        /// enumerated.
+        /// </summary>
+        public Enum EnumType { get; }
```
paulhazen: This should not be within the `ConfigFieldAttribute` class, because it is only used when the config field attribute is representing an enum value. Instead, a class should be created for `ConfigFieldAttributes` of type Enum:

```c#
public class EnumFieldAttribute : ConfigFieldAttribute
{
    public Enum EnumType { get; }
    /* ... */
}
```

However, since this PR does not introduce any area where this functionality is utilized, it should be excluded from this PR.

com.playeveryware.eos/Runtime/Core/Config/Attributes/ConfigFieldAttribute.cs
```diff
@@ -60,5 +67,13 @@ public ConfigFieldAttribute(
...
             FieldType = type;
         }
+
+        public ConfigFieldAttribute(string label,
+            Enum enumType,
+            string tooltip = null,
+            int group = -1) : this(label, ConfigFieldType.Enum, tooltip, group)
+        {
+            EnumType = enumType;
+        }
```
paulhazen: In keeping with earlier feedback regarding not including a field member to keep track of the Enum, this constructor should be removed.

com.playeveryware.eos/Runtime/Core/Config/Attributes/ConfigFieldType.cs
```diff
@@ -78,6 +78,13 @@ public enum ConfigFieldType
         /// arbitrary functionality within a ConfigEditor.
         /// </summary>
         Button,
+
+        /// <summary>
+        /// An enum with one required value.
+        /// This will render as a dropdown list, where exactly one value can be 
+        /// selected.
+        /// </summary>
+        Enum
```
paulhazen: Until the `EnumConfigFieldAttribute` class is implemented, I don't think that this should be included here.

com.playeveryware.eos/Runtime/Core/Config/EOSConfig.cs
```diff
@@ -335,6 +336,14 @@ protected EOSConfig() : base("EpicOnlineServicesConfig.json")
             "to the SDK.", 4)]
         public bool hackForceSendInputDirectlyToSDK;
 
+        [ConfigField("Default Activate Overlay Button",
+            ConfigFieldType.Enum,
+            "Users can press the buttons associated with this value to activate " +
+            "the Epic Social Overlay. If this is multiple buttons, the user " +
+            "presses them in combination to activate the Epic Social Overlay. " +
+            "Platforms without a custom override will use this value.", 4)]
+        public InputStateButtonFlags defaultActivateOverlayButtonCombination;
```
paulhazen: Because of how `InputStateButtonFlags` is implemented - it may be clearer to name this `activatingOverlayInput`, the type being `InputStateButtonFlags` implies that the value can be the result of a bitwise operation on the input state button flags enum.

**However**, _separate_ from this being what I believe the correct implementation _should_ be - we are currently limited to the functionality provided by Unity's `JsonUtility`, because these values need to be serializable. 

This means that the best data type to use is `List<string>`. If the field stored a non-flaggable `enum`, then the field could be a `string`, but because the intent is to provide the ability to define a _set_ of buttons that activate the overlay, it needs to be a `List`. Some more work will need to be done on the rendering side to support this.

Assets/Plugins/Source/Editor/ConfigEditors/ConfigEditor.cs
```diff
@@ -261,6 +261,9 @@ protected void RenderConfigFields()
                                 onClick();
                             }
                             break;
+                        case ConfigFieldType.Enum:
+                            field.FieldInfo.SetValue(config, GUIEditorUtility.RenderInputField(field.FieldDetails, field.FieldInfo.GetValue(config) as Enum, labelWidth));
+                            break;
```
paulhazen: I think that until we have a need for this `case`, it should not be included. This is in context with other comments made in this review. \*This\* conversation can be marked as resolved so long as the others are discussed.

Assets/Plugins/Source/Editor/Utility/GUIEditorUtility.cs
```diff
@@ -374,6 +388,14 @@ public static bool RenderInputField(ConfigFieldAttribute configFieldDetails, boo
             });
         }
 
+        public static T RenderInputField<T>(ConfigFieldAttribute configFieldDetails, T value, float labelWidth, string tooltip = null) where T : Enum
+        {
+            return InputRendererWrapper<T>(configFieldDetails.Label, value, labelWidth, tooltip, (s, b, arg3, arg4) =>
+            {
+                return (T)EditorGUILayout.EnumFlagsField(CreateGUIContent(configFieldDetails.Label, tooltip), value, GUILayout.ExpandWidth(true));
+            });
+        }
```
paulhazen: I think that until we have a need for this function, it should not be included. This is in context with other comments made in this review. \*This\* conversation can be marked as resolved so long as the others are discussed.

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -533,10 +533,12 @@ private PlatformInterface CreatePlatformInterface()
             //-------------------------------------------------------------------------
             private void InitializeOverlay(IEOSCoroutineOwner coroutineOwner)
             {
+                EOSConfig configData = Config.Get<EOSConfig>();
+
                 // Sets the button for the bringing up the overlay
                 var friendToggle = new SetToggleFriendsButtonOptions
```
paulhazen, line 539: Instead of using the nomenclature of "activating the overlay," since that could mean more than one thing - does it make sense to change the language (such as the name of the field member added to `EOSConfig`) to something more specific like, `toggleFriendsInput`?

paulhazen (changes requested):

com.playeveryware.eos/Runtime/Core/Config/Attributes/ConfigFieldAttribute.cs
WispyMouse: I can agree to that focus. I think there's *some merit* in including it because it makes clear that this is something that needs support when we do implement that, but it's in your brain now. Will exclude. Good point on it not being applicable to other kinds of field attributes.

andrew-hirata-playeveryware (changes requested):
As mentioned in the review meeting, we need to change enum to be a string to safeguard ourselves against any additions to the enum by the EOS SDK.

Assets/Plugins/Source/Editor/Utility/GUIEditorUtility.cs
paulhazen: I've "unresolved" this conversation because it appears that this function is still included in the changes.

com.playeveryware.eos/Runtime/Core/Config/EOSConfig.cs
```diff
@@ -441,6 +444,19 @@ public AuthScopeFlags GetAuthScopeFlags()
                 AuthScopeFlagsExtensions.TryParse);
         }
 
+        /// <summary>
+        /// Returns a single InputStateButtonFlags enum value that results from a
+        /// bitwise OR operation of all the defaultActivateOverlayButtonCombination flags on this
```
paulhazen: The documentation should be updated to reflect the new name for the field member.

com.playeveryware.eos/Runtime/Core/EOS_SDK_Additions/Extensions/InputStateButtonFlagsExtensions.cs
```diff
@@ -0,0 +1,131 @@
...
+            { "FaceButtonTop", InputStateButtonFlags.FaceButtonTop },
+            { "LeftShoulder", InputStateButtonFlags.LeftShoulder },
+            { "LeftThumbstick", InputStateButtonFlags.LeftThumbstick },
+            { "LeftTrigger", InputStateButtonFlags.LeftTrigger },
+            { "RightShoulder", InputStateButtonFlags.RightShoulder },
+            { "RightThumbstick", InputStateButtonFlags.RightThumbstick },
+            { "RightTrigger", InputStateButtonFlags.RightTrigger },
+            { "SpecialLeft", InputStateButtonFlags.SpecialLeft },
+            { "SpecialRight", InputStateButtonFlags.SpecialRight }
+        };
```
paulhazen: Because there is a 1:1 mapping between the `string` key for this `Dictionary` and the field name it maps to, there is actually no need for this `CustomMappings` field. 

The `CustomMappings` field member defined in other extension classes (like [`AuthScopeFlagsExtensions.cs`](https://github.com/PlayEveryWare/eos_plugin_for_unity/blob/2a18ec7e3fddff1c35baf37195a9d7989be1c4bd/com.playeveryware.eos/Runtime/Core/Utility/Extensions/AuthScopeFlagsExtensions.cs) is used because there are string values that need to be correlated to enum values whose names were changed at some point. 

Adding the `CustomMappings` field in these extension classes is done to support backwards-compatibility for users who have string values for their enum selections that no longer map to the current implementation of `AuthScopeFlags` (which we have no control over since it's a part of the EOS SDK). 

Since the values for `InputStateButtonFlags` can be determined by calling `.ToString()` on the `InputStateButtonFlags` enum value this process is unnecessary.

Assets/Plugins/Source/Editor/EditorWindows/EOSSettingsWindow.cs
```diff
@@ -347,6 +349,12 @@ private void OnDefaultGUI()
             GUIEditorUtility.AssigningBoolField("Always send Input to Overlay",
                 ref mainEOSConfigFile.alwaysSendInputToOverlay, 190,
                 "If true, the plugin will always send input to the overlay from the C# side to native, and handle showing the overlay. This doesn't always mean input makes it to the EOS SDK.");
+
+            InputStateButtonFlags toggleFriendsButtonCombinationEnum = mainEOSConfigFile.GetToggleFriendsButtonCombinationFlags();
+            GUIEditorUtility.AssigningEnumField<InputStateButtonFlags>("Default Activate Overlay Button",
+                ref toggleFriendsButtonCombinationEnum, 190,
+                "Users can press the button(s) associated with this value to activate the Epic Social Overlay. Not all combinations are valid; the SDK will log an error at the start of runtime if an invalid combination is selected.");
+            mainEOSConfigFile.toggleFriendsButtonCombination = toggleFriendsButtonCombinationEnum.FlagsToStrings();
```
paulhazen: This call to `FlagsToStrings()` can be replaced with the following call using [`EnumUtility.cs`](https://github.com/PlayEveryWare/eos_plugin_for_unity/blob/2a18ec7e3fddff1c35baf37195a9d7989be1c4bd/com.playeveryware.eos/Runtime/Core/Utility/EnumUtility.cs):
```c#
EnumUtility<InputStateButtonFlags>.GetEnumerator(toggleFriendsButtonCombinationEnum)
    .Select(enumValue => enumValue.ToString())
    .ToList();
```
You will also need to add a using reference at the top of the file:
```c#
using System.Linq;
```
(which is fine to do since this is all Editor code)

com.playeveryware.eos/Runtime/Core/EOS_SDK_Additions/Extensions/InputStateButtonFlagsExtensions.cs
```diff
@@ -0,0 +1,131 @@
...
+            return EnumUtility<InputStateButtonFlags>.TryParse(stringFlags, CustomMappings, out result);
+        }
+
+        /// <summary>
+        /// Parses an InputStateButtonFlags value in to its component strings.
+        /// These strings can then be safely stored in a configuration value.
+        /// </summary>
+        /// <param name="flags">Enum to separate into parts.</param>
+        /// <returns>A list of strings representing these flags.</returns>
+        public static List<string> FlagsToStrings(this InputStateButtonFlags flags)
```
paulhazen: Per my other comment regarding the place where this function is called - this function can be removed.

paulhazen (changes requested):

Assets/Plugins/Source/Editor/Utility/GUIEditorUtility.cs
WispyMouse: Oh, I must of missed one. Let's get 'em.

WispyMouse: Removed!

com.playeveryware.eos/Runtime/Core/Config/EOSConfig.cs
WispyMouse: Oh, yup! In fact that's why I normally use the <seealso cref="toggleFriendsButtonCombination"/>, it makes things auto change when I use that rename functionality. I added it now, here.

com.playeveryware.eos/Runtime/Core/EOS_SDK_Additions/Extensions/InputStateButtonFlagsExtensions.cs
WispyMouse: I wasn't sure that was the behaviour, so I made a tiny dotnetfiddle.

```csharp
[Flags]
	public enum FlaggedEnum : int
	{
		None = 0,
		Aaa = 1 << 0,
		Bbb = 1 << 1,
		Ccc = 1 << 2
	}
	
	public static void Main()
	{
		FlaggedEnum myFlaggedEnum = FlaggedEnum.Aaa;
		Console.WriteLine(myFlaggedEnum.ToString());
	}
```

The ToString resulted in printing out the name of the enum, so we can use that. I was concerned it would print out the numeric value.

Assets/Plugins/Source/Editor/EditorWindows/EOSSettingsWindow.cs
WispyMouse: Alertness raised at seeing LINQ, and then lowered on seeing that yup this is editor code. We'll do that here.

WispyMouse: That means we can get rid of FlagsToStrings, so I'll do so now. I was actually uncertain which way you'd lean, on whether there should be a function to do it or using this.

WispyMouse: Implemented above as suggested.

andrew-hirata-playeveryware (approved):

paulhazen (approved):

PR #916, opened 2024-09-18
Symlink Documentation

paulhazen:
This PR replaces the markdown files in the root of the repository with symbolic links that point to the files in `com.playeveryware.eos`. It is part of the larger task of preparing to make a release and born out of a desire not to duplicate work by maintaining two copies of `CHANGELOG.md`.

The reason for doing this:
- Removes the need to maintain two copies of each markdown file (`README.md`, `CHANGELOG.md`, and `LICENSE.md`), which reduces the changes that changes diverge between the copies, and in turn reduces the chances that we have inconsistencies in our documentation.

Things that were tested to confirm that this works as expected:
- Repository was freshly checked out on Windows, and Unix to confirm that the respective file systems treated the symbolic links as expected.
- The links within each markdown file were tested on GitHub - and the existing relative paths used in each file were properly honored.
- A package was created, and the markdown files that were moved during the package creation process are moved into the expected locations and behave appropriately.

#EOS-2115


andrew-hirata-playeveryware:
I would actually prefer the exact opposite of this: the main files being in the UPM directory. We have one task in the triage list to encapsulate part of this work.

paulhazen:
> I would actually prefer the exact opposite of this: the main files being in the UPM directory. We have one task in the triage list to encapsulate part of this work.

This is what is done. The files are in `com.playeveryware.eos/`, and the symlinks are in the root of the git repository.

andrew-hirata-playeveryware (approved):

paulhazen:
I am closing this PR because the solution isn't quite right. While it _does_ do _most_ of what we want - it makes the symlinks when selected within the GitHub UI display their contents - which is just the path to the file being pointed to. This is not a good user experience.

PR #918, opened 2024-09-18
Release 3.3.3

paulhazen:
These changes represent the upcoming release (`v3.3.3`)

#EOS-2115


andrew-hirata-playeveryware:
This branch appears to have been made off of `stable` instead of being based off of `development` ?

paulhazen:
> This branch appears to have been made off of `stable` instead of being based off of `development` ?

In error, I had opened the PR _into_ `stable`, but it _is_ being made off of `development`.

.github/workflows/check-for-package-difference.yml
paulhazen, line 1: This file was deleted because it is an extraneous file that no longer serves a meaningful purpose.

_It's deleted as part of **this** release_, because the check fails, so to keep it in would show a failed checkmark - which sends the wrong message to users, particularly in the context of it being a release.

tools/scripts/print-commits.sh
```diff
@@ -0,0 +1,17 @@
...
+
+# Check if a tag was found
+if [ -z "$MOST_RECENT_TAG" ]; then
+  echo "No tags found in the repository."
+  exit 1
+fi
+
+# Print commit messages, hashes, and timestamps between the most recent tag and the current branch
+echo "Commit messages between branch '$CURRENT_BRANCH' and the most recent tag '$MOST_RECENT_TAG':"
+git log --pretty=format:"%cd - %an - %h - %s" --date=short $MOST_RECENT_TAG..$CURRENT_BRANCH
```
paulhazen, line 17: This file is included in this release because it is a script I commonly use as part of my process for writing the `CHANGELOG.md` file, and I thought including it would be worthwhile. 

Please let me know if that is insufficient reason to include it as part of the release, and I will exclude the change.

com.playeveryware.eos/Runtime/Core/EOSPackageInfo.cs
paulhazen, line 1: This file (`EOSPackageInfo.cs` has meaningful implementation added to it because (as part of the normal release process) we have been manually updating the contents such that they reflect the values in the `package.json` file. 

The added implementation improves our release process by removing that step from here on out - and reducing the steps to create a release seems like a happy thing to do.

In summary: either this file gets these changes (and we never have to change it again), or this file gets updated with string literals for `PackageName` and `Version` field members / properties.

.github/workflows/check-for-package-difference.yml
andrew-hirata-playeveryware: I'm confused by this delete: the package.json file should indeed still exist, and this check was added to ensure that the package.json file is not the same when merged into stable.

com.playeveryware.eos/Runtime/Core/EOSPackageInfo.cs
andrew-hirata-playeveryware: This should be a separate PR that we can look at that can be added into a future release.

tools/scripts/print-commits.sh
andrew-hirata-playeveryware: This ideally should be added in it's own PR that goes through a separate review and can be merged into `development`.

.github/workflows/check-for-package-difference.yml
paulhazen: Good question - this needs to be reconsidered.

tools/scripts/print-commits.sh
paulhazen: 10-4, I will break this out into its own PR.

com.playeveryware.eos/Runtime/Core/EOSPackageInfo.cs
paulhazen: Sounds good - I will open this into a separate PR (and do the appropriate tasking related to it).

paulhazen:
I am closing this PR because I am going to recreate the branch based off of `development` for the purpose of clarity.

PR #919, opened 2024-09-18
Copy Docs Workflow

paulhazen:
> [!WARNING]
> There was a previous [PR](https://github.com/PlayEveryWare/eos_plugin_for_unity/pull/916) that was opened to solve this problem, but the solution was not ideal. The description of that PR has been repeated in parts below. As the (now closed) PR's final comment reads, the reason for abandoning the approach was as follows:
> > I am closing this PR because the solution isn't quite right. While it does do most of what we want - it makes the symlinks when selected within the GitHub UI display their contents - which is just the path to the file being pointed to. This is not a good user experience.

This PR introduces a GitHub workflow that _ only when changes are detected to the documentation markdown files_ will trigger, copying the markdown files into the root of the repository from where they reside in `com.playeveryware.eos`. Local changes that are made to these files are now ignored via the `.gitignore` file. 

When the _workflow_ action copies the files and commits the change, it overrides the `.gitignore`. This finally creates the solution wherein there is no need to maintain two copies of each markdown file - and prevents confusion regarding which to edit by excluding any edits made to the files in the root of the repository.

The reason for doing this:
- Removes the need to maintain two copies of each markdown file (`README.md`, `CHANGELOG.md`, and `LICENSE.md`), which reduces the chances that changes diverge between the copies, and (in turn) reduces the chances that we have inconsistencies in our documentation.

> [!NOTE]
> There are whitespace changes intentionally made to `CHANGELOG.md`, that _caused_ the workflow introduced in this PR to be triggered within this branch. 

#EOS-2115


WispyMouse (approved):

.github/workflows/copy-docs.yml
```diff
@@ -0,0 +1,33 @@
+name: Sync Docs
+
+on:
+  push:
+    branches:
+      - '**'  # Trigger on any branch
+    paths:
+      - 'com.playeveryware.eos/README.md'
+      - 'com.playeveryware.eos/LICENSE.md'
+      - 'com.playeveryware.eos/CHANGELOG.md'
```
andrew-hirata-playeveryware: From the view point of keeping the CHANGELOG relative to the code it effects and to keep it consistent, I was intending to have the CHANGLOG _not_ be in the root of the repo. 

For the README, wasn't it the case where one may configure something in the .github directory to have it reference a different readme.md file? If so, that would be the preferred solution to this.

For the license file, I don't believe we need to worry about syncing it up, unless someone tries to be sneaky and has a different version between them.

andrew-hirata-playeveryware (changes requested):

.github/workflows/copy-docs.yml
paulhazen: > From the view point of keeping the CHANGELOG relative to the code it effects and to keep it consistent, I was intending to have the CHANGLOG not be in the root of the repo.

Makes sense. I have updated the action accordingly.

> For the README, wasn't it the case where one may configure something in the .github directory to have it reference a different readme.md file? If so, that would be the preferred solution to this.

With respect to putting it in the `.github` directory, this does not work as well as I had thought it might. 

The way this GitHub feature works is one places the README.md file **itself** within the `.github` directory. That would mean that it would _still_ either need to be linked to so that there is a symlink within `com.playeveryware.eos/` that points to `.github/README.md` (which feels convoluted), or have a process that copies it, which defeats the entire purpose.

Therefore, IMO: Maintaining the automation of copying the `README.md` file into the root of the directory and adding `README.md` to the `.gitignore` file _still makes sense_ considering that the workflow _only runs_ when changes to `README.md` are detected (meaning 99% of the time it will not run - and incur zero cost for execution). This way `README.md` at the root of the repository _is always guaranteed_ to have contents identical to `com.playeveryware.eos/README.md`

Just to address all the points in your comment:
> For the license file, I don't believe we need to worry about syncing it up, unless someone tries to be sneaky and has a different version between them.

Yeah that makes sense. I've removed that file from consideration.

andrew-hirata-playeveryware: Perhaps I'm mistaken, but I believe that the file under `.github/` can itself be a symlink to another location. i.e. a `.github/README.md` would be a symlink to `com.playeveryware.eos/README.me`

paulhazen:
Closing this PR as another approach has been found that is far superior.

PR #922, opened 2024-09-20
Documentation Link Checker

paulhazen:
This PR contains the script I created to identify broken links in the documentation, and a workflow for GitHub that triggers whenever a markdown file has changed. Configured properly, it can be used to prevent merging broken links into `stable`, `development`, and/or `release` branches (whichever we would prefer).


tools/scripts/check_links.py
```diff
@@ -0,0 +1,238 @@
...
+        if not is_external_link(link):
+            abs_link = compute_abs_link(md_dir, link)
+    else:
+        if is_external_link(link):
+            try:
+                response = session.head(link, allow_redirects=True, timeout=10, headers=headers)
+                if response.status_code != 403 and (response.status_code >= 400 or response.status_code == 405):
+                    response = session.get(link, allow_redirects=True, timeout=10, headers=headers)
+                    if response.status_code >= 400 and response.status_code != 403:
+                        result = response.status_code
```
paulhazen: This needs more documentation to explain why `403` is excluded from consideration.

paulhazen: This has been updated to include documentation and to be clearer.

.github/workflows/check-markdown-links.yml
```diff
@@ -0,0 +1,24 @@
+name: Check Markdown Links
+
+on:
+  push:
+    branches: ["feat/documentation-link-checker"]
```
andrew-hirata-playeveryware, line 5: This requires another change later to set it to the correct branch, right?

andrew-hirata-playeveryware (approved):

.github/workflows/check-markdown-links.yml
paulhazen: That is correct. I figure that determining _where_ we want this to be enabled would be a conversation for another PR.

arthur740212 (approved):

PR #926, opened 2024-09-23
`CHANGELOG.md` scripting assistant

paulhazen:
This PR introduces a script that can be helpful in writing the `CHANGELOG.md` file for a release by printing all of the commit messages between the last tag in the repository and the current branch.

#EOS-2133

### DIFF
--- a/tools/scripts/print-commits.sh
+++ b/tools/scripts/print-commits.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+# Get the current branch name
+CURRENT_BRANCH=$(git rev-parse --abbrev-ref HEAD)
+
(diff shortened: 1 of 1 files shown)


tools/scripts/print-commits.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
andrew-hirata-playeveryware, line 1: Minor pedantic remark: 
I generally prefer  to use `#!/usr/bin/env bash` because it searches the path for the version of the tool, bash in this case, installed on a POSIX/POSIX like environment.

andrew-hirata-playeveryware (approved):

arthur740212 (approved):

PR #927, opened 2024-09-23
`EOSPackageInfo` reading from `package.json`

paulhazen:
This PR introduces a change to the `EOSPackageInfo` class. Previously, the values it made available were changed via `string` literals. This changes the implementation such that the values from `package.json` are read dynamically and returned, removing the need to update this file in the future for each release.

#EOS-2134

### DIFF
--- a/com.playeveryware.eos/Runtime/Core/EOSPackageInfo.cs
+++ b/com.playeveryware.eos/Runtime/Core/EOSPackageInfo.cs
@@ -22,21 +22,92 @@
 
 namespace PlayEveryWare.EpicOnlineServices
 {
+    using System;
+    using UnityEngine;
+    using Utility;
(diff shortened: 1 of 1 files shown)


arthur740212 (changes requested):
Found this during a test build.

There is a compile error because `GetProjectPath` does not exist for in unity player.
The package also won't exist if it is not in editor, so we couldn't grab info from a `package.json` in a player.

paulhazen:
> Found this during a test build.
> 
> There is a compile error because `GetProjectPath` does not exist for in unity player. The package also won't exist if it is not in editor, so we couldn't grab info from a `package.json` in a player.

Excellent point. Since this was not originally tasked out - I'm going to add it to our needs triage and close the PR for the time being.

PR #932, opened 2024-09-25
Swap `UnityEngine.JsonUtility` for `Newtonsoft.Json`

paulhazen:
This PR swaps usage of Unity's built in `JsonUtility` with `Newtonsoft.Json`. It has been tested in release as a package (both in editor and as standalone) and functions properly.

> [!NOTE] 
> To affect this change in a manner that declares our dependencies explicitly, the `com.unity.modules.jsonserialize` dependency was replaced in `package.json` (and subsequently `packages-lock.json`) with a dependency on `com.unity.nuget.newtonsoft-json`. 
> 
> Documentation regarding this package and why it is the preferred means to include Newtonsoft.Json (aka JSON.NET) can be found [here](https://docs.unity3d.com/Packages/com.unity.nuget.newtonsoft-json@3.0/manual/index.html)

#EOS-1977

### DIFF
--- a/com.playeveryware.eos/Runtime/Core/Utility/JsonUtility.cs
+++ b/com.playeveryware.eos/Runtime/Core/Utility/JsonUtility.cs
@@ -23,68 +23,70 @@
 namespace PlayEveryWare.EpicOnlineServices.Utility
 {
     using UnityEngine;
+    using Newtonsoft.Json;
+    using System;
 
     /// <summary>
-    /// Contains functions to interact with various json data.
+    /// Contains functions to interact with various JSON data.
     /// </summary>
     public static class JsonUtility
     {
         /// <summary>
         /// Tries to parse the given JSON into an object.
         /// </summary>
-        /// <typeparam name="T">The type to deserialize from the given json.
+        /// <typeparam name="T">
+        /// The type to deserialize from the given JSON.
         /// </typeparam>
         /// <param name="json">
         /// The JSON to deserialize from.
         /// </param>
-        /// <param name="obj">The object to contain the deserialized value.
+        /// <param name="obj">
+        /// The object to contain the deserialized value.
         /// </param>
         /// <returns>
-        /// True if the json was successfully deserialized, false otherwise.
+        /// True if the JSON was successfully deserialized, false otherwise.
         /// </returns>
         private static bool TryFromJson<T>(string json, out T obj)
         {
             obj = default;
             try
             {
-                obj = UnityEngine.JsonUtility.FromJson<T>(json);
+                obj = JsonConvert.DeserializeObject<T>(json);
                 return true;
             }
-            catch
+            catch (Exception ex)
             {
                 Debug.LogError($"Unable to parse object of type " +
                                $"\"{typeof(T).FullName}\" from " +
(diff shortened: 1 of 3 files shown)


com.playeveryware.eos/Runtime/Core/Utility/JsonUtility.cs
```diff
@@ -23,38 +23,42 @@
...
+                obj = JsonConvert.DeserializeObject<T>(json);
                 return true;
             }
-            catch
+            catch (Exception ex)
             {
                 Debug.LogError($"Unable to parse object of type " +
                                $"\"{typeof(T).FullName}\" from " +
-                               $"JSON: \"{json}\".");
+                               $"JSON: \"{json}\". Exception: {ex.Message}");
```
WispyMouse: nit: Since "FullName" and "json" are wrapped in quotes, maybe exception should be too?

com.playeveryware.eos/Runtime/Core/Utility/JsonUtility.cs
```diff
@@ -115,40 +122,38 @@ public static T FromJsonFile<T>(string filepath)
...
+            }
+            try
+            {
+                JsonConvert.PopulateObject(json, obj);
+            }
+            catch (Exception ex)
+            {
+                Debug.LogError($"Unable to populate object of type " +
+                               $"\"{typeof(T).FullName}\" from " +
+                               $"JSON: \"{json}\". Exception: {ex.Message}");
```
WispyMouse: If you apply quotes above, consider applying them here too.

com.playeveryware.eos/Runtime/Core/Utility/JsonUtility.cs
```diff
@@ -70,21 +74,25 @@ private static bool TryFromJson<T>(string json, out T obj)
...
+        /// <param name="pretty">
+        /// Whether to make the JSON pretty.
+        /// </param>
         /// <returns>
         /// String representation of the given object serialized.
         /// </returns>
         public static string ToJson(object obj, bool pretty = false)
         {
-            return UnityEngine.JsonUtility.ToJson(obj, pretty);
+            return JsonConvert.SerializeObject(obj, pretty ? Formatting.Indented : Formatting.None);
```
WispyMouse, line 91: Do we want to nullcheck obj here?

WispyMouse (approved):

com.playeveryware.eos/Runtime/Core/Utility/JsonUtility.cs
paulhazen: Changed it in both instances within `JsonUtility`. Good catch, I like making these things consistent - particularly when the quotation marks confer meaning like this.

com.playeveryware.eos/Runtime/Core/Utility/JsonUtility.cs
paulhazen: We do not - because `Newtonsoft` (like many other Json serialization libraries) meaningfully translates the fact that an object is null into the serialized data.

andrew-hirata-playeveryware (approved):

PR #933, opened 2024-09-26
Data Layer Conversion

paulhazen:
This is a draft PR that demonstrates the beginning of the conversion efforts at the data layer between the existing config and the new config, by demonstrating the serializer for `Utf8String` using Unit Tests.

#EOS-2148


Assets/Tests/PlayMode/JsonConverterTests.cs
```diff
@@ -0,0 +1,208 @@
...
+        {
+            string normalString = "blah blah blah";
+            
+            Utf8String test = new Utf8String(normalString);
+
+            var json = JsonConvert.SerializeObject(test, _settings);
+
+            Utf8String deserializedTest = JsonConvert.DeserializeObject<Utf8String>(json, _settings);
+
+            // TODO: I would argue that this is a bug in Epic's code (because this test fails)
```
WispyMouse, line 69: Yup, the test fails. I wonder what should be done? I'm not content with the idea of having a test that fails that *in theory* should succeed but isn't for reasons we can never change... What should be done here?

paulhazen:
Closing this PR, as the efforts contained within have been more appropriately implemented in another branch from which a PR has been opened.

PR #935, opened 2024-09-26
Allow `Config.Get<>` and `Config.GetAsync<>` to return `Config` class with default values

paulhazen:
This PR introduces the ability to use the various `Get` functions in the `Config` class to retrieve a `Config` class without reading values from a file. This is to enable more nuanced parsing of `Config` serialized data to maintain backwards compatibility. It is part of a larger effort to enable config override improvements.

#EOS-2141


paulhazen:
Closing because work has not been properly tasked yet.

PR #937, opened 2024-09-27
Config Validation Unit tests

arthur740212:
Provides attributes to mark test types on a config field.
Tests for fields in `EOSConfig`
Generic FieldValidator that grabs the marked test types of a field


Assets/Tests/PlayMode/EOSConfigTests.cs
```diff
@@ -218,7 +218,7 @@ public void DeploymentOverride_MustBeValid()
                 }
             };
 
-            if (!config.TryGetFailingValidatorAttributes(out List<FieldValidatorFailure> failingAttributes))
+            if (!FieldValidator.TryGetFailingValidatorAttributes(config.sandboxDeploymentOverrides[0], out List<FieldValidatorFailure> failingAttributes))
```
WispyMouse: Here, I think we should run this on both the config and on the targeted override. Running it on the config ensures that the config is checking its innards for more problems.

com.playeveryware.eos/Runtime/Core/Config/Attributes/FieldValidator.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        public FieldValidatorFailure(FieldInfo failingField,
+            FieldValidatorAttribute failingAttribute,
+            string failingMessage)
+        {
+            FieldInfo = failingField;
+            FailingAttribute = failingAttribute;
+            FailingMessage = failingMessage;
+        }
+    }
```
WispyMouse, line 43: nit: add new line

com.playeveryware.eos/Runtime/Core/Config/Attributes/FieldValidator.cs
```diff
@@ -0,0 +1,64 @@
...
+    public class FieldValidator
+    {
+        public static bool TryGetFailingValidatorAttributes(object target, out List<FieldValidatorFailure> failingValidatorAttributes)
```
WispyMouse: Something possibly lost from this change is that it isn't checking fields inside itself that also have attributes. Should the config validator be able to do that all in one call? Is there a way we can recursively ask children for their attributes?

com.playeveryware.eos/Runtime/Core/Config/Config.cs
```diff
@@ -220,7 +220,7 @@ public static T Get<T>() where T : Config
         /// <returns>
```
paulhazen: Instead of being made `public`, if these things need to be utilized outside of the `Config` class, I believe that they should be moved to a shared location outside of where either is used. Perhaps they can be moved into the `EnumUtility` class itself?

paulhazen (changes requested):
I started a review, but then noticed that there were quite a lot of differences between the branch being merged from and the `development` branch (about 125 commits).

I first noticed this looking at the layout of the `eos_plugin_for_unity/Assets/Tests/PlayMode/` directory, which in `development` looks like this:

image

I would imagine that the `EOSConfigTests` file that is added would be put under the `Config` directory.

Once I noticed that the branch was so far out of date, I stopped my review because I imagine a lot might be changed, and my feedback would be more clear afterwards.

com.playeveryware.eos/Runtime/Core/Config/Attributes/FieldValidator.cs
arthur740212: Yes, this is next task on deck.

Assets/Tests/PlayMode/EOSConfigTests.cs
arthur740212: This will be updated alongside the recursive validate task

com.playeveryware.eos/Runtime/Core/Config/Config.cs
arthur740212: It is reverted instead because it isn't used anywhere in the plugin

com.playeveryware.eos/Runtime/Core/Config/Attributes/ExpandFieldAttribute.cs
```diff
@@ -0,0 +1,35 @@
...
+ * SOFTWARE.
+ */
+
+namespace PlayEveryWare.EpicOnlineServices
+{
+    using System;
+
+    /// <summary>
+    /// This attribute is used to decorate a field member within a config class
+    /// that represents a path to a file.
```
paulhazen: This comment appears to be copied from another attribute. It should be updated to describe the purpose of this attribute - which isn't clear to me.

com.playeveryware.eos/Runtime/Core/Config/Attributes/GUIDFieldValidatorAttribute.cs
```diff
@@ -0,0 +1,58 @@
...
+
+    [AttributeUsage(AttributeTargets.Field)]
+
+    public class GUIDFieldValidatorAttribute : FieldValidatorAttribute
+    {
+        public const string NotAGuidMessage = "The field value could not be parsed into a Guid.";
+
+        public override bool FieldValueIsValid(object toValidate, out string configurationProblemMessage)
+        {
+            if (!(toValidate is string stringValue))
```
paulhazen, line 36: I _think_ you can actually change this to be:
```c#
if (toValidate is not string stringValue)
```

com.playeveryware.eos/Runtime/Core/Config/Attributes/ParsesToAuthScopeFieldValidatorAttribute.cs
```diff
@@ -0,0 +1,48 @@
...
+    {
+        protected override bool CanParse(List<string> flags)
+        {
+            if (!AuthScopeFlagsExtensions.TryParse(flags, out AuthScopeFlags results))
+            {
+                return false;
+            }
+
+            return true;
+        }
```
paulhazen: This should no longer be needed - since the `List<string>` type of Enums disappeared for `EOSConfig`.

com.playeveryware.eos/Runtime/Core/Config/Attributes/ParsesToEnumFieldValidatorAttribute.cs
```diff
@@ -0,0 +1,67 @@
...
+                configurationProblemMessage = FailedToParseTokensMessage;
+                return false;
+            }
+
+            configurationProblemMessage = string.Empty;
+            return true;
+        }
+
+        protected abstract bool CanParse(List<string> flags);
+    }
```
paulhazen: As with the previous file - this one should no longer be needed.

Assets/Tests/PlayMode/EOSConfigTests.cs
paulhazen, line 1: I believe that this should be moved to `Assets/Tests/PlayMode/Config/EOSConfigTests.cs`

com.playeveryware.eos/Runtime/Core/Config/EOSConfig.cs
```diff
@@ -200,6 +212,7 @@ protected EOSConfig() : base("EpicOnlineServicesConfig.json")
             "Platform option flags",
             3)]
         [JsonConverter(typeof(ListOfStringsToAuthScopeFlags))]
+        [ParsesToAuthScopeFieldValidator]
         public AuthScopeFlags authScopeOptionsFlags;
```
paulhazen: This should no longer need the field validator. None of the field members that are `enum` types should need them, since it is not possible to have an invalid value for them - unless the JSON itself is edited - and in that case the error should be generated from the JSON data layer.

com.playeveryware.eos/Runtime/Core/Config/Attributes/DevelopmentEnvironmentFieldValidatorAttribute.cs
```diff
@@ -0,0 +1,88 @@
...
+            }
+
+            if (Guid.TryParse(toValidate, out _))
+            {
+                return true;
+            }
+
+            return false;
+        }
+    }
```
paulhazen: I believe that this class should be named `SandboxFieldValidator` or something that makes it clearer what it's purpose is.

Additionally, what you could do to reduce the duplicated code between this class and the GUID validator, is you could have this class _extend_ the GUID validator, and the override could call the base implementation from the GUID validator, but then also test for the private pattern.

paulhazen (changes requested):

com.playeveryware.eos/Runtime/Core/Config/Attributes/DevelopmentEnvironmentFieldValidatorAttribute.cs
```diff
@@ -0,0 +1,88 @@
...
+            {
+                configurationProblemMessage = string.Empty;
+                return true;
+            }
+
+            configurationProblemMessage = FieldDidNotMatchMessage;
+            return false;
+        }
+
+        bool isValidEntry(string toValidate)
```
paulhazen: Is this function utilized anywhere? It seems to do similar things to the function defined above it.

andrew-hirata-playeveryware: If this isn't used, can we remove it?

andrew-hirata-playeveryware (approved):

com.playeveryware.eos/Runtime/Core/Config/Attributes/ParsesToAuthScopeFieldValidatorAttribute.cs
arthur740212: This can be removed, because it is specific on `AuthScopeFlags`

com.playeveryware.eos/Runtime/Core/Config/Attributes/ParsesToEnumFieldValidatorAttribute.cs
arthur740212: This will be kept for now because it is a generic enum parse test. And we only dealt with EOSConfig enums, while other enums might not have the utility ready.

com.playeveryware.eos/Runtime/Core/Config/Attributes/ParsesToAuthScopeFieldValidatorAttribute.cs
arthur740212: Actually it can't because it needs to implement CanParse from its base

com.playeveryware.eos/Runtime/Core/Config/Attributes/GUIDFieldValidatorAttribute.cs
arthur740212: Yes, but it isn't necessary

com.playeveryware.eos/Runtime/Core/Config/EOSConfig.cs
arthur740212: This is kept and the test was changed to test for invalid input instead

com.playeveryware.eos/Runtime/Core/Config/Attributes/ParsesToAuthScopeFieldValidatorAttribute.cs
paulhazen: I think that `ParsesToEnumFieldValidatorAttribute` can be removed altogether - there should not be a need for it since there are no longer any `enum` values that are represented by `List<string>`.

com.playeveryware.eos/Runtime/Core/Config/Attributes/ParsesToEnumFieldValidatorAttribute.cs
```diff
@@ -0,0 +1,68 @@
...
```
paulhazen: I do not believe that this class is needed any longer.

com.playeveryware.eos/Runtime/Core/Config/Attributes/ParsesToPlatformFlagFieldValidatorAttribute.cs
```diff
@@ -0,0 +1,47 @@
...
+    using System;
+    using System.Collections.Generic;
+    using System.Reflection;
+    using System.Text;
+    using System.Text.RegularExpressions;
+
+    [AttributeUsage(AttributeTargets.Field)]
+
+    public class ParsesToPlatformFlagFieldValidatorAttribute : ParsesToEnumFieldValidatorAttribute
+    {
```
paulhazen: I do not believe that this class is needed any longer.

com.playeveryware.eos/Runtime/Core/Config/Attributes/ParsesToUlongFieldValidatorAttribute.cs
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+namespace PlayEveryWare.EpicOnlineServices
+{
+    using System;
+    using System.Reflection;
+
+    [AttributeUsage(AttributeTargets.Field)]
+
+    public class ParsesToUlongFieldValidatorAttribute : FieldValidatorAttribute
```
paulhazen: I do not believe that this class is needed any longer.

com.playeveryware.eos/Runtime/Core/Config/Attributes/SandboxIDFieldValidatorAttribute.cs
```diff
@@ -0,0 +1,68 @@
...
+            {
+                configurationProblemMessage = $"The field value is not of type string.";
+                return false;
+            }
+
+            if (string.IsNullOrEmpty(singleStringValue))
+            {
+                configurationProblemMessage = string.Empty;
+                return true;
+            }
```
paulhazen: This implies to me that `SandboxID` _can_ be null or empty - but I do not think that this is the case.

paulhazen (changes requested):

com.playeveryware.eos/Runtime/Core/Config/Attributes/SandboxIDFieldValidatorAttribute.cs
WispyMouse: Correct, that's what this logic does. Looking at the original spec of the task, that's not one of the things this should allow. So we'll make sure to change this soon to not allow empties.

arthur740212: Done

paulhazen (approved):

PR #938, opened 2024-09-27
Config Field Members Type Update

paulhazen:
This PR makes use of the newly introduced `Newtonsoft.Json` library to change a variety of field members in a variety of `Config` implementing classes to have more appropriate and easier to validate types.

At the data deserialization layer - logic has been added where appropriate to accept JSON data that is either in `string` format, or in the new type. This allows usage of the `EnumField` in the Unity Editor for `EOSConfig` and the various `PlatformConfig` classes.

Unit tests have been added that guarantee the functionality of these conversions by testing that a serialized object with an `enum` field member can be deserialized from either a JSON array of string values (old implementation) or a straightforward serialization of the `enum` value (new implementation).

Similar unit tests have been written to guarantee that values that used to be `string` can now be their appropriate nullable value (such as `float?`, `int?`, etc.) as well as their non-nullable variants.

#EOS-2148


paulhazen:
Can we guarantee that this works correctly within the native code where this is necessary to be read?

arthur740212 (approved):

Assets/Plugins/Source/Editor/EditorWindows/EOSSettingsWindow.cs
```diff
@@ -342,26 +342,35 @@ private void OnDefaultGUI()
...
+                "Platform Flags",
+                ref mainEOSConfigFile.platformOptionsFlags,
+                190,
+                "Flags used to initialize EOS Platform. Available " +
+                "flags are defined in PlatformFlags.cs");
+
+            GUIEditorUtility.AssigningEnumField(
+                "Auth Scope Flags",
+                ref mainEOSConfigFile.authScopeOptionsFlags,
+                210,
```
WispyMouse: Can this be changed to 190? It'd line things up better, I think.
image

Assets/Plugins/Source/Editor/Utility/GUIEditorUtility.cs
```diff
@@ -96,47 +97,44 @@ public static void AssigningTextField(string label, ref string value, float labe
...
+            if (value.HasValue)
             {
-                value = null;
+                textToDisplay = value.Value.ToString();
             }
-            catch (OverflowException)
+            
+            string newTextValue = EditorGUILayout.TextField(guiLabel, textToDisplay, GUILayout.ExpandWidth(true));
+            
+            if (GUILayout.Button("Clear", GUILayout.MaxWidth(50)))
```
WispyMouse, line 120: selectedfield
This clear button functions correctly in that it eliminates the value. *But*, if you have that value currently selected and then press clear, the value will appear to still be in the config field. Clicking away will reveal it was cleared properly. Is there a way we can make this clear button also clear the selection?

Assets/Plugins/Source/Editor/Utility/GUIEditorUtility.cs
```diff
@@ -154,46 +152,43 @@ public static void AssigningBoolField(string label, ref bool value, float labelW
             EditorGUIUtility.labelWidth = originalLabelWidth;
         }
 
-        public static void AssigningFloatToStringField(string label, ref string value, float labelWidth = -1, string tooltip = null)
+        public static void AssigningFloatToStringField(string label, ref float? value, float labelWidth = -1, string tooltip = null)
```
WispyMouse, line 155: I'm looking at this and it *feels* like this is so similar to the other function, it's a sign we want to make things more generic somehow. But I can't really think of a way to do that in a way that's actually cleaner or better than just repeating ourselves...

> delegate bool TryParseFunc<T>(string input, out T output);
> public static void AssigningValueToStringField<T>(string label, ref T? value,  TryParseFunc<T> tryParseFunc, float labelWidth = -1, string tooltip = null)
> ...

Is that any better? Maybe we have this wrapper function signature that calls an inner function? Iunno, spitballing. I'd be fine with this as is.

Assets/StreamingAssets/EOS/EpicOnlineServicesConfig.json
```diff
@@ -1,17 +1,26 @@
...
+  "productVersion": "1.0",
+  "productID": "f7102b835ed14b5fb6b3a05d87b3d101",
+  "sandboxID": "ab139ee5b644412781cf99f48b993b45",
+  "deploymentID": "c529498f660a4a3d8a123fd04552cb47",
+  "sandboxDeploymentOverrides": [],
+  "isServer": false,
+  "clientSecret": "aXPlP1xDH0PXnp5U+i+M5pYHhaE1a8viV0l1GO422ms",
+  "clientID": "xyza7891wPzGRvRf4SkjlIF8YuqlRLbQ",
+  "encryptionKey": "1111111111111111111111111111111111111111111111111111111111111111",
+  "platformOptionsFlags": "LoadingInEditor, DisableSocialOverlay, Reserved1, WindowsEnableOverlayD3D9, WindowsEnableOverlayD3D10, WindowsEnableOverlayOpengl, ConsoleEnableOverlayAutomaticUnloading",
```
WispyMouse: Is it intentional that this is all defined? Is this the actual default value?

Assets/StreamingAssets/EOS/EpicOnlineServicesConfig.json
```diff
@@ -1,17 +1,26 @@
...
+  "tickBudgetInMilliseconds": 0,
+  "taskNetworkTimeoutSeconds": 0.0,
+  "ThreadAffinity_networkWork": null,
+  "ThreadAffinity_storageIO": null,
+  "ThreadAffinity_webSocketIO": null,
+  "ThreadAffinity_P2PIO": null,
+  "ThreadAffinity_HTTPRequestIO": null,
+  "ThreadAffinity_RTCIO": null,
+  "alwaysSendInputToOverlay": true,
+  "initialButtonDelayForOverlay": 0.0,
```
WispyMouse: I feel like initialButtonDelayForOverlay should be null by default, not 0.0. I believe that if it is null EOS will use default behaviour, while 0.0 will cause it to actually use 0.0.

Assets/Tests/PlayMode/Config/JsonConverters/ListOfStringsToEnumConverterTests.cs
```diff
@@ -0,0 +1,102 @@
...
+            { "ThirdLetter", TestOrderEnum.C },
+            { "AppleStartsWith", TestOrderEnum.A }
+        };
+
+        [Flags]
+        private enum TestOrderEnum : int
+        {
+            None = 0x00000,
+            A = 0x00001,
+            B = 0x00004,
```
WispyMouse, line 46: Should this be 0x00002? If tests are passing, I guess it doesn't matter much.

Assets/Plugins/Source/Editor/EditorWindows/EOSSettingsWindow.cs
paulhazen: Good idea - fixed.

Assets/StreamingAssets/EOS/EpicOnlineServicesConfig.json
paulhazen: No - this was to demonstrate that things were being serialized the way that we wanted them to be. I have now updated them to revert to the correct default values.

Assets/Tests/PlayMode/Config/JsonConverters/ListOfStringsToEnumConverterTests.cs
paulhazen: It should not - because if it was, then the bitwise operators would not work they way we want them to. When the values are explicitly defined they need to be spaced such that the bitwise operators do not result in one of the enum values implying the other by default.

Assets/Plugins/Source/Editor/Utility/GUIEditorUtility.cs
paulhazen: Unfortunately, if the field is still selected, then it is considered as something that is still being edited, as in the value is not accessible (or clearable) until focus for the field is lost. Clicking the 'clear' button doesn't move the cursor, so until that happens the value cannot be cleared.

Assets/Plugins/Source/Editor/Utility/GUIEditorUtility.cs
paulhazen: Because eventually these will all use the `RenderInput` methods that inspect the config fields and do all the rendering, I opted not to optimize these functions since they will be going away in the near future.

Assets/StreamingAssets/EOS/EpicOnlineServicesConfig.json
paulhazen: You are correct. I've fixed this now.

com.playeveryware.eos/Runtime/Core/Config/JsonConverter/ListOfStringsToEnumConverters.cs
```diff
@@ -0,0 +1,235 @@
...
+        /// </summary>
+        /// <param name="objectType">
+        /// The object to convert _to_.
+        /// </param>
+        /// <returns>
+        /// True if the conversion is supported, false otherwise.
+        /// </returns>
+        public override bool CanConvert(Type objectType)
+        {
+            return typeof(T).IsEnum;
```
WispyMouse, line 151: Isn't T always an Enum, because of the generic type restriction on the class?

Assets/Tests/PlayMode/Config/JsonConverters/ListOfStringsToEnumConverterTests.cs
WispyMouse: That surprises me, given PlatformFlags being laid out like this:

```csharp
namespace Epic.OnlineServices.Platform
{
	[System.Flags]
	public enum PlatformFlags : ulong
	{
		None = 0x0,
		/// <summary>
		/// A bit that indicates the SDK is being loaded in a game editor, like Unity or UE4 Play-in-Editor
		/// </summary>
		LoadingInEditor = 0x00001,
		/// <summary>
		/// A bit that indicates the SDK should skip initialization of the overlay, which is used by the in-app purchase flow and social overlay. This bit is implied by <see cref="LoadingInEditor" />
		/// </summary>
		DisableOverlay = 0x00002,
		/// <summary>
		/// A bit that indicates the SDK should skip initialization of the social overlay, which provides an overlay UI for social features. This bit is implied by <see cref="LoadingInEditor" /> or <see cref="DisableOverlay" />
		/// </summary>
		DisableSocialOverlay = 0x00004,
		/// <summary>
		/// A reserved bit
		/// </summary>
		Reserved1 = 0x00008,
```

Here, it uses 0x00002 as the next value. Is this different, somehow?

WispyMouse (changes requested):
*Something* about this changeset results in the Editor hanging when trying to run things. I can't pinpoint where that is. When that bug has been addressed, this is approved to go. Had some small comments that you've already addressed.

com.playeveryware.eos/Runtime/Core/Config/JsonConverter/ListOfStringsToEnumConverters.cs
paulhazen: Yes - however implementing the abstract class still requires implementation of this function, and this is the most sensical way to do that. Sort of serves as an extra check.

Assets/Tests/PlayMode/Config/JsonConverters/ListOfStringsToEnumConverterTests.cs
paulhazen: This is not different; it is intended functionality (see the comments on each successive value within the enum definition).

For the purposes of the _test_ I needed an example where this _wasn't_ the case.

paulhazen:
> _Something_ about this changeset results in the Editor hanging when trying to run things. I can't pinpoint where that is. When that bug has been addressed, this is approved to go. Had some small comments that you've already addressed.

Good catch! This issue has (I believe) been resolved.

WispyMouse (approved):
Issue resolved. Good to go!

PR #944, opened 2024-10-01
feat: Authentication Listener knows the difference between a Connect and Auth login

WispyMouse:
Previously the Authentication Listeners were receiving a singular message for authentication change. This changeset introduces an enum that identifies a login as either an Auth or Connect login. This will allow for subscribers to run things that require Connect authentication meaningfully separately.

Bundling with this is the relevant area of concern, the Player Data Storage Service, now only querying files on a Connect login.

#EOS-2165

### DIFF
--- a/com.playeveryware.eos/Runtime/Core/AuthenticationListener.cs
+++ b/com.playeveryware.eos/Runtime/Core/AuthenticationListener.cs
@@ -35,6 +35,15 @@ namespace PlayEveryWare.EpicOnlineServices
     public class AuthenticationListener: IAuthInterfaceEventListener, IConnectInterfaceEventListener, IDisposable
     {
         /// <summary>
+        /// Identifies the kind of authentication change.
+        /// </summary>
+        public enum AuthenticationLevelChangeType
+        {
+            Auth,
+            Connect
+        }
+
+        /// <summary>
         /// Used to describe functions that handle change in authentication
         /// state.
         /// </summary>
@@ -42,7 +51,7 @@ namespace PlayEveryWare.EpicOnlineServices
         /// True if the authentication state has changed to authenticated, False
         /// otherwise.
         /// </param>
-        public delegate void AuthenticationChangedEventHandler(bool authenticated);
+        public delegate void AuthenticationChangedEventHandler(bool authenticated, AuthenticationLevelChangeType changeType);
 
         /// <summary>
         /// Event that triggers when the state of authentication has changed.
@@ -104,7 +113,8 @@ namespace PlayEveryWare.EpicOnlineServices
         /// </summary>
         /// <param name="attemptedState"></param>
         /// <param name="attemptResult"></param>
-        private void TriggerAuthenticationChangedEvent(bool attemptedState, Result attemptResult)
+        /// <param name="changeType">The type of authentication change.</param>
+        private void TriggerAuthenticationChangedEvent(bool attemptedState, Result attemptResult, AuthenticationLevelChangeType changeType)
         {
             // If the attempt to change the state of authentication did not 
             // succeed, then log a warning and stop.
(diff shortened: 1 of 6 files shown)


com.playeveryware.eos/Runtime/Core/AuthenticationListener.cs
```diff
@@ -42,7 +51,7 @@ public class AuthenticationListener: IAuthInterfaceEventListener, IConnectInterf
         /// True if the authentication state has changed to authenticated, False
         /// otherwise.
         /// </param>
-        public delegate void AuthenticationChangedEventHandler(bool authenticated);
+        public delegate void AuthenticationChangedEventHandler(bool authenticated, AuthenticationLevelChangeType changeType);
```
paulhazen: Since the introduction and increased use of the `event` concept within our project, I've had the opportunity to learn more about common practices for implementing them effectively. If we're considering expanding or refactoring these implementations, I'd like to propose a refined approach based on these learnings:

## 1. Implement `AuthenticationChangedEventArgs` Class:

Within the `AuthenticationListener`, we can introduce a class called `AuthenticationChangedEventArgs`:

```csharp
public class AuthenticationChangedEventArgs : EventArgs
{
    public AuthenticationChangeOrigin Origin { get; }
    public bool IsAuthenticated { get; }

    public AuthenticationChangedEventArgs(bool isAuthenticated, AuthenticationChangeOrigin origin)
    {
        IsAuthenticated = isAuthenticated;
        Origin = origin;
    }
}
```

## 2. Use the Built-in `EventHandler<TEventArgs>` Delegate:**

By utilizing the built-in `EventHandler<TEventArgs>` delegate, we can simplify our event definition:

```csharp
public event EventHandler<AuthenticationChangedEventArgs> AuthenticationChanged;
```

**3. Update Event Invocation:**

When invoking the `AuthenticationChanged` event, we can adjust the invocation to:

```csharp
AuthenticationChanged?.Invoke(this, new AuthenticationChangedEventArgs(attemptedState, changeType));
```

> [!NOTE]
> The inclusion of the `sender` parameter (typed as `object`) in the `EventHandler<TEventArgs>` delegate allows us to track **who** is raising the event. This is particularly beneficial for debugging and provides flexibility if we need to access the sender down the call chain.

---

**Explanation and Benefits:**

- **Alignment with Standard Practices:** This approach adheres to established .NET conventions for event handling, which can improve code readability and maintainability.
  
- **Simplification:** Using `EventHandler<TEventArgs>` eliminates the need to define custom delegate types, streamlining our codebase.

- **Enhanced Debugging and Extensibility:** Passing the `sender` parameter and utilizing `EventArgs` derivatives provides context and allows for easier extension of event data in the future.

**Final Thoughts:**

I drafted the above without the aid of Visual Studio, so there might be minor implementation details to adjust.

com.playeveryware.eos/Runtime/Core/AuthenticationListener.cs
```diff
@@ -34,6 +34,15 @@ namespace PlayEveryWare.EpicOnlineServices
...
     public class AuthenticationListener: IAuthInterfaceEventListener, IConnectInterfaceEventListener, IDisposable
     {
+        /// <summary>
+        /// Identifies the kind of authentication change.
+        /// </summary>
+        public enum AuthenticationLevelChangeType
+        {
+            Auth,
+            Connect
+        }
```
paulhazen: I think that a more appropriate name for this `enum` would be `AuthenticationChangeOrigin` to indicate from which system the authentication change originated.

paulhazen (changes requested):

com.playeveryware.eos/Runtime/Core/AuthenticationListener.cs
paulhazen: This was discussed within the team - and is something that we should consider in the future - but it was deemed that now was not the right time to consider these changes. Consequently, I am marking this conversation as resolved.

com.playeveryware.eos/Runtime/Core/AuthenticationListener.cs
WispyMouse: This was renamed to "LoginChangeKind". How does that feel?

paulhazen (approved):

andrew-hirata-playeveryware (approved):

PR #945, opened 2024-10-01
fix: Unsubscribe from file list update after leaving the player data storage scene

WispyMouse:
Unsubscribing from the change list event whenever the object is destroyed.

_Normally_ a C# object can eventually have its garbage collected and removed from events. However, the MonoBehaviour gets into a zombie persistence state without the GameObject attached being around.

#EOS-2150


Assets/Scripts/StandardSamples/UI/Storage/Player/UIPlayerDataStorageMenu.cs
```diff
@@ -302,6 +302,13 @@ public override void Hide()
 
             currentSelectedFile = string.Empty;
             currentInventory = null;
+            PlayerDataStorageService.Instance.OnFileListUpdated -= UpdateFileListUI;
```
paulhazen, line 305: I am fairly certain that we want to stay subscribed to the event during the file list update - because when a sample is "hidden," we still want it to get updated when appropriate so that when it is subsequently "shown," it's UI is up-to-date.

paulhazen (changes requested):

Assets/Scripts/StandardSamples/UI/Storage/Player/UIPlayerDataStorageMenu.cs
paulhazen: Per conversation I am going to resolve this conversation - the need to unsubscribe is a peculiarity of the Unity game object lifecycle system.

paulhazen (approved):

andrew-hirata-playeveryware (approved):

PR #946, opened 2024-10-02
Samples/button ux

WispyMouse:
Improved Player Data Storage and Title Data Storage samples by giving their buttons interactive states. They'll automatically become not interactable, or become interactable again, when the passed conditions are met.

image

Please give feedback on:

- Is this the appropriate way to set up an event?
- Is this an appropriate way to call the event?
- Is this an appropriate way to subscribe to the event?
- Is this style of referencing buttons breaking any fundamental UI philosophy?
- What do y'all think about the [SerializeReference] attribute? The purpose is to not expose the UX buttons at a public level, but by being a SerializeReference field it can still be set in the editor.

#EOS-2149


Assets/Scripts/StandardSamples/UI/Common/SampleSelectableStateHandler.cs
```diff
@@ -0,0 +1,108 @@
...
+                }
+            }
+
+            if (attachedSelectable != null)
+            {
+                attachedSelectable.interactable = newState.Interactable;
+            }
+        }
+
+        public static void RaiseSampleSelectableStateChange()
```
andrew-hirata-playeveryware: To me the word 'Raise' implies that an exception or signal is being 'raised'; perhaps we can change it to something like 'notify', 'invoke', 'dispatch'

andrew-hirata-playeveryware (changes requested):

Assets/Scripts/StandardSamples/UI/Common/SampleSelectableStateHandler.cs
WispyMouse: Renamed to NotifySampleSelectableUpdate.

andrew-hirata-playeveryware (approved):

arthur740212 (approved):

Assets/Scripts/StandardSamples/UI/Common/SampleSelectableStateHandler.cs
```diff
@@ -0,0 +1,108 @@
...
+{
+    using System.Collections;
+    using UnityEngine;
+    using UnityEngine.UI;
+    using UnityEngine.EventSystems;
+    using System;
+
+    public class SampleSelectableStateHandler : MonoBehaviour
+    {
+        public struct SampleInteractableNewState
```
paulhazen: I think this name could use some improvement - The _struct_ describes "state," not a "new" state. I think this is another example of a good place to use a class that extends from `EventArgs`:

```c#
public class StateChangeEventArgs : EventArgs { /*...*/ }
```

Assets/Scripts/StandardSamples/UI/Common/SampleSelectableStateHandler.cs
```diff
@@ -0,0 +1,108 @@
...
+                Interactable = shouldBeInteractable;
+                NewTooltipText = newTooltipText;
+            }
+        }
+
+        public delegate SampleInteractableNewState GetSampleInteractable();
+
+        private static event Action UpdateSelectableStates;
+
+        private GetSampleInteractable sampleInteractableFunction { get; set; }
```
paulhazen: This is confusing to me for a number of reasons:
1. It's a private property, but has the word function in it
2. The return type has the word Get in it.

In the very least I think it needs some documentation around it for clarity.

Assets/Scripts/StandardSamples/UI/Common/SampleSelectableStateHandler.cs
```diff
@@ -0,0 +1,108 @@
...
+                    // TODO: Restore the original tooltip text if it is interactable
+                    // At the moment, none of the buttons that this was applied to
+                    // had any tooltips to begin with
+                    attachedTooltip.Text = string.Empty;
+                }
+                else
+                {
+                    attachedTooltip.Text = newState.NewTooltipText;
+                }
+            }
```
paulhazen: Am I correct to infer from this that what this PR does is remove the tooltip from buttons that should not be selectable?

paulhazen (changes requested):

Assets/Scripts/StandardSamples/UI/Common/SampleSelectableStateHandler.cs
WispyMouse: The opposite! This PR:

- Removes tooltips from buttons that are currently Interactable
- Adds tooltips back to things that have become non-interactive
- Determine what tooltip text to use in the above case
- Uses the attached functions to determine if it should be active or not

Assets/Scripts/StandardSamples/UI/Common/SampleSelectableStateHandler.cs
WispyMouse: Sure thing, we'll use that args as the base class.

Assets/Scripts/StandardSamples/UI/Common/SampleSelectableStateHandler.cs
WispyMouse: The naming is pretty confusing, eh? The intention here is supposed to be,

- When this Handler is told to update, run this function.
- The return function will answer if the object should be interactable. If it is, make the attached behavior interactable.
- Otherwise, set the tooltip to this value.

This function is provided by other classes that point to specific UI Elements, giving them specific implementations of this function. So each of these SampleSelectableStateHandlers needs a reference to their own function.

Thus `GetSampleInteractable` should be answering "is this sample element interactable?" What do you suppose we should rename it to?

WispyMouse:
This is being converted to draft to represent how there's a significant piece of work remaining; currently the managers have reference to the state handler ux things, it should be the other way around.

Assets/Scripts/StandardSamples/UI/Storage/Title/UITitleStorageMenu.cs
```diff
@@ -227,5 +232,19 @@ public void UpdateFileContent(Result result)
                 Debug.LogErrorFormat("UITitleStorageMenu - '{0}' file content was not found in cached data storage.", FileNameTextBox.InputField.text);
             }
         }
+
+        public void MustHaveAtLeastOneTagValidator(SelectableStateHandler toUpdate)
```
paulhazen, line 236: image

Is Visual Studio lying to me - or is this function no longer utilized?

WispyMouse: It's only referenced _in the Editor_, not in the csproj. It's referenced in a UnityEvent.

Exactly what you're experiencing right now was the most prominent argument for not doing it the way I did. One counterpoint is that if you used Rider instead of Visual Studio, it'd tell you where it was use. The other is that all things like this will have the same argument, showing what kind of thing it's related to.

paulhazen: I recall that now. Okay.

paulhazen (approved):

Issue #941, opened 2024-10-01
Plugin not working on Android builds


Harsche:
**Describe the bug**
The plugin doesn't work on Android builds, it throws an error when the scene containing the EOSManager script is loaded, and then every call to EOSManager throws a null reference exception.

**To Reproduce**
1. Create a new Unity project
2. Import the EOS plugin and the samples
3. Do the basic configuration required by the plugin and also the setup for Android builds
4. Add any of the sample scenes to the build
5. Build for Android and run
6. See error

**Expected behavior**
No error should appear and the plugin should work fine, which is what happens for the Windows executables I've tested.

**Screenshots**
![Imagem do WhatsApp de 2024-09-30 à(s) 16 47 35_5958e178](https://github.com/user-attachments/assets/78d607b7-2fad-4d22-b8b7-8c10ccafcaab)

**Smartphone**
 - Device: Xiaomi Mi 9
 - OS: Android 10 QKQ1.1190825.002
 - Unity Version: 2022.3.38 (also tested on 2021.3.44)
 - Plugin Version: 3.3.3

**Additional context**
From what I've seen, it appears to be related to using UnityWebRequest from outside the main thread.
Here's the error log:

```
Autoconnected Player "Autoconnected Player" UnityException: Create can only be called from the main thread.
Constructors and field initializers will be executed from the loading thread when loading a scene.
Don't use this function in the constructor or field initializers, instead move initialization code to the Awake or Start function.
  at (wrapper managed-to-native) 
```

```
UnityEngine.Networking.DownloadHandlerBuffer.Create(UnityEngine.Networking.DownloadHandlerBuffer)
  at UnityEngine.Networking.DownloadHandlerBuffer.InternalCreateBuffer () [0x00001] in /home/bokken/build/output/unity/unity/Modules/UnityWebRequest/Public/DownloadHandler/DownloadHandler.bindings.cs:223 
  at UnityEngine.Networking.DownloadHandlerBuffer..ctor () [0x00008] in /home/bokken/build/output/unity/unity/Modules/UnityWebRequest/Public/DownloadHandler/DownloadHandler.bindings.cs:228 
  at UnityEngine.Networking.UnityWebRequest.Get (System.String uri) [0x00001] in /home/bokken/build/output/unity/unity/Modules/UnityWebRequest/Public/WebRequestExtensions.cs:16 
[...]

PR #947, opened 2024-10-03
fix: Android Threading & IO in Unity

paulhazen:
## The Problem: Unique IO Requirements in Unity on Android

In Unity on Android devices, IO operations must be performed using `UnityWebRequest`, whose internal `Create` function must be called from the main thread.

In version `v3.3.3`, we re-implemented `async` functions using `Task.Run` to execute asynchronously. This allowed code to return to the main thread seamlessly and simplified the codebase by reducing branching.

However, this approach doesn't work on Android because `Task.Run` still executes `async` functions on a background thread, conflicting with `UnityWebRequest.Create` needing to run on the main thread.

## The Solution

We reworked the `async` functions to use only synchronous methods, avoiding the use of background threads.

Changes made:

- Updated naming conventions to clearly distinguish between asynchronous and synchronous functions.
- Used compiler conditionals to exclude asynchronous IO functions when running on non-editor Android platforms, preventing accidental usage.
- Replaced synchronous counterparts to asynchronous functions with implementations that do not use background threads, ensuring seamless operation for the caller without violating Android's main-thread requirements.

This PR addresses issue #941 

#EOS-2167


com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -562,12 +562,6 @@ private void InitializeOverlay(IEOSCoroutineOwner coroutineOwner)
             // NOTE: on some platforms the EOS platform is init'd by a native dynamic library. In
             // those cases, this code will early out.
             public void Init(IEOSCoroutineOwner coroutineOwner)
-            {
-                Init(coroutineOwner, EOSPackageInfo.ConfigFileName);
-            }
-
-            //-------------------------------------------------------------------------
-            public void Init(IEOSCoroutineOwner coroutineOwner, string configFileName)
```
andrew-hirata-playeveryware, line 570: I don't think this refactor should be done in this PR.

paulhazen, line 570: This should be restored, because removing it changes the API.

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -758,6 +752,13 @@ public void SetLogLevel(LogCategory Category, LogLevel Level)
             /// </summary>
             private void InitializeLogLevels()
             {
+                // TODO: Restore original implementation for this function, but
```
andrew-hirata-playeveryware: This should be wrapped in an android specific conditional compile.

andrew-hirata-playeveryware (changes requested):

com.playeveryware.eos/Runtime/Core/Utility/FileSystemUtility.cs
```diff
@@ -526,10 +513,27 @@ public static async Task<string> ReadAllTextAsync(string path)
...
+        /// Reads all text from the indicated file.
+        /// </summary>
+        /// <param name="path">Filepath to the file to read from.</param>
+        /// <returns>The contents of the file at the indicated path as a string.</returns>
+        public static string ReadAllText(string path)
+        {
+#if UNITY_ANDROID && !UNITY_EDITOR
+            return AndroidFileIOHelper.ReadAllText(path);
+#else
+            return Task.Run(() => ReadAllTextAsync(path)).GetAwaiter().GetResult();
```
paulhazen: This should be changed to not use the Async function, but to use the synchronous function instead.

Assets/Plugins/Android/Core/AndroidFileIOHelper.cs
```diff
@@ -23,23 +23,43 @@
...
     {
-        public static async Task<string> ReadAllText(string filePath)
+        public static bool FileExists(string filePath)
         {
             using UnityWebRequest request = UnityWebRequest.Get(filePath);
-            request.timeout = 2; //seconds till timeout
-            UnityWebRequestAsyncOperation operation = request.SendWebRequest();
+            request.timeout = 2;
+            request.SendWebRequest();
+            while (!request.isDone) { }
```
WispyMouse, line 38: I was pretty uncertain about this, but it appears as though this works properly. I made this testing class to validate the behaviour, and it correctly does finish. The app appears to be locked up until it's done, but that's the intentional and desired synchronous action.

```csharp
    private void Update()
    {
        if (Input.GetKeyDown(KeyCode.A))
        {
            BeginTest();
        }
    }

    private void BeginTest()
    {
        StatusText.text = "Beginning Test";

        using UnityWebRequest request = UnityWebRequest.Get(@"REALLY HUGE ZIP FILE");
        request.timeout = 2;
        request.SendWebRequest();

        StatusText.text = "Waiting for test to finish";

        while (!request.isDone) { }

        StatusText.text = $"Test finished {request.result}";
    }
```

com.playeveryware.eos/Runtime/Core/Utility/FileSystemUtility.cs
paulhazen: This has been done now. Resolving conversation.

com.playeveryware.eos/Runtime/Core/EOSManager.cs
paulhazen: The original implementation has been restored. I'm marking this conversation as resolved.

com.playeveryware.eos/Runtime/Core/EOSManager.cs
WispyMouse: This sounds like there's a known issue with Android using SetLogLevel. Does it not work with it? Can people still define custom log levels for Android?

WispyMouse (approved):

com.playeveryware.eos/Runtime/Core/EOSManager.cs
paulhazen: The issue with SetLogLevel is unique to Android, but resolving it requires refactoring some of the LogLevel code - which is outside the scope of this PR.

Functionality has been changed so that it uses an Android specific conditional for compile time, and I'll add a task to address it for Android specifically at a later time.

andrew-hirata-playeveryware (approved):

PR #953, opened 2024-10-08
fix: Native Enum Parsing

paulhazen:
In a recent change to the native parsing, an error was made in the continue conditional. This PR resolves that issue and adds some comments to make it clearer what is happening.


andrew-hirata-playeveryware (approved):

WispyMouse (changes requested):
Discussed offline, but essentially;

I believe strings_to_enum_values has c_strings. Map->find will compare the reference rather than the value. This results in failing to parse any strings, because it won't think anything is in the list.

This needs to use a strcmp or similar way of comparing c_strings. We could also change the key from a `const char*` to a `std::string` perhaps?

paulhazen:
> Discussed offline, but essentially;
> 
> I believe strings_to_enum_values has c_strings. Map->find will compare the reference rather than the value. This results in failing to parse any strings, because it won't think anything is in the list.
> 
> This needs to use a strcmp or similar way of comparing c_strings. We could also change the key from a `const char*` to a `std::string` perhaps?

This has been resolved now. Changing the key from `const char*` to `std::string` did the trick, good catch on that @WispyMouse!

WispyMouse (approved):

PR #955, opened 2024-10-10
Fix: `IOSBuilder` not using correct `FileSystemUtility` class

leoleotte:
Looks like an old usage of this Utility class has been leftover, this is causing erros when trying to use the plugin with iOS as the selected platform.


paulhazen (approved):
Great catch!

WispyMouse (approved):

PR #962, opened 2024-10-16
feat(connect,google) : Connect Login with Google ID

arthur740212:
First pass of google ID connect login.
GetCredentialAsync parameters (Client ID and nonce) can be configured in AndroidConfig

Further split SignInWithGoogleManager into its own assembly? 
Answer : Not necessary, because there is no external dependency. Unless we want to make an upm just for the login part.

EOS-2170

### DIFF
--- a/Assets/Plugins/Android/Core/AndroidConfig.cs
+++ b/Assets/Plugins/Android/Core/AndroidConfig.cs
@@ -25,6 +25,8 @@ namespace PlayEveryWare.EpicOnlineServices
     // Flags specifically for Android
     public class AndroidConfig : PlatformConfig
     {
+        public string GoogleLoginClientID;
+        public string GoogleLoginNonce;
         static AndroidConfig()
         {
             RegisterFactory(() => new AndroidConfig());
--- a/Assets/Plugins/Android/login.java
+++ b/Assets/Plugins/Android/login.java
@@ -1,0 +1,121 @@
+/*
+* Copyright (c) 2024 PlayEveryWare
+* 
+* Permission is hereby granted, free of charge, to any person obtaining a copy
+* of this software and associated documentation files (the "Software"), to deal
+* in the Software without restriction, including without limitation the rights
+* to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+* copies of the Software, and to permit persons to whom the Software is
+* furnished to do so, subject to the following conditions:
+* 
+* The above copyright notice and this permission notice shall be included in all
+* copies or substantial portions of the Software.
+* 
+* THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+* IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+* FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+* AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+* LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+* OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+* SOFTWARE.
+*/
+
+package com.playeveryware.googlelogin;
+
+import com.google.android.libraries.identity.googleid.GetSignInWithGoogleOption;
+import com.google.android.libraries.identity.googleid.GoogleIdTokenCredential;
+import com.google.android.libraries.identity.googleid.GoogleIdTokenParsingException;
+
+import androidx.credentials.GetCredentialRequest;
+import androidx.credentials.GetCredentialResponse;
+import androidx.credentials.exceptions.GetCredentialException;
+import androidx.credentials.Credential;
+import androidx.credentials.CustomCredential;
+import androidx.credentials.CredentialManager;
+import androidx.credentials.CredentialManagerCallback;
+
+import android.app.Activity;
+import android.content.Context;
+import android.os.CancellationSignal;
+import android.util.Log;
+
+import java.util.concurrent.Executors;
+
+
+public class login extends Activity
+{
+    private String name;
+    private String token;
+    private static login instance;
+
+    public login()
+    {
+        this.instance = this;
+    }
+    public static login instance()
+    {
+        if(instance == null)
+        {
+            instance = new login();
+        }
+        return instance;
+    }
+
+    public String getResultName()
+    {
+        return name;
+    }
+    public String getResultIdToken()
+    {
+        return token;
+    }
+
+    public void SignInWithGoogle(String clientID, String nonce, Context context, CredentialManagerCallback callback)
+    {
+        GetSignInWithGoogleOption signInWithGoogleOption = new GetSignInWithGoogleOption.Builder(clientID)
+                .setNonce(nonce)
+                .build();
+
+        GetCredentialRequest request = new GetCredentialRequest.Builder()
+                .addCredentialOption(signInWithGoogleOption)
+                .build();
+
+        CredentialManager credentialManager = CredentialManager.create(this);
+        credentialManager.getCredentialAsync(context,request,new CancellationSignal(),Executors.newSingleThreadExecutor(),callback);
+    }
+
+    public void handleFailure(GetCredentialException e)
+    {
+        Log.e("Unity", "Received an invalid google id token response", e);
+    }
+
+    public void handleSignIn(GetCredentialResponse result)
+    {
+        Credential credential = result.getCredential();
+
+        if (credential instanceof CustomCredential) 
+        {
+            if (GoogleIdTokenCredential.TYPE_GOOGLE_ID_TOKEN_CREDENTIAL.equals(credential.getType())) 
+            {
+                try
+                {
+                    GoogleIdTokenCredential googleIdTokenCredential = GoogleIdTokenCredential.createFrom(credential.getData());
+                    name = googleIdTokenCredential.getDisplayName();
+                    token = googleIdTokenCredential.getIdToken();
+                }
+                catch (Exception e)
+                {
+                    if (e instanceof GoogleIdTokenParsingException)
+                    {
+                        Log.e("Unity", "Received an invalid google id token response", e);
+                    }
+                    else
+                    {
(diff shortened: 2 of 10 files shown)


Assets/Scripts/StandardSamples/SignInWithGoogle/SignInWithGoogleManager.cs
```diff
@@ -0,0 +1,101 @@
...
+            loginObject = loginClass.CallStatic<AndroidJavaObject>("instance");
+
+            /// Create the proxy class and pass instances to be used by the callback
+            EOSCredentialManagerCallback javaCallback = new EOSCredentialManagerCallback();
+            javaCallback.loginObject = loginObject;
+            javaCallback.callback = callback;
+
+            AndroidConfig config = AndroidConfig.Get<AndroidConfig>();
+            /// SignInWithGoogle(String clientID, String nonce, Context context, CredentialManagerCallback callback)
+            loginObject.Call("SignInWithGoogle", config.GoogleLoginClientID, config.GoogleLoginNonce, activity, javaCallback);
```
WispyMouse, line 64: Seems like config.GoogleLoginClientID and config.GoogleLoginNonce are required. We should consider making validator attributes or running some kind of test on this, or otherwise informing the user if these are null or empty.

Assets/Plugins/Android/login.java
```diff
@@ -0,0 +1,121 @@
+/*
```
WispyMouse, line 1: What is the origin of this file? If it is something we made, then maybe there's some stylistic things to consider, but I also don't know Java conventions well.

Assets/Scripts/StandardSamples/SignInWithGoogle/SignInWithGoogleManager.cs
```diff
@@ -0,0 +1,101 @@
...
+        {
+            /// Parses the response resilt into google credentials
+            loginObject.Call("handleSignIn",credentialResponseResult);
+
+            /// Fetches neccesary data for logging into EOS
+            name = loginObject.Call<string>("getResultName");
+            idToken = loginObject.Call<string>("getResultIdToken");
+
+            /// Invoke Connect Login with fetched Google ID
+            callback.Invoke(idToken, name);
```
WispyMouse: Would it make sense to validate that name and idToken are not null at this time?

Assets/Scripts/StandardSamples/SignInWithGoogle/SignInWithGoogleManager.cs
```diff
@@ -0,0 +1,101 @@
...
+        /// Proxy class to receive Android callbacks in C#
+        /// </summary>
+        public EOSCredentialManagerCallback() : base("androidx.credentials.CredentialManagerCallback") { }
+
+        /// <summary>
+        /// Succeeding Callback of GetCredentialAsync  
+        /// GetCredentialAsync is called in com.playeveryware.googlelogin.login.SignInWithGoogle)
+        /// </summary>
+        /// <param name="credentialResponseResult"></param>
+        public void onResult(AndroidJavaObject credentialResponseResult)
```
WispyMouse: Since we're back in C# land, can this be `OnResult`? Or does it need to match the capitalization?

Assets/Scripts/StandardSamples/SignInWithGoogle/SignInWithGoogleManager.cs
```diff
@@ -0,0 +1,101 @@
...
+* LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+* OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+* SOFTWARE.
+*/
+
+using UnityEngine;
+
+namespace PlayEveryWare.EpicOnlineServices.Samples
+{
+    class EOSCredentialManagerCallback : AndroidJavaProxy
```
WispyMouse: It feels strange for a "class" to be called "callback", but it might have to be this way?

Assets/Scripts/StandardSamples/UI/Login/UILoginMenu.cs
```diff
@@ -687,11 +687,15 @@ private void ConfigureUIForConnectLogin()
             {
 
                 //case ExternalCredentialType.GogSessionTicket:
-                //case ExternalCredentialType.GoogleIdToken:
                 //case ExternalCredentialType.ItchioJwt:
                 //case ExternalCredentialType.ItchioKey:
                 //case ExternalCredentialType.AmazonAccessToken:
-
+#if !UNITY_ANDROID || UNITY_EDITOR
```
WispyMouse, line 693: This makes me wonder on a philosophy thing for our login screen. Should we present all options, even ones not valid on the platform?

Assets/Scripts/StandardSamples/UI/Login/UILoginMenu.cs
```diff
@@ -1171,6 +1179,16 @@ private void CreateDeviceCallback(ref Epic.OnlineServices.Connect.CreateDeviceId
             }
         }
 
+        private void ConnectGoogleId()
+        {
+            SignInWithGoogleManager signInWithGoogleManager = new();
+
+            signInWithGoogleManager.GetGoogleIdToken((string token, string username) =>
```
WispyMouse, line 1186: Looking at `GetGoogleIdToken`, it seems as though the callback is passed to java. The callback isn't run if there's some critical failure during set up. Should we teach the callback to report in if there's an error, so that can be logged to the user? Or is it enough that inside GetGoogleIdToken exceptions and such are printed?

WispyMouse (approved):
Left some thoughts. The java portion feels like it could use more elegant spacing and capitalization, but maybe that's just a java thing?

Assets/Plugins/Android/Core/AndroidConfig.cs
```diff
@@ -25,6 +25,8 @@ namespace PlayEveryWare.EpicOnlineServices
     // Flags specifically for Android
     public class AndroidConfig : PlatformConfig
     {
+        public string GoogleLoginClientID;
```
andrew-hirata-playeveryware, line 28: I don't know if we want to have these in the Android Config: for the time being, I would be okay if the values are in the sample code and need to be filled out by the person running the code.

Assets/Scripts/StandardSamples/SignInWithGoogle/SignInWithGoogleManager.cs
andrew-hirata-playeveryware: It has to match the java interface that it's adopting.

Assets/Scripts/StandardSamples/SignInWithGoogle/SignInWithGoogleManager.cs
andrew-hirata-playeveryware: It doesn't have to be that way for the Android helper stuff to work, but it makes it clearer (to me at least) that it is implementing one of the callback classes.

Assets/Scripts/StandardSamples/UI/Login/UILoginMenu.cs
andrew-hirata-playeveryware: At one time we were thinking about having them listed but disabled / grey'd out.

Assets/Scripts/StandardSamples/UI/Login/UILoginMenu.cs
andrew-hirata-playeveryware: I think having a way to invoke the callback in the error case makes sense. Good catch.

andrew-hirata-playeveryware (changes requested):

Assets/Plugins/Android/login.java
arthur740212: I wrote this from scratch. And I am not familiar with any java conventions

Assets/Scripts/StandardSamples/UI/Login/UILoginMenu.cs
arthur740212: Currently, all options will be shown in UI on every platform.
But there will be a message implying "Unsupported" if the platform isn't right

Assets/Scripts/StandardSamples/SignInWithGoogle/SignInWithGoogleManager.cs
arthur740212: Added it in the lambda.

Assets/Scripts/StandardSamples/SignInWithGoogle/SignInWithGoogleManager.cs
arthur740212: Added a null check on ClientID.
Nonce can be null.

andrew-hirata-playeveryware (approved):

paulhazen (approved):
My only suggestion is to consider putting the Android Proxy class thing into the SignInWithGoogleManager class as a nested class.

Assets/Scripts/StandardSamples/SignInWithGoogle/SignInWithGoogleManager.cs
```diff
@@ -94,6 +95,12 @@ public void GetGoogleIdToken(System.Action<string, string> callback)
             javaCallback.callback = callback;
 
             AndroidConfig config = AndroidConfig.Get<AndroidConfig>();
+
+            if (config.GoogleLoginClientID == null)
```
WispyMouse: Can this check String.IsNullOrEmpty instead of checking directly to null?

Assets/Scripts/StandardSamples/UI/Login/UILoginMenu.cs
```diff
@@ -1185,6 +1185,12 @@ private void ConnectGoogleId()
 
             signInWithGoogleManager.GetGoogleIdToken((string token, string username) => 
             {
+                if (token == null)
```
WispyMouse: Can this check String.IsNullOrEmpty instead of direct null comparison?

WispyMouse (changes requested):
I think the two null checks introduced should check for String.IsNullOrEmpty, not just directly null.

Assets/Scripts/StandardSamples/SignInWithGoogle/SignInWithGoogleManager.cs
arthur740212: Done!

Assets/Scripts/StandardSamples/UI/Login/UILoginMenu.cs
arthur740212: Done!

WispyMouse (approved):

PR #964, opened 2024-10-18
Config Override System

paulhazen:
This is the first PR in a multi-PR effort to expand, simplify, and make clearer the mechanism by which different settings are applied to different platforms. 

In this PR the data within the config that pertains to the _Product_ the plugin is for (as defined under "Product Information" within Epic Dev Portal). Has been extracted into a `ProductConfig` class. These settings are shared and all configurations regardless of platform will need to utilize them.

> [!WARNING]
> No config values have been removed from where they currently reside - until we are prepared to turn this feature on, they will remain.
>
> That said, this PR _does_ include implementation to automatically import the old config values into the new config. 

The functionality of this change has been disabled by default, until a time that more components of the override system are in place, and we are able to turn them all on at once.

This window is currently disabled, but if you enable it, it gives you an idea of the changed behavior:
image

#EOS-1982


com.playeveryware.eos/Runtime/Core/Common/SetOfNamed.cs
```diff
@@ -0,0 +1,178 @@
...
+
+    /// <summary>
+    /// Class that stores a unique set of values, where each value has a name
+    /// associated to it. Items are sorted by the name. No two items of the same
+    /// value can be added (regardless of if they have different names). Also,
+    /// no two items can have the same name (regardless of if they have
+    /// different values).
+    /// </summary>
+    /// <typeparam name="T">The type being wrapped with a name.</typeparam>
+    public class SetOfNamed<T> : HashSet<Named<T>> where T : IEquatable<T>
```
paulhazen, line 36: Would be good to collect input on whether this is a code smell (inheritance vs composition).

Assets/Plugins/Source/Editor/EditorWindows/NewConfigWindow.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        protected override async Task AsyncSetup()
+        {
+            _productConfigEditor = new();
+            await _productConfigEditor.LoadAsync();
+        }
+
+        protected override void RenderWindow()
+        {
+            _ = _productConfigEditor.RenderAsync();
```
WispyMouse, line 64: There's something weird with this, "synchronously open a UI, asynchronously load stuff, take actions before done", kind of approach... but I don't think Unity's UI system really gives you a much better way to do things. Here this is fine, I don't strongly feel it should change.

Some part of me wonders if we should use "ContinueWith" to set some kind of, "okay this screen is now ready for you to do things" state, and before that's finished the user can't take actions.

Assets/Plugins/Source/Editor/Utility/GUIEditorUtility.cs
```diff
@@ -211,8 +213,480 @@ public static void AssigningEnumField<T>(string label, ref T value, float labelW
...
+        /// </summary>
+        /// <typeparam name="T">
+        /// The type of input being rendered.
+        /// </typeparam>
+        /// <param name="rect">
+        /// The area in which the field should be rendered.
+        /// </param>
+        /// <param name="value">The value to put in the field.</param>
+        /// <returns>The value entered in the field.</returns>
+        private delegate T RenderBasicFieldDelegate<T>(Rect rect, T value);
```
WispyMouse, line 246: Does "basic" imply "does not contain GUILayoutOptions and instead has Rect"? What makes something "Basic"?

paulhazen: It's a bit of a historical thing. Unity's editor UI toolkit is built upon ImGUI and has undergone several refactors such that Unity actually has a few different ways to render things in ImGUI. The _newer_ method does not require that you specify the `Rect` to draw in, but the _older_ (hence I'm calling "basic") method _does_. 

The control that I use to render lists requires use of the old set of methods instead of the new ones - so when the delegate needs to be used in that context, it will have different parameters.

Assets/Plugins/Source/Editor/EditorWindows/NewConfigWindow.cs
paulhazen: Unity's UI system (used for rendering things in scenes) is distinct from Unity's UI Toolkit (used for rendering things in the editor). These functions take place in the context of the latter. The two technologies have very different origins and hence different design paradigms.

The distinction between these different systems is documented by Unity [here](https://docs.unity3d.com/Manual/UIToolkits.html).

com.playeveryware.eos/Runtime/Core/Common/SetOfNamed.cs
paulhazen: Input collected - consensus was that this is an acceptable use of inheritance. Consequently, I am closing this conversation.

WispyMouse, line 36: I believe HashSet uses IEqualityComparer<T> if it is available. If it isn't available, then it's doing a more expensive check *I think*. Can you consider requiring T to implement IEqualityComparer?

If you'd like I can whip up a demo that shows the utility of this suggestion.

com.playeveryware.eos/Runtime/Core/Common/SetOfNamed.cs
```diff
@@ -0,0 +1,178 @@
...
+        /// </param>
+        /// <returns>
+        /// True if there is an item in the collection with the indicated value,
+        /// false otherwise.
+        /// </returns>
+        public bool Contains(T item)
+        {
+            foreach (Named<T> namedItemInSet in this)
+            {
+                if (namedItemInSet.Value.Equals(item))
```
WispyMouse, line 175: If .Value is null, will this have a nullreference exception? Can it ever be null?

com.playeveryware.eos/Runtime/Core/Config/Attributes/ConfigFieldType.cs
```diff
@@ -75,6 +75,26 @@ public enum ConfigFieldType
         /// </summary>
         TextList,
 
+        /// <summary>
+        /// A Guid value that is named.
```
WispyMouse: I feel like if I read this sentence, I still don't "understand". Could this comment include an example?

com.playeveryware.eos/Runtime/Core/Common/SetOfNamed.cs
paulhazen: Per offline conversation - given that these collections will 99.99% of the time contain at most 2 items, the cost of the equality check is not a significant enough concern to warrant further development. But it's good to be aware of, thank you for calling it to our attention! :grin:

com.playeveryware.eos/Runtime/Core/Config/Deployment.cs
```diff
@@ -0,0 +1,64 @@
...
+
+    public struct Deployment : IEquatable<Deployment>
+    {
+        public SandboxId SandboxId;
+
+        public Guid DeploymentId;
+
+        public bool Equals(Deployment other)
+        {
+            return SandboxId.Equals(other.SandboxId) && DeploymentId.Equals(other.DeploymentId);
```
WispyMouse, line 35: Can either of these ever be *null*? Would this possibly cause a nullref?

com.playeveryware.eos/Runtime/Core/Config/Deployment.cs
```diff
@@ -0,0 +1,64 @@
...
+            {
+                return false;
+            }
+
+            if (ReferenceEquals(this, obj))
+            {
+                return true;
+            }
+
+            if (obj.GetType() != this.GetType())
```
WispyMouse: Could we reduce one boxing operation by saying "if (!obj is Deployment deployment)" and then later using that in the Equals operation?

com.playeveryware.eos/Runtime/Core/Config/ProductConfig.cs
```diff
@@ -0,0 +1,172 @@
...
+                Value = config.sandboxID
+            };
+
+            Deployment oldDeployment = new()
+            {
+                DeploymentId = Guid.Parse(config.deploymentID),
+                SandboxId = sandboxId
+            };
+
+            if (!Environments.AddDeployment(oldDeployment))
```
WispyMouse, line 135: Is there ever a world where this is null, because it hasn't set PrepareConfig to initialize the default not-null instance?

com.playeveryware.eos/Runtime/Core/Config/SandboxId.cs
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        public override bool Equals(object obj)
+        {
+            if (obj is null)
+            {
+                return false;
+            }
+
+            if (obj.GetType() != this.GetType())
```
WispyMouse: I think we can do the same unboxing trick here. "is SandboxId otherCasted" and then reuse otherCasted.

com.playeveryware.eos/Runtime/Core/Config/SandboxId.cs
```diff
@@ -0,0 +1,82 @@
...
+                    }
+                }
+
+                _value = value;
+            }
+        }
+
+        public readonly bool Equals(SandboxId other)
+        {
+            return _value == other._value;
```
WispyMouse, line 59: Should we use "==" or `_value?.Equals(other._value)`?

com.playeveryware.eos/Runtime/Core/Config/ProductConfig.cs
```diff
@@ -0,0 +1,172 @@
...
+        {
+            // Import explicitly set sandbox and deployment
+            SandboxId sandboxId = new()
+            {
+                Value = config.sandboxID
+            };
+
+            Deployment oldDeployment = new()
+            {
+                DeploymentId = Guid.Parse(config.deploymentID),
```
WispyMouse, line 131: Is there ever a world where we want to use Guid.TryParse instead of Guid.Parse? I wonder what would happen in this script if the previous eos config was not null, but did not fill out its deployment id. This would be an unusual state for a user to be in.

com.playeveryware.eos/Runtime/Core/Config/ProductionEnvironments.cs
```diff
@@ -0,0 +1,128 @@
...
+
+        /// <summary>
+        /// Adds a Sandbox to the Production Environment.
+        /// </summary>
+        /// <param name="sandbox">
+        /// The sandbox to add.</param>
+        /// <returns>
+        /// True if the sandbox was added successfully, false otherwise.
+        /// </returns>
+        public bool AddSandbox(Named<SandboxId> sandbox = null)
```
WispyMouse: It feels weird that we can add a null sandbox. Why would we want to do this?

com.playeveryware.eos/Runtime/Core/Config/ProductionEnvironments.cs
```diff
@@ -0,0 +1,128 @@
...
+        /// The Sandbox to remove from the production environment.
+        /// </param>
+        /// <returns>
+        /// True if the Sandbox was removed, false otherwise. If there is a
+        /// defined deployment that references the Sandbox, then removing it is
+        /// disallowed.
+        /// </returns>
+        public bool RemoveSandbox(Named<SandboxId> sandbox)
+        {
+            foreach (Named<Deployment> deployment in Deployments)
```
WispyMouse, line 77: Could SetOfNamed<T>.Contains be leveraged here?

Assets/Plugins/Source/Editor/EditorWindows/NewConfigWindow.cs
WispyMouse, line 1: It isn't clear to me from the UI what each field is. While it can be pretty easy to follow through (EOS Plugin for Unity is the name of the plugin, and the next guid is the ID) I feel like it suggests we really want labels. Perhaps mouseover labels that can be set to describe what the field is.

WispyMouse, line 1: When I insert an invalid GUID into the Production Information pane, the window immediately unsets it. If there's validation being performed that tells it my provided configuration is invalid, I want that to be surface to the user through at least a debug log.

WispyMouse, line 1: The description of a field, specifically for sandboxes, does not word wrap. The explanation is longer than my window. Could it word wrap?

WispyMouse, line 1: It wasn't immediately clear to me that the Deployments list associates with a sandbox that has been entered above, because the default value is "Sandbox". It become apparent to me after entering a differnet name for the sandbox. I'm unsure if this suggests a UX change is needed.

WispyMouse (approved):
I have a bunch of thoughts and such, but nothing that would block accepting this work. It looks nice!

Assets/Plugins/Source/Editor/EditorWindows/NewConfigWindow.cs
paulhazen: Validation UI feedback is something that's coming, but not currently implemented. I'll create a task for this and close this conversation once I do.

paulhazen: I will look into this.

paulhazen: I understand the concern - I believe it's acceptable.

paulhazen: I'm not going to make this change just yet - because the config window is disabled by default and does not represent the final user interface.

paulhazen: image
image
image

Good idea, done!

com.playeveryware.eos/Runtime/Core/Common/SetOfNamed.cs
paulhazen: Good catch, fixed this.

com.playeveryware.eos/Runtime/Core/Config/Attributes/ConfigFieldType.cs
paulhazen: I clarified the documentation.

com.playeveryware.eos/Runtime/Core/Config/Deployment.cs
paulhazen: Neither of these can ever be null because they are `struct` types.

com.playeveryware.eos/Runtime/Core/Config/Deployment.cs
paulhazen: Good idea! Even better you can (and I did) do this:
```c#
if (obj is not Deployment deployment)
{
    return false;
}

return Equals(deployment);
```

Further refinement of the `Equals` function yields an even more elegant solution:
```c#
public override bool Equals(object obj)
{
    return obj is Deployment deployment && Equals(deployment);
}
```

com.playeveryware.eos/Runtime/Core/Config/ProductConfig.cs
paulhazen: The reason I chose _not_ to use `TryParse` is because if the `deploymentID` from the previous `EOSConfig` cannot be parsed to a Guid, then errors will already have been thrown - I didn't want to complicate matters by explaining to the user that their `deploymentID` value is wrong outside the context of it being used. The import code should only happen if there are existing values in the old format.

com.playeveryware.eos/Runtime/Core/Config/ProductConfig.cs
paulhazen: Because these import functions are all called just after the line 
```c#
Environments ??= new();
```
it is not possible for this to be `null`. Since the import functions are `private`, it's not possible for others to improperly call the function at a time when it would be `null`.

com.playeveryware.eos/Runtime/Core/Config/ProductionEnvironments.cs
paulhazen: Good point, I am not certain why I have that default set. I removed it.

com.playeveryware.eos/Runtime/Core/Config/ProductionEnvironments.cs
paulhazen: Not to a degree that would be beneficial. The only way a sandbox is removed is by passing the whole object - since that's how `HashSet` works.

However, in case between the time the list is rendered and the time this code executes the Sandbox in question has been renamed - then the removal will still function properly since it _only_ looks at the `SandboxId` value.

com.playeveryware.eos/Runtime/Core/Config/SandboxId.cs
paulhazen: The reason `==` is used here is because two `SandboxId` structs might _both_ have `_value` field members that are `null`, in which case they should be considered equal.

com.playeveryware.eos/Runtime/Core/Config/SandboxId.cs
paulhazen: Yep, good call. Updated to reflect the appropriate change.

Assets/Plugins/Source/Editor/EditorWindows/NewConfigWindow.cs
WispyMouse: Ah, I may have over-used a word. I'm unconcerned with what happens here at runtime, just editor time. It's *concerning to me* that at editor time, the developer implementing our plugin *could theoretically* run actions that the editor window is not yet ready to receive because asynchronous operations are running. I wish we could design our editor tools in a way that this is not possible.

However, several of our windows already work like this. I'm not greatly concerned, and I also don't have a solution for what to do for the UX. Maybe in years from now where we're doing all sorts of even the smallest UX improvements, my idea here could be implemented. I'm not gonna block this PR on the idea.

com.playeveryware.eos/Runtime/Core/Config/Config.cs
```diff
@@ -208,6 +218,11 @@ public static async Task<T> GetAsync<T>() where T : Config
             // Use the factory method to create the config.
             T instance = (T)factory();
 
+            if (!await FileSystemUtility.FileExistsAsync(instance.FilePath) && instance._allowDefaultIfFileNotFound)
```
arthur740212, line 221: Will this break in Android?

com.playeveryware.eos/Runtime/Core/Common/SetOfNamed.cs
WispyMouse: Sure, I can agree with that.

I *want* to have the time to make everything as optimally implemented as possible. There's this story in my head where a user who is relatively new to programming is learning through struggle, continuously running into walls until they can accomplish what they want. For this theoretical user, I'd want every bit of code they see to be a learning moment. If they want to implement a HashSet, similar to how you've done here, they'd learn how to do it *the right way* by looking at it.

But sometimes *the right way* is to make something that fits the need, not something that fits every theoretical angle.

Assets/Plugins/Source/Editor/EditorWindows/NewConfigWindow.cs
WispyMouse: Risk assessed, accepted, good to go.

WispyMouse: Fair! I'll remember this for the next time this is being iterated on, I'll want it fixed by then, but I don't want to block on this.

com.playeveryware.eos/Runtime/Core/Config/Attributes/ConfigFieldType.cs
WispyMouse: Oh awesome, this makes way more sense to me.

com.playeveryware.eos/Runtime/Core/Config/Deployment.cs
WispyMouse: It appears as though SandboxId, DeploymentId, Deployment, string, and Guid are all structs. So you're right, these can't possibly be null.

arthur740212 (approved):

com.playeveryware.eos/Runtime/Core/Config/Config.cs
paulhazen: It will not break in Android because `GetAsync` itself is not available in Android.

com.playeveryware.eos/Runtime/Core/Config/Deployment.cs
WispyMouse: "not"! *"not"*?? You can just use English, what? I love it, even if your refined idea doesn't utilize it. I'm likely to implement things the way you have it here from now on.

com.playeveryware.eos/Runtime/Core/Config/ProductConfig.cs
WispyMouse: Sounds believable, ship it.

com.playeveryware.eos/Runtime/Core/Config/SandboxId.cs
WispyMouse: I think I'd still want to use .Equals, it can be significant. Perhaps you could have an if statement for "is _value == null && other._value == null".

Actually since _value is a string, could you use `String.IsNullOrEmpty` on both instead of null checking directly?

Assets/Plugins/Source/Editor/EditorWindows/NewConfigWindow.cs
WispyMouse: Oh excellent, this is a good implementation for the problem.

com.playeveryware.eos/Runtime/Core/Config/SandboxId.cs
paulhazen: The following statement:
```c#
_value == null && other._value == null
```
I believe will _always_ evaluate to the same value as:
```c#
_value == other._value
```
Same thing as 
```c#
string.IsNullOrEmpty(_value) && string.IsNullOrEmpty(other_.value)
```

I believe that all three lines will in all cases evaluate to the same `true` or `false` value. Is there an edge case I'm forgetting?

paulhazen: Ah - I think you are correct here. I'm also correct that those three lines will always come to the same evaluation. The distinction is in how strings might be compared - and allows (I believe) for culture-specific considerations when evaluating string equivalency. I'll update the implementation accordingly. Interesting catch and a good reminder, thanks Mint! :)

paulhazen: Okay I did more research and discovered that I was wrong - string comparison overrides the `==` operator to properly do the value comparison between the two. Several reputable sources lead me to believe the original implementation using the `==` operator is the correct approach.

PR #965, opened 2024-10-18
 Fix/deployment override

arthur740212:
fix(init,config) : Remove false alarm when sandbox override is empty

#EOS-2187

Should probably do #EOS-2190 as well before the next release as they are highly related.


paulhazen (approved):

WispyMouse (approved):

WispyMouse (changes requested):
As discussed in person: It appears as though our own EOS Project for this plugin uses an old deployment system. I suspect, based on the work that introduced the changes that this removes, that EOS may have a _different_ way of deploying for older projects. You'll require an Epic project with permission to see Epic Games Store's drop down and release management pipeline. Please test it in there, to validate that this code isn't necessary.

WispyMouse:
Or alternatively, remove the component of this PR that is related to the native code.

WispyMouse (approved):

PR #968, opened 2024-10-23
Config Override System

paulhazen:
This PR introduces the ability of `PlatformConfigEditor<>` to be used without having specific implementations for each platform and places a lot of the configuration values for overrides into the `PlatformConfig` class.

These changes affect the user interface within the `EOSSettingsWindow`'s platform settings, and migrates old configuration values into the new `PlatformConfig` implementation.

#EOS-1982


Assets/Plugins/Android/Core/AndroidConfig.cs
```diff
@@ -25,8 +25,21 @@ namespace PlayEveryWare.EpicOnlineServices
...
+            ConfigFieldType.Text,
+            "Get your project's Google Login Client ID from the " +
+            "Google Cloud dashboard",
+            -1,
+            "https://console.cloud.google.com/apis/dashboard")]
         public string GoogleLoginClientID;
+
+        [ConfigField("Google Login Nonce",
+            ConfigFieldType.Text,
+            "Use a nonce to improve sign-ing security on Android.",
```
WispyMouse, line 38: Shall tooltips end in punctuation, or shall they end without it?

Assets/Plugins/Source/Editor/EditorWindows/EOSSettingsWindow.cs
```diff
@@ -148,18 +148,40 @@ protected override async Task AsyncSetup()
...
+                // Windows platform overrides are currently not displayed,
+                // because the values for Windows are read from the main
+                // EpicOnlineServicesConfig.json file.
+                if (platform == PlatformManager.Platform.Windows)
+                {
                     continue;
+                }
 
-                toolbarStrings.Add(editor.GetLabelText());
+                if (!PlatformManager.TryGetConfigType(platform, out Type configType) || null == configType)
```
WispyMouse, line 162: It feels weird to have a TryGet that also checks if the out was null. Shouldn't PlatformManager.TryGetConfigType just never return a null value?

Assets/Plugins/Source/Editor/EditorWindows/NewConfigWindow.cs
```diff
@@ -21,12 +21,13 @@
  */
 
 // Uncomment the following line to see the experimental new config window
-//#define ENABLE_NEW_CONFIG_WINDOW
+#define ENABLE_NEW_CONFIG_WINDOW
```
WispyMouse: Is this intentionally ready to be defined by default? Is maintaining this config bit a Strangler Fig thing?

com.playeveryware.eos/Runtime/Core/Config/SandboxId.cs
```diff
@@ -38,6 +38,13 @@ readonly get
             }
             set
             {
+                // allow null value
+                if (value == null)
```
WispyMouse, line 42: Given that the string is the value provided, could this use "IsNullOrEmpty", and sanitize that to a standard either empty string or null?

com.playeveryware.eos/Runtime/Core/PlatformManager.cs
```diff
@@ -38,20 +37,23 @@ public static class PlatformManager
...
+            XboxSeriesX = 0x8,
+            iOS = 0x10,
+            Linux = 0x20,
+            macOS = 0x40,
+            PS4 = 0x80,
+            PS5 = 0x100,
+            Switch = 0x200,
+            Steam = 0x400,
+            Console = PS4 | PS5 | XboxOne | XboxSeriesX | Switch,
+            All = Unknown | Windows | Android | XboxOne | XboxSeriesX | iOS | Linux | macOS | PS4 | PS5 | Switch | Steam
```
WispyMouse: `All` including `Unknown` is a bit funny to me. It's a flag, so _every value_ includes Unknown, doesn't it?

com.playeveryware.eos/Runtime/Core/PlatformManager.cs
```diff
@@ -209,28 +214,20 @@ public static Type GetConfigType(Platform platform)
...
-        /// <param name="platform">The platform to get the specific file extension for.</param>
-        /// <returns>File extension for the dynamic library used by the indicated platform.</returns>
-        public static string GetDynamicLibraryExtension(Platform platform)
+        public static bool TryGetConfigType(Platform platform, out Type configType)
         {
-            return PlatformInformation[platform].DynamicLibraryExtension;
-        }
+            configType = null;
 
+            bool typeFound = PlatformInformation.TryGetValue(platform, out PlatformInfo value);
```
WispyMouse, line 221: I feel like `PlatformInformation.TryGetValue` should return false if a null value is detected.

WispyMouse (changes requested):
It's a little difficult to parse how fully this has been implemented, because I can't make builds at the moment out of public. There's some things relating to `ProductConfig.Write` and `Config.WriteAsync` not existing outside of Editor time, but being called by Editor code. When we've separately addressed those issues, or included a solution into this PR, I can continue testing.

com.playeveryware.eos/Runtime/Core/PlatformManager.cs
```diff
@@ -99,21 +102,21 @@ public static Platform CurrentPlatform
...
-            AddPlatformInfo(Platform.PS4,         "PS4",           "eos_ps4_config.json"        );
-            AddPlatformInfo(Platform.PS5,         "PS5",           "eos_ps5_config.json"        );
-            AddPlatformInfo(Platform.Switch,      "Switch",        "eos_switch_config.json"     );
+            AddPlatformInfo(Platform.Android,     "Android",       "eos_android_config.json", typeof(AndroidConfig), "Android");
+            AddPlatformInfo(Platform.iOS,         "iOS",           "eos_ios_config.json",     typeof(IOSConfig),     "iPhone");
+            AddPlatformInfo(Platform.Linux,       "Linux",         "eos_linux_config.json",   typeof(LinuxConfig),   "Standalone");
+            AddPlatformInfo(Platform.macOS,       "macOS",         "eos_macos_config.json",   typeof(MacOSConfig),   "Standalone");
+            AddPlatformInfo(Platform.Steam,       "Steam",         "eos_steam_config.json",   null,         "Standalone");
+            AddPlatformInfo(Platform.XboxOne,     "Xbox One",      "eos_xb1_config.json",     null, "XboxOne");
+            AddPlatformInfo(Platform.XboxSeriesX, "Xbox Series X", "eos_xsx_config.json",     null, "XboxOne");
```
WispyMouse, line 111: Should the string here say "XboxOne" or something else? Perhaps there's a different phrase for it? Or is it the same one, because they all use similar GDK settings?

com.playeveryware.eos/Runtime/Core/Config/PlatformConfig.cs
```diff
@@ -26,17 +26,37 @@ namespace PlayEveryWare.EpicOnlineServices
     // referenced in the Epic namespace.
 #if !EOS_DISABLE
     using Epic.OnlineServices.IntegratedPlatform;
+    using Epic.OnlineServices.Auth;
+    using Epic.OnlineServices.Platform;
+    using Epic.OnlineServices.UI;
+    using UnityEditor.VersionControl;
```
WispyMouse: Why is version control necessary? Should this perhaps only be defined with UnityEditor?

com.playeveryware.eos/Runtime/Core/Config/PlatformConfig.cs
```diff
@@ -51,28 +71,430 @@ public abstract class PlatformConfig : Config
...
+        /// equal to zero as using the default value for the EOS SDK, which is
+        /// 30 seconds.
+        ///
+        /// This value is only used when the <see cref="NetworkStatus"/> is not
+        /// <see cref="NetworkStatus.Online"/>.
+        /// <seealso cref="PlatformInterface.GetNetworkStatus"/>
+        /// </summary>
+        [ConfigField("Network Timeout Seconds",
+            ConfigFieldType.Double,
+            "Indicates the maximum number of seconds that EOS SDK " +
```
WispyMouse: This is missing a really important piece of context. This timeout only applies *while EOS is not online yet*. It doesn't cover calls after becoming online.

https://dev.epicgames.com/docs/api-ref/structs/eos-platform-options

Assets/Plugins/Android/Core/AndroidConfig.cs
paulhazen: They should end in punctuation, they use standard grammar like everything else IMO.

Assets/Plugins/Source/Editor/EditorWindows/EOSSettingsWindow.cs
paulhazen: It does _feel_ weird, but the circumstance is a valid one. It's possible for a platform to be defined and the config type to not be defined - in which case it would return true indicating that the key `platform` exists in the dictionary, but that the value _at_ that key is `null`. 

This will be the case (for instance) for the key with the value `PlatformManager.Platform.Unknown`.

Assets/Plugins/Source/Editor/EditorWindows/NewConfigWindow.cs
paulhazen: Good catch - no it should be disabled. I've fixed that.

com.playeveryware.eos/Runtime/Core/Config/SandboxId.cs
paulhazen: Nope - a distinction needs to be made between a `null` value (which indicates that the user has not defined it yet) and an _invalid string_ which could be empty and would need to go through the validation checks that follow this conditional.

com.playeveryware.eos/Runtime/Core/PlatformManager.cs
paulhazen: No - `Unknown` is a distinct value. And it's appropriate for it to be 0 because by default the Platform should be `Unknown`.

com.playeveryware.eos/Runtime/Core/PlatformManager.cs
paulhazen: Again, in this circumstance there is an `Unknown` platform, for which `PlatformInfo` should be null - but we may want to define some characteristics of it within the `PlatformInfo` struct.

com.playeveryware.eos/Runtime/Core/PlatformManager.cs
paulhazen: Nope - these are the values that Unity uses to describe the icon within Unity that we can extract. Unity does not have a distinct logo for Xbox Series X vs Xbox One, so the same one is used.

com.playeveryware.eos/Runtime/Core/PlatformManager.cs
paulhazen: For the sake of clarity, I've changed `All` to be `Any`.

com.playeveryware.eos/Runtime/Core/Config/PlatformConfig.cs
paulhazen: Nope. I've removed it.

com.playeveryware.eos/Runtime/Core/Config/PlatformConfig.cs
paulhazen: Good note - I've changed it to make it much clearer.

arthur740212 (approved):

Assets/Plugins/Android/Core/AndroidConfig.cs
WispyMouse: It's the smallest of nits, but there's some inconsistent application of punctuation in these tooltips. Consider adjusting them. Resolve this conversation at your leisure either way, it's definitely just a nit.

Assets/Plugins/Source/Editor/EditorWindows/EOSSettingsWindow.cs
WispyMouse: That's definitely... strange. I'll take your word for it that this is how it needs to be, though.

com.playeveryware.eos/Runtime/Core/Config/SandboxId.cs
WispyMouse: Ah, that's worthwhile. I sort of feel like the correct approach might to be treating null and empty as entirely interchangeable, but I don't feel strongly about it. I'm leaving his comment unresolved so you can see it, but resolve it after you do.

com.playeveryware.eos/Runtime/Core/PlatformManager.cs
WispyMouse: Oh weird. Well, when in Unity, do as Unity does!

WispyMouse (approved):

PR #995, opened 2024-11-04
fix: Make changes to enable config-centric code to compile without Un…

paulhazen:
This PR adds some compile conditionals around components of the config system so that they can be utilized in a non Unity Engine context.

#EOS-2231


com.playeveryware.eos/Runtime/Core/Utility/FileSystemUtility.cs
```diff
@@ -516,12 +523,12 @@ public static async Task<string> ReadAllTextAsync(string path)
         }
 #endif
 
-        /// <summary>
-        /// Reads all text from the indicated file.
-        /// </summary>
-        /// <param name="path">Filepath to the file to read from.</param>
-        /// <returns>The contents of the file at the indicated path as a string.</returns>
-        public static string ReadAllText(string path)
+                /// <summary>
```
arthur740212: nit : formatting looks off

paulhazen: Good catch, I fixed it.

WispyMouse:
I turned on the EXTERNAL_TO_UNITY and tried to build. I don't also have EOS_DISABLE defined. I had a bunch of places not know how to reference stuff. Should I expect that EXTERNAL_TO_UNITY would work in this state, or is this starting to introduce it? Do I need that and also the EOS_DISABLE?

WispyMouse (approved):

arthur740212 (approved):

PR #997, opened 2024-11-05
fix: Sample Manager Menu Runs "ShowInternal" Only When It Should

WispyMouse:
When logging in to a sample, only run "Show" once when it is appropriate, and "Hide" once when it is appropriate. Before this change, "Show" was an overrideable function that child menus utilized. There was logic in all of them to run on various situations, some of which only should be run a single time. This introduces a function change to Show to make it no longer overrideable, instead instructing child implementers to override "ShowInternal".

The Metrics sample in particular was throwing an exception because Show was attempting to run logic before it had properly set itself up using this new paradigm. Now it has an "Initialize" variable that tracks if this should happen.

This has some risk in that it could be that some sample code was depending on this to run when it has. I've run each sample individually, and didn't find any other issues.

#EOS-2232

### DIFF
--- a/Assets/Scripts/StandardSamples/UI/Common/SampleMenu.cs
+++ b/Assets/Scripts/StandardSamples/UI/Common/SampleMenu.cs
@@ -219,7 +219,7 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
         /// <summary>
         /// Shows the SampleMenu.
         /// If this menu was not hidden, nothing happens.
-        /// If the menu was hidden, this calls <see cref="OnShow"/>.
+        /// If the menu was hidden, this calls <see cref="ShowInternal"/>.
         /// </summary>
         public void Show()
         {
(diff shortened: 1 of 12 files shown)


Assets/Scripts/StandardSamples/UI/Common/SampleMenu.cs
```diff
@@ -285,9 +302,31 @@ public virtual void Hide()
...
         }
 
+        /// <summary>
+        /// If this menu is shown through <see cref="Show"/> successfully, then
+        /// this method is run so that implementing child classes can respond
+        /// to becoming available.
+        /// </summary>
+        protected virtual void ShowInternal()
+        {
+
```
paulhazen: I think you should add a comment to the empty function that indicates that the default behavior is to take no action.

paulhazen (approved):

arthur740212 (approved):

PR #998, opened 2024-11-06
feat: Introduce native code console application

paulhazen:
In an effort to ease the development efforts within the native code component of the project, this PR introduces a ConsoleApplication in C++ with a reference to the `GfxPluginNativeRender` project, allowing developers to do debugging without requiring the interaction of Unity.

This PR also updates the version of Visual Studio that the projects in the NativeCode section are compatible with. This does not affect user requirements for the binaries, they remain unchanged.

#EOS-2231


lib/NativeCode/DynamicLibraryLoaderHelper/DynamicLibraryLoaderHelper.sln
```diff
@@ -31,6 +33,12 @@ Global
 		{251D4477-4C90-4FE1-94A2-52377327D3B2}.Release|x64.Build.0 = Release|x64
 		{251D4477-4C90-4FE1-94A2-52377327D3B2}.Release|x86.ActiveCfg = Release|Win32
 		{251D4477-4C90-4FE1-94A2-52377327D3B2}.Release|x86.Build.0 = Release|Win32
+		{D7462EAF-0024-4EA5-9D89-DBBBC7766333}.Debug|x64.ActiveCfg = Debug|x64
+		{D7462EAF-0024-4EA5-9D89-DBBBC7766333}.Debug|x64.Build.0 = Debug|x64
+		{D7462EAF-0024-4EA5-9D89-DBBBC7766333}.Debug|x86.ActiveCfg = Debug|Win32
+		{D7462EAF-0024-4EA5-9D89-DBBBC7766333}.Debug|x86.Build.0 = Debug|Win32
+		{D7462EAF-0024-4EA5-9D89-DBBBC7766333}.Release|x64.ActiveCfg = Release|x64
+		{D7462EAF-0024-4EA5-9D89-DBBBC7766333}.Release|x86.ActiveCfg = Release|Win32
```
WispyMouse, line 41: Because these .Debug lines have a .Build, should the Release also have a .Build associated with it? Or no, because you would never build this console application and deliver it to people?

paulhazen: You hit the nail on the head - we never build this console application, nor do we ever deliver it to people. It's for development purposes only.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/config.cpp
```diff
@@ -27,13 +27,177 @@
...
+EOSConfig eos_config_from_json_value(json_value_s* config_json)
+{
+    // Create platform instance
+    json_object_s* config_json_object = json_value_as_object(config_json);
+    json_object_element_s* iter = config_json_object->start;
+    EOSConfig eos_config;
+
+    while (iter != nullptr)
+    {
+        if (!strcmp("productName", iter->name->string))
```
WispyMouse, line 52: This surprises me a bit. "If productName cannot be compared to the current iterator, then do things." Is that what "!strcmp" is implying here? I would have expected it to do the thing it's in if it were *true*.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/dllmain.cpp
```diff
@@ -44,161 +47,126 @@ extern "C"
...
+                logging::log_inform(("Deployment ID override specified: " + deploymentArg).c_str());
                 eos_config.deploymentID = deploymentArg;
             }
         }
     }
+}
+
+// Called by unity on load. It kicks off the work to load the DLL for Overlay
+#if PLATFORM_32BITS
+#pragma comment(linker, "/export:UnityPluginLoad=_UnityPluginLoad@4")
```
WispyMouse, line 118: Is the @4 a magic number? Where does that come from? Is that an argument to the linker?

WispyMouse (commented):
This seems good, but I want to run this in my local before signing off. Just gonna make a build and make sure that it works.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/config.cpp
paulhazen: Yeah, me too - but if `strcmp` returns `0` if the two strings are equal, and in C++ `0` equates to `false`. I _also_ don't like how it looks 😆

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/dllmain.cpp
paulhazen: I honestly do not know - it's something that was there already (GitHub is interpreting it as an add because of how extensively the file was changed, but the line existed in the old implementation).

lib/NativeCode/DynamicLibraryLoaderHelper/DynamicLibraryLoaderHelper/DynamicLibraryLoaderHelper.vcxproj
```diff
@@ -23,78 +23,45 @@
...
+    <WindowsTargetPlatformVersion>10.0</WindowsTargetPlatformVersion>
   </PropertyGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'" Label="Configuration">
+  <!-- Consolidated Configuration Groups -->
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32' Or '$(Configuration)|$(Platform)'=='Debug|x64'" Label="Configuration">
     <ConfigurationType>StaticLibrary</ConfigurationType>
     <UseDebugLibraries>true</UseDebugLibraries>
-    <PlatformToolset>v141</PlatformToolset>
+    <PlatformToolset>v143</PlatformToolset>
```
arthur740212: Is 143 vs 2022?
I think our default is 2019 which is 142

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/NativeRender.vcxproj
```diff
@@ -23,81 +23,46 @@
...
     <ProjectName>GfxPluginNativeRender</ProjectName>
   </PropertyGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'" Label="Configuration">
+  <!-- Consolidated Configuration Groups -->
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32' Or '$(Configuration)|$(Platform)'=='Debug|x64'" Label="Configuration">
     <ConfigurationType>DynamicLibrary</ConfigurationType>
     <UseDebugLibraries>true</UseDebugLibraries>
-    <PlatformToolset>v141</PlatformToolset>
+    <PlatformToolset>v143</PlatformToolset>
```
paulhazen: We should keep this the same.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/NativeRender.vcxproj
```diff
@@ -23,81 +23,46 @@
     <ProjectGuid>{251D4477-4C90-4FE1-94A2-52377327D3B2}</ProjectGuid>
     <Keyword>Win32Proj</Keyword>
     <RootNamespace>NativeRender</RootNamespace>
-    <WindowsTargetPlatformVersion>10.0.17763.0</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformVersion>10.0</WindowsTargetPlatformVersion>
```
paulhazen: We should keep this the same.

arthur740212 (approved):

lib/NativeCode/DynamicLibraryLoaderHelper/DynamicLibraryLoaderHelper/DynamicLibraryLoaderHelper.vcxproj
paulhazen: I restored it to `v141`, which is actually Visual Studio 2017

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/config.cpp
```diff
@@ -27,13 +27,177 @@
...
+        }
+        else if (!strcmp("clientID", iter->name->string))
+        {
+            eos_config.clientID = json_value_as_string(iter->value)->string;
+        }
+        else if (!strcmp("clientSecret", iter->name->string))
+        {
+            eos_config.clientSecret = json_value_as_string(iter->value)->string;
+        }
+        if (!strcmp("encryptionKey", iter->name->string))
```
WispyMouse, line 104: This should probably be `else if`.

WispyMouse (approved):

PR #1004, opened 2024-11-12
House-keeping changes for upcoming release

paulhazen:
This PR introduces a variety of disconnected small changes so that the release PR can be more focused. Each change is documented here in the description.

Changes:
- Change `.gitignore` file to only ignore the `.csproj` files generated by Unity in the root of the repository. Ensure changes to any `.csproj` files in the `lib/NativeCode/` directory are captured.
- Sort the `using` statements, remove unused ones, and move them into the `namespace` scope within `PackageFileUtility.cs`.
- Remove `[Serializable]` attribute from `MacOSConfig` and `WindowsConfig` that are no longer needed.
- Add `GuidExtensions` class to more easily get a string representation of a `Guid` that is lowercase and without dashes, which is the format that the EOS SDK requires for many of its API calls.
- Add `VersionUtility` to compare versions for semantic equality.
- Add `Wrapped` class to facilitate enclosing a struct value in a reference.
- Switch `WrappedInitializeThreadAffinity` to make use of the `Wrapped` class.
- Add `VersionConverter` to the list of converters used by Newtonsoft.
- Use the `readonly` modifier on the relevant functions within the `SandboxId` struct and adds a `ToString` method.
- Add constructor to `Named` and alter `SetOfNamed` to better accommodate unnamed new values.
- Move encryption key validation into `EOSClientCredentials` class, rename `ConfigFieldType.NamedGuid` to `ConfigFieldType.Guid`, and added helper function to `GUIEditorUtility` to render section headers.
- Move Utility classes that are not unique to EOS into the common directory.

#EOS-2235


Assets/Plugins/Source/Editor/Utility/GUIEditorUtility.cs
```diff
@@ -395,15 +395,20 @@ static void HandleEnumField(
...
             object target,
             ConfigFieldAttribute fieldDetails,
             Func<object, object> getValue,
             Action<object, object> setValue,
             float labelWidth);
 
+        public static void RenderSectionHeader(string label)
+        {
+            GUILayout.Label(label.ToUpper(), EditorStyles.boldLabel);
+            Rect rect = EditorGUILayout.GetControlRect(false, 1);  // Set the height to 1 pixel
```
WispyMouse, line 419: What is the utility of this grey line? I'm guessing this creates an underline? I don't think I see it in the EOS Configuration panel.

image

Assets/Plugins/Source/Editor/Utility/GUIEditorUtility.cs
```diff
@@ -696,7 +701,7 @@ private static void RenderSetOfNamed<T>(
             Action<Rect, Named<T>> renderItemFn,
             Action addNewItemFn,
             Action<Named<T>> removeItemFn
-        ) where T : IEquatable<T>
+        ) where T : IEquatable<T>, new()
```
WispyMouse, line 715: Why does this have a "new()" constraint now?

The following is a question born of my unknowledge: Does this requiring "new()" make it limited to classes that can have parameterless constructors? Can you still pass in a struct as the type for T?

Assets/Plugins/Windows/Core/WindowsConfig.cs
```diff
@@ -25,7 +25,6 @@ namespace PlayEveryWare.EpicOnlineServices
     using System;
 
     // Flags specifically for Windows
-    [Serializable]
```
WispyMouse, line 28: You mentioned these aren't needed anymore. Does having the base class `PlatformConfig` utilizing `SerializableAttribute` remove the need to make child classes mark themselves as Serializable?

Assets/Scripts/StandardSamples/UI/Storage/Title/UITitleStorageMenu.cs
```diff
@@ -49,23 +49,6 @@ public class UITitleStorageMenu : SampleMenu
         public Text FileContent;
         private List<string> CurrentTags = new List<string>();
 
-        protected override void Awake()
```
WispyMouse, line 52: Why was this removed? Is it because Encryption Key moved elsewhere / can't validate it anymore? I understand if this should instead use the SelectableStateHandler thing, but I don't know why this was removed.

com.playeveryware.eos/Runtime/Core/Common/Named.cs
```diff
@@ -41,6 +41,20 @@ public class Named<T> : IEquatable<Named<T>>, IComparable<Named<T>>, IEquatable<
...
+        /// Create a new Named with the given name and value.
+        /// </summary>
+        /// <param name="name">The name for the value.</param>
+        /// <param name="value">The value.</param>
+        /// <exception cref="ArgumentNullException">
+        /// Thrown if either Name or Value is null.
+        /// </exception>
+        public Named(string name, T value)
+        {
+            Name = name ?? throw new ArgumentNullException(nameof(name));
```
WispyMouse: Is an empty name valid here? If not, could we use "IsNullOrEmpty"/WhiteSpace instead of a direct nullcheck?

WispyMouse: Given that these are set in a required constructor with no intended way to change things, should `Name` and `Value` be `readonly`?

com.playeveryware.eos/Runtime/Core/Common/Named.cs
```diff
@@ -55,7 +69,7 @@ public class Named<T> : IEquatable<Named<T>>, IComparable<Named<T>>, IEquatable<
         /// </returns>
         public static Named<T> FromValue(T value, string name)
```
WispyMouse, line 67: nit: It feels a bit funny to see the parameters for the constructor be "name then value", but here it's "value then name". Would it make sense to make these more consistent?

com.playeveryware.eos/Runtime/Core/Common/Named.cs
```diff
@@ -55,7 +69,7 @@ public class Named<T> : IEquatable<Named<T>>, IComparable<Named<T>>, IEquatable<
...
         {
-            return new Named<T>() { Name = name, Value = value };
+            return new Named<T>(name, value);
         }
 
         /// <summary>
```
WispyMouse, line 72: nit: This comment says "Compares one named value to another, for sorting purposes." I feel like it should mention in this comment blurb how it is *only comparing the name of the values*. You can infer that from the <result>, but perhaps this could be tweaked.

com.playeveryware.eos/Runtime/Core/Common/SetOfNamed.cs
```diff
@@ -66,7 +66,11 @@ public SetOfNamed(string defaultNamePattern = "NamedItem")
         /// </returns>
         public bool Add(T value)
         {
-            // Determines the 
+            // If value is null, then set it to either be default or a new 
+            // instance of the value type.
+            value ??= typeof(T).IsValueType ? default(T) : new T();
```
WispyMouse, line 71: Ah, I mentioned this in another comment, but this is where this feels weird.

SetOfNamed requires T to implement `new()`... but only use it if the object isn't a value type. Will this still work with structs that don't have a parameter-less constructor?

Thank you for the comment explaining what this does, I didn't recognize `??=`.

com.playeveryware.eos/Runtime/Core/Common/Utility/VersionUtility.cs
```diff
@@ -0,0 +1,60 @@
...
+        /// <param name="v1">The first version.</param>
+        /// <param name="v2">The second version.</param>
+        /// <returns>
+        /// True if, when ignoring any undefined components of both versions,
+        /// the versions are considered to be equal. The default equality
+        /// comparator for Version will return false if either version has a
+        /// component that is undefined. This utility function circumvents that
+        /// limitation.
+        /// </returns>
+        public static bool AreVersionsEqual(Version v1, Version v2)
```
WispyMouse, line 41: Does `System.Version.CompareTo` do what we need here? Or does "The default equality comparator for Version will return false if either version has a component that is undefined" imply that it doesn't work for our needs?

com.playeveryware.eos/Runtime/Core/Common/Utility/VersionUtility.cs
```diff
@@ -0,0 +1,60 @@
...
+        {
+            // Compare major versions
+            if (v1.Major != v2.Major) return false;
+
+            // Compare minor versions
+            if (v1.Minor != v2.Minor) return false;
+
+            // Compare build versions if both are specified
+            if (v1.Build != -1 && v2.Build != -1 && v1.Build != v2.Build) return false;
```
WispyMouse: Let's say I have "1.1.1" (`Version A`) and "1.1" (`Version B`). `Version B` does not specify a Build or Revision. This logic will notice that `Version B` does not have those specified, and will find itself to be equivalent to `Version A`. Is that desirable?

com.playeveryware.eos/Runtime/Core/Config/ProductConfig.cs
```diff
@@ -104,24 +100,55 @@ static ProductConfig()
...
+        }
+
+        private void MigrateProductNameVersionAndId(PreviousEOSConfig config)
         {
-            ProductId ??= new Named<Guid>();
-            ProductId.Name = config.productName;
-            if (!Guid.TryParse(config.productID, out ProductId.Value))
+            ProductName = config.productName;
+
+            if (!Version.TryParse(config.productVersion, out Version))
```
WispyMouse: nit: this reads a bit funny to me. It makes sense in the editor, but it *looks* like you're saying "Take this Version object, try to parse it using this productVersion string, and set the output to Version", but it's really "Version has a static function TryParse, and put the result in to Version".

com.playeveryware.eos/Runtime/Core/Config/SandboxId.cs
```diff
@@ -66,7 +66,7 @@ public readonly bool Equals(SandboxId other)
             return _value == other._value;
         }
 
-        public override bool Equals(object obj)
+        public override readonly bool Equals(object obj)
```
WispyMouse, line 78: Functions can be readonly? o_O

What... what does that *do*? It promises that the function won't change state? That's kind of amazing, I'll need to use this all over the place in my own projects. Good stuff.

arthur740212 (approved):

com.playeveryware.eos/Runtime/Core/Config/SandboxId.cs
```diff
@@ -75,5 +75,14 @@ public override readonly int GetHashCode()
...
             return (_value?.GetHashCode()) ?? 0;
         }
+
+        public override readonly string ToString()
+        {
+            // SandboxId is _sometimes_, but _not always_ a GUID value. In case
+            // it was entered into the configuration with a format that includes
+            // dashes and upper case letters, this function removes the former
+            // and changes the latter.
+            return _value.Replace("-", "").ToLower();
```
WispyMouse: How do you feel about setting _value to this (Replacing - and setting to Lower) when the value is set, rather than doing that work here?

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -490,7 +490,7 @@ private PlatformInterface CreatePlatformInterface()
 #else
                     configData.platformOptionsFlags.Unwrap();
 #endif
-                if (configData.IsEncryptionKeyValid())
+                if (EOSClientCredentials.IsEncryptionKeyValid(configData.encryptionKey))
```
WispyMouse, line 491: This seems really close to the `IsEncryptionKeyValid` function. Can we just use that?

com.playeveryware.eos/Runtime/Core/EOS_SDK_Additions/EOSClientCredentials.cs
```diff
@@ -29,13 +29,21 @@ namespace PlayEveryWare.EpicOnlineServices
...
     {
         public string ClientId;
         public string ClientSecret;
         public readonly string EncryptionKey;
 
+        private static readonly Regex s_invalidEncryptionKeyRegex;
+
+        static EOSClientCredentials()
+        {
+            s_invalidEncryptionKeyRegex = new Regex("[^0-9a-fA-F]");
```
WispyMouse, line 44: I didn't know the encryption key had to be a GUID. Is this a new limitation, or something that was always secretly present?

WispyMouse: > This 64-character hexadecimal string builds a 256-bit key that EOS uses to encrypt user data.
https://dev.epicgames.com/docs/game-services/title-storage

This is entirely correct, it should be a hex of this length. I wonder if there's anyone out there who is in violation of this rule that'll be hosed by the update, or if was never valid to provide a key that didn't match these criteria.

WispyMouse: Do you happen to know if this field should be case sensitive?

WispyMouse (approved):
Lots of thoughts a little bits. Nothing that I feel like should stop you from merging this in.

Defining EOS_DISABLE has a lot of misses at the moment. I don't know how correct it's meant to be from this PR, but we definitely need to clean this up before the release.

Assets/Plugins/Source/Editor/Utility/GUIEditorUtility.cs
paulhazen: Yep! requiring `new()` means that it must have a parameter-less constructor. Because by definition `struct` objects satisfy that requirement, it works for them as well.

Assets/Plugins/Windows/Core/WindowsConfig.cs
paulhazen: It does not - does the attribute still exist on the base class? The attribute is used mostly for Unity's `JsonUtility`, but since we have supplanted that with Newtonsoft, it's no longer needed, and keeping them in could lead to some confusion in maintaining the project over time.

Assets/Scripts/StandardSamples/UI/Storage/Title/UITitleStorageMenu.cs
paulhazen: This was removed because with the new config system it is not possible to get into a situation where the encryption key is invalid - unless the user is trying _really_, _really_ hard. You basically need to edit the json config file manually. And when the encryption key is invalid, it will log an error when it comes time to use it instead of checking pre-emptively. I felt like that would be a more appropriate place to alert the user since it will be closely attached to its usage.

In the old system, there is a button to "generate" an encryption key, in the new system it's just created for you and there is not a manual way to alter it. Still works with backwards compatibility because users' old configuration files will be migrated, so the old encryption key they may have manually defined will still be present.

The details of that change are a little out of scope for _this_ PR, but the simplest change was to make sure that the encryption key invalidity is mentioned right when it's created, and then if it's still bad only when it's used.

com.playeveryware.eos/Runtime/Core/Common/Named.cs
paulhazen: There is a way to change things, the field's `Name` and `Value` are both marked `public`.

com.playeveryware.eos/Runtime/Core/Config/ProductConfig.cs
paulhazen: Ah, I do see how that's a bit icky. I think the best solution would probably be to change the `Version` field member to be named `ProductVersion`. Thoughts?

com.playeveryware.eos/Runtime/Core/Common/Named.cs
paulhazen: My thought process was that grammatically "value" is the subject, so it should come first.

com.playeveryware.eos/Runtime/Core/Common/Utility/VersionUtility.cs
paulhazen: Your inference is correct - it doesn't work for our needs.

com.playeveryware.eos/Runtime/Core/Config/SandboxId.cs
paulhazen: I know, right?!? I was surprised too. The modifier `readonly` can be applied only to functions within `struct` objects and is used by the compiler to provide guarantees that calling the function does not alter the value of the struct itself. I _think_ what it basically does behind the scenes is it makes the function a static const expression - something it knows can be done safely if the `readonly` modifier is employed.

com.playeveryware.eos/Runtime/Core/Config/SandboxId.cs
paulhazen: The reason I don't do that _here_ is because it would be very confusing to have the `ToString()` method actually alter the value. 

I _can_ however see the value in perhaps using it whenever `_value` _is_ being set (like in the `set` section of the Property).

com.playeveryware.eos/Runtime/Core/EOS_SDK_Additions/EOSClientCredentials.cs
paulhazen: I believe that the field is case sensitive, in that an encryption key with different case cannot be used to decrypt a file encrypted with the original case.

com.playeveryware.eos/Runtime/Core/EOSManager.cs
paulhazen: It is very close, however currently `EOSConfig` does not have an `EOSClientCredentials` field member, so you cannot acquire it and call the function from the instance. This will change in the upcoming changes, but moving the implementation of checking the validity of the encryption key into the `EOSClientCredentials` class first seemed like a reasonable approach to make the changes more incremental.

com.playeveryware.eos/Runtime/Core/Config/ProductConfig.cs
paulhazen: I went ahead and changed the field member to `ProductVersion`.

Assets/Plugins/Source/Editor/Utility/GUIEditorUtility.cs
paulhazen: It's used in an upcoming change - at least it's more visible within it. It is not a surprise that there is no visible distinction for the time being.

Assets/Plugins/Windows/Core/WindowsConfig.cs
paulhazen: It in fact did - so I also removed it from `PlatformConfig`, and `Config`, and a handful of others. Good catch.

com.playeveryware.eos/Runtime/Core/Common/SetOfNamed.cs
paulhazen: Parameterless constructors are automatically added to every struct definition, they will always have one. However, for the purposes of more explicit syntax and greater clarity, I wrote it this way.

com.playeveryware.eos/Runtime/Core/Common/Utility/VersionUtility.cs
paulhazen: Hmmmm, good point. Lemme marinate on that for a little bit.

paulhazen: Nice catch! I updated the implementation so that if one version has the component defined and the other one does not, they are not considered equal.

The built-in comparator doesn't work for us because it (rather appropriately) does not consider versions to be equal if they both have matching undefined components, because if it's undefined you cannot say with absolute confidence that the versions are equal. But that's a little strict for our purposes.

com.playeveryware.eos/Runtime/Core/Common/Named.cs
paulhazen: Good point, I updated the comment to be clearer.

com.playeveryware.eos/Runtime/Core/Config/SandboxId.cs
paulhazen: Actually, I realize that this implementation is incorrect - because valid sandbox ids can start with a `p-`, and this function would mess that up. I've updated the implementation to consider it more properly.

com.playeveryware.eos/Runtime/Core/Common/Named.cs
WispyMouse: Should there be? Structs are weird and all, but it feels strange that I can create a `Named`, it enforces that it has certain qualities (isn't null), but then I can set those properties directly to null.

paulhazen: Good point - I removed the exception throwing component - if we need to re-introduce it later, we can consider a better approach. Doesn't seem like it was important for these narrower changes.

PR #1007, opened 2024-11-13
fix: Allow `ProductVersion` to be a string.

paulhazen:
This PR allows the `ProductVersion` field member of the `ProductConfig` class to be any string, not just a version.

#EOS-2235


arthur740212 (approved):

WispyMouse (approved):
As noted in Slack;

The ProductVersion must be a Version. However, in the current release, it is a string. While it is moving from Version to String here, the Version-fix never reached public. In order to unblock further work, this has been approved. An item has been created in the backlog to address this.

PR #1011, opened 2024-11-15
feat: New Config Window

paulhazen:
This PR "flips the switch" on enabling the new `EOSSettingsWindow` (pictured at the end of this PR description).

This is done in preparation for the upcoming release - to make sure that the configuration classes are properly presented to the user and properly utilized in the various places where configuration values need to be determined.

image

#EOS-2235


Assets/StreamingAssets/EOS/eos_android_config.json
```diff
@@ -26,6 +31,6 @@
   "alwaysSendInputToOverlay": false,
   "initialButtonDelayForOverlay": 0.0,
   "repeatButtonDelayForOverlay": 0.0,
-  "toggleFriendsButtonCombination": "SpecialLeft",
-  "_configValuesMigrated": false
+  "toggleFriendsButtonCombination": "None",
```
WispyMouse, line 34: Why is this now "None" instead of the default "SpecialLeft"?

WispyMouse: I see now, this is specific to Android which doesn't support the overlay.

Assets/StreamingAssets/EOS/eos_android_config.json
```diff
@@ -3,13 +3,18 @@
...
-  "clientCredentials": null,
+  "clientCredentials": {
+    "ClientId": "xyza7891wPzGRvRf4SkjlIF8YuqlRLbQ",
+    "ClientSecret": "aXPlP1xDH0PXnp5U+i+M5pYHhaE1a8viV0l1GO422ms",
+    "EncryptionKey": "1111111111111111111111111111111111111111111111111111111111111111"
+  },
+  "isServer": false,
   "platformOptionsFlags": "None",
-  "authScopeOptionsFlags": "NoFlags",
+  "authScopeOptionsFlags": "BasicProfile, FriendsList, Presence",
```
paulhazen, line 17: We need to make sure to broadcast this and make it clear in the documentation that this might be a change.

com.playeveryware.eos/Runtime/Core/Common/Named.cs
```diff
@@ -52,21 +54,24 @@ public Named(string name, T value)
...
+            string oldName = Name;
+            Name = newName;
+            if (notify)
+            {
+                NameChanged?.Invoke(this, new ValueChangedEventArgs<string>(oldName, newName));
+            }
+
+            // If the name has not been set to the new name then it was 
+            // not able to be set.
+            return !string.Equals(Name, oldName);
```
WispyMouse: nit: If this is meant to say what the comment says, could we instead say `return string.Equals(Name, newName)`?

com.playeveryware.eos/Runtime/Core/Common/SetOfNamed.cs
```diff
@@ -55,6 +64,28 @@ public SetOfNamed(string defaultNamePattern = "NamedItem")
...
+        }
+
+        /// <summary>
+        /// Returns the name of a new item to be added to the collection.
+        /// </summary>
+        /// <returns>The name to give to the added set.</returns>
+        private string GetNewItemName()
+        {
+            string newItemName = _defaultNamePattern;
+            int increment = 0;
```
WispyMouse: I feel strange about this code; doing string interpolation and comparison continually sounds more expensive than needed. Could we just keep `increment` around and keep adding to that?

com.playeveryware.eos/Runtime/Core/Common/SetOfNamed.cs
```diff
@@ -103,31 +148,7 @@ private string GetNewItemName()
...
-        public bool Add(string name, T value)
-        {
-            // Add to the collection using the base implementation, so long as
-            // the name for the new item is not used for another named item 
-            // already in the collection, and so long as the value is not the
-            // same as the value of another item in the collection.
-            return !ContainsName(name) &&
-                   !Contains(value) &&
-                   base.Add(Named<T>.FromValue(value, name));
+            return Add(default(T));
```
WispyMouse: Should this use logic similar to the above with `value ??= typeof(T).IsValueType ? default : new T();`?

com.playeveryware.eos/Runtime/Core/Common/SetOfNamed.cs
```diff
@@ -184,6 +205,24 @@ public bool Contains(T item)
...
+            if (_removePredicate != null && !_removePredicate(item.Value))
+            {
+                return false;
+            }
+            
+            // Remove the event subscription
+            item.NameChanged -= OnItemNameChanged;
+
+            // Otherwise attempt to remove the item
+            base.Remove(item);
```
WispyMouse, line 234: Should this return the result of this function? `base.Remove` makes me think that it has the same return type as this function, which is `bool`. So should that pass through in the result?

com.playeveryware.eos/Runtime/Core/Common/SetOfNamed.cs
```diff
@@ -184,6 +205,24 @@ public bool Contains(T item)
 
             return false;
         }
+
+        public new bool Remove(Named<T> item)
```
WispyMouse, line 221: Why did you opt to use `new` here instead of `override`? I'm guessing that this is from inheriting from List.

com.playeveryware.eos/Runtime/Core/Config/Config.cs
```diff
@@ -129,6 +149,32 @@ protected Config(
...
+                return false;
+            }
+
+            Debug.LogWarning(
+                $"Config file with schemaVersion \"{CURRENT_SCHEMA_VERSION}\"" +
+                " has been read into memory, and needs to be migrated to " +
+                $"schemaVersion \"{CURRENT_SCHEMA_VERSION}\".");
+
+            return true;
+        }
```
paulhazen, line 219: This could be easily Unit-tested! :)

com.playeveryware.eos/Runtime/Core/Config/EOSConfig.cs
```diff
@@ -75,6 +75,13 @@ static EOSConfig()
         protected EOSConfig() : base("EpicOnlineServicesConfig.json")
         { }
 
+        protected override bool NeedsMigration()
+        {
+            // Because EOSConfig is becoming obsolete, it should never
+            // be "migrated", but left-as-is.
+            return false;
+        }
```
paulhazen: A more verbose description of why this is the case would be a welcome addition to this function.

com.playeveryware.eos/Runtime/Core/Config/ProductionEnvironments.cs
```diff
@@ -72,19 +61,17 @@ public bool AddNewSandbox()
...
         /// </returns>
-        public bool RemoveSandbox(Named<SandboxId> sandbox)
+        private bool CanSandboxBeRemoved(SandboxId sandbox)
         {
             foreach (Named<Deployment> deployment in Deployments)
             {
-                if (deployment.Value.SandboxId.Equals(sandbox.Value))
+                if (deployment.Value.SandboxId.Equals(sandbox))
                 {
-                    // TODO: Tell user that the sandbox cannot be removed,
```
WispyMouse, line 81: Was this TODO accomplished?

com.playeveryware.eos/Runtime/Core/Common/SetOfNamed.cs
paulhazen: Discussed this in meeting - however, adding some more comments to describe how and why this works would be good.

com.playeveryware.eos/Runtime/Core/Common/SetOfNamed.cs
paulhazen: Probably good to relocate this into an "Editor" directory to make it clear that something so non-performant is only going to execute in the editor.

paulhazen: Instead of taking that approach, I changed the implementation to use a binary search algorithm to determine the increment to use for the new name. The maximum increment is equal to the number of items in the collection already, and the minimum increment is one. The new implementation performs a binary search between the two values and continues until the smallest increment is determined.

This approach was favored over moving the collection to an editor-only context because the code that performs migration makes use of SetOfNamed, and the migration code is not guaranteed to only run in the editor (by design).

com.playeveryware.eos/Runtime/Core/Common/Named.cs
paulhazen: Changed per suggestion. Marking as resolved.

com.playeveryware.eos/Runtime/Core/Config/EOSConfig.cs
paulhazen: Done. Marking conversation as resolved.

Assets/StreamingAssets/EOS/eos_android_config.json
paulhazen: I can confirm that current implementation (in `stable`) does not consider the `overrideValues` for auth scope options flags, so there is no need to highlight this change. See the below screenshot, which is the result of searching for the field name in question across the entire solution in the `stable` branch:

image

As you can see above, the values are never overridden by the `overrideValues` field member's contents.

arthur740212 (approved):

andrew-hirata-playeveryware (approved):

WispyMouse (changes requested):
Currently cannot build Windows standalone in this branch

com.playeveryware.eos/Runtime/Core/Config/Config.cs
```diff
@@ -239,8 +285,11 @@ public static async Task<T> GetAsync<T>() where T : Config
             s_cachedConfigs.Add(typeof(T), instance);
 #endif
 
-            // Call prepare function.
-            instance.MigrateConfig();
+            if (instance.NeedsMigration())
+            {
+                instance.MigrateConfig();
+                await instance.WriteAsync();
```
WispyMouse: `WriteAsync` and `Write` are both defined to only be included in the editor. Trying to compile a standalone build, or selecting "Player" as the project, will show that this isn't included, and fail to build.

paulhazen: Nice catch, I've fixed this now with compile-time conditionals.

WispyMouse (approved):

PR #1014, opened 2024-11-20
fix: Correct issue where Netwonsoft was incorrectly attempting to ser…

paulhazen:
Correct issue where Netwonsoft was incorrectly attempting to serialize a property that doesn't need to be.


arthur740212 (approved):

WispyMouse (approved):
This is the first bit of code in this assembly (com.playeveryware.eos-Editor) that references Newtonsoft.JSON. It doesn't have an explicit reference to the assembly, depending on the fact that the Packages folder's Newtonsoft.Json is marked as "auto referenced". I think this is safe; we have a dependency on this package, this package will always mark itself as auto referenced, but would it make sense to explicitly include a reference in the asmdef to Newtonsoft.Json?

PR #1015, opened 2024-11-20
fix: Correct Guid Serialization

paulhazen:
Because the eos sdk needs to have any Guid value be input as a string with no dashes, this PR introduces a custom GuidConverter that performs that task automatically.

This prevents the native code from needing to do the manipulation.

The associated JSON files were updated as well.

Since the way the converter works seems like a more straightforward and non-intrusive means of accomplishing what GuidExtensions used to do, that was also removed in favor of using the `.ToString("N").ToLowerInvariant()` method.


com.playeveryware.eos/Runtime/Core/Config/JsonConverter/GuidConverter.cs
```diff
@@ -0,0 +1,67 @@
...
+            var stringValue = reader.Value?.ToString();
+
+            if (string.IsNullOrEmpty(stringValue))
+            {
+                return objectType == typeof(Guid?) ? (Guid?)null : Guid.Empty;
+            }
+
+            if (!Guid.TryParseExact(stringValue, "N", out Guid value))
+            {
+                value = Guid.Parse(stringValue);
```
paulhazen: Might be a good idea to use `TryParse`?

paulhazen: This has been resolved - the behavior is to use TryParse, and if it fails to set the returning value to be `Guid.Empty`. This behavior is consistent with other classes that implement the `JsonConverter` class.

com.playeveryware.eos/Runtime/Core/Config/JsonConverter/GuidConverter.cs
```diff
@@ -0,0 +1,70 @@
...
+*/
+
+namespace PlayEveryWare.EpicOnlineServices
+{
+    using Newtonsoft.Json;
+    using System;
+
+    public class GuidConverter : JsonConverter
+    {
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
```
WispyMouse, line 30: I noticed that `JsonConverter` uses `object?` for the parameter `value`, `existingValue`, and the return type for `ReadJson`. Should we match that, or does the `?` not work with our version of C#?

com.playeveryware.eos/Runtime/Core/Config/JsonConverter/GuidConverter.cs
```diff
@@ -0,0 +1,70 @@
...
+        {
+            if (value is Guid guid)
```
WispyMouse, line 32: Will this work if `value` is `Guid?`, or only if the value type is exactly `Guid`?

paulhazen: It's not possible for value to be `Guid?`

com.playeveryware.eos/Runtime/Core/Config/JsonConverter/GuidConverter.cs
paulhazen: `?` works - but this is only applied to field members that are of type `Guid`, not any that are of type `Guid?`

WispyMouse (approved):
Left some thoughts, but otherwise approved.

com.playeveryware.eos/Runtime/Core/Config/JsonConverter/GuidConverter.cs
andrew-hirata-playeveryware: Perhaps I don't understand this JSON writer well enough, but why is possible to Read JSON where the type is a `Guid?` but not write one?

paulhazen: It is possible - it's just that we don't have a need to implement it for `Guid?`

We _could_ add it for future-proofing - but that would be adding functionality that isn't currently needed, which seemed like a bad thing to do. Support for `Guid?` would be easy to add though - if you'd like it added.

andrew-hirata-playeveryware (approved):

PR #1016, opened 2024-11-20
fix: Correct IPMF Enum serialization

paulhazen:
This PR corrects an error in the JSON serialization and deserialization. In previous versions, a `0` was used to indicate that the `integratedPlatformManagementFlags` property was not set. 

This conflicts with current means of deserialization in both the managed and native components of the code.

This PR mantains backwards compatibility with previous formats but writes the values to json using the corrected property name, while still reading in the old name.

To facilitate this behavior, the `ListOfStringsToEnumConverter` was changed such that if the underlying value parsed is determined to be less than the lowest value in the enum, the value returned is the lowest value.

The associated JSON files are updated to reflect this behavior.


com.playeveryware.eos/Runtime/Core/Config/JsonConverter/ListOfStringsToEnumConverters.cs
```diff
@@ -199,7 +179,30 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
...
+
+            // Get the lowest value in the enum
+            int lowest = Int32.MaxValue;
+            foreach(int enumIntValue in Enum.GetValues(typeof(T)))
+            {
+                if (lowest > enumIntValue)
+                {
+                    lowest = enumIntValue;
+                }
+            }
```
paulhazen: This should be replace with a single call to the `GetLowest()` function in the `EnumUtility` class.

paulhazen: Fixed to utilize `EnumUtility`'s `GetLowest` function.

andrew-hirata-playeveryware (approved):

arthur740212 (approved):

PR #1017, opened 2024-11-20
fix: Correct the `WrappedInitializeThreadAffinity` class to track the `ApiVersion` property

paulhazen:
This PR introduces storing the `ApiVersion` property in both the JSON and the data structure. This value is utilized in native code when performing initialization.


Assets/StreamingAssets/EOS/eos_android_config.json
```diff
@@ -26,7 +26,8 @@
     "HttpRequestIo": 0,
     "RTCIo": 0,
     "EmbeddedOverlayMainThread": 0,
-    "EmbeddedOverlayWorkerThreads": 0
+    "EmbeddedOverlayWorkerThreads": 0,
+    "ApiVersion": 3
```
andrew-hirata-playeveryware, line 30: This is the API field from the SDK? If so, it should _not_ be included like this.

paulhazen: `InitializeThreadAffinity` has _it's own_ `ApiVersion` attribute. Including it here allows us to detect instances where the object defined in the EOS SDK differs from the version for which there are stored configuration values.

A notable example of this is recently both fields `EmbeddedOverlayMainThread` and `EmbeddedOverlayWorkerThreads` were added to the object, and the `ApiVersion` was bumped from `2` to `3`.

paulhazen:
Closing for the time being - as a larger discussion is required to think carefully about the implications of this change.

PR #1019, opened 2024-11-20
chore: Organize `.cpp` files and `.h` files into appropriate subdirectories

paulhazen:
This PR organizes the source files and header files within the NativeRender (GfxPluginNativeRender) project into corresponding `src` and `headers` directories.

This was a preliminary step taken before implementing changes to the native code that enable the new config system. The changes are put here to clearly delineate them from the rest of the work.


arthur740212 (approved):

lib/NativeCode/DynamicLibraryLoaderHelper/ConsoleApplication/ConsoleApplication.cpp
```diff
@@ -21,7 +21,8 @@
  */
 
 #include <iostream>
-#include <config.h>
+
+#include "headers/config.h"
```
andrew-hirata-playeveryware: I think that instead of 'headers' the path should be 'include', which will match the rest of the project.

andrew-hirata-playeveryware (changes requested):

lib/NativeCode/DynamicLibraryLoaderHelper/ConsoleApplication/ConsoleApplication.cpp
paulhazen: My thought process was that `include` tends to imply external dependencies. In other places where we specify a directory from which to "include" from, they are always instances where what is being included is not something we own directly (such as `.h` files provided by the EOS SDK). Therefore, it seemed appropriate to make a distinction by using `headers`. 

I could go either way on this though - just wanted to mention it - do you still think we should use `include`?

andrew-hirata-playeveryware: I can see how that could be inferred, and it is a reasonable assumption, but in general 'include' is just a place to place headers for C and C++ in a Unix like system. We currently have that same setup for the native code for the [DynamicLibraryLoader ](https://github.com/PlayEveryWare/eos_plugin_for_unity/tree/stable/lib/NativeCode/include)code, though it isn't 100% setup the way I would like to be setup.

paulhazen: Done - changed `headers` to `include`, and updated corresponding references.

andrew-hirata-playeveryware (approved):

PR #1025, opened 2024-11-26
chore: Prep Native Projects for Expansion

paulhazen:
This PR introduces a few changes to the native code projects that will help facilitate expanding functionality to encompass the new config system.

The changes and the reasons behind them are outlined below:

## Rename `config` to `config_legacy`

The files `config.h` and `config.cpp` have been renamed `config_legacy.h` and `config_legacy.cpp` respectively, also updating the `#include` statements that reference either file. This was done because the config system to be introduced makes use of a `config.h` file in a different directory and having two `.h` files with the same name in the same directory structure can cause problems. This also helps clearly delineate between legacy functionality and soon-to-be introduced new functionality.

## Clearer PostBuild event commands

Prior to this PR, the postbuild event commands for both `DynamicLibraryLoaderHelper` and `GfxNativePluginRender` copied all files from the output directory into the corresponding `Assets/Windows/[x86|x64]/` directories. This PR introduces a change that guarantees only the `.dll` and `.dll.meta` files are copied over.

## Unify `dllimport` and `dllexport` helper macros

Prior to this PR - there was inconsistent usage of the macros for exporting and importing functions to/from the resulting dll. This PR unifies the approach, puts the macro definitions into their own file, and adds very verbose documentation explaining exactly how they function. In addition to these, the scripting preprocessor definition `PEW_EOS_EXPORT` was introduced to make these work consistently between projects.

## `NormalizedPlatform`

Different projects have different "Platform" specifications. This property was added to the `Directory.Build.props` file that resides in the solution directory. It helps normalize paths for projects that prefer the platform specification `Win32`. This is not new behavior - it is just codified here. In previous iterations, under the Win32 section, file paths would use string literals to specify that the file should be copied to `x86` instead of `Win32`. Instead of having those spread around the solution's projects in various places, this allows us to just reference one value and know that it will be consistently applied across the projects in the solution.

## Added logging overloads

To help facilitate easier logging (removing the need to work with combining different `const char*` values into a single message), overload functions were added for `log_warn`, `log_info`, and `log_error` inside `logging.h` that accept a standard `const std::string& log_string` parameter. This allows for easy string concatenation inside log functions when needed.

#EOS-1982

### DIFF
--- a/lib/NativeCode/DynamicLibraryLoaderHelper/DynamicLibraryLoaderHelper/DynamicLibraryLoaderHelper.vcxproj
+++ b/lib/NativeCode/DynamicLibraryLoaderHelper/DynamicLibraryLoaderHelper/DynamicLibraryLoaderHelper.vcxproj
@@ -93,9 +93,9 @@
     </Link>
     <PostBuildEvent>
       <!-- Copy the meta file stored in the solution directory, since it's values have been properly set in such a way that Unity interacts with the DLL in the intended manner. -->
-      <Command>copy $(SolutionDir)$(ProjectName)-$(Platform).dll.meta $(OutputUnityAssetsDirectory) /Y</Command>
+      <Command>copy $(SolutionDir)$(ProjectName)-$(NormalizedPlatform).dll.meta $(OutputUnityAssetsDirectory) /Y</Command>
       <!-- Copy just the output .dll file into the appropriate assets directory alongside the aforementioned .dll.meta file.-->
-      <Command>copy "$(TargetDir)$(ProjectName)-$(Platform).dll" "$(OutputUnityAssetsDirectory)" /Y</Command>
+      <Command>copy "$(TargetDir)$(ProjectName)-$(NormalizedPlatform).dll" "$(OutputUnityAssetsDirectory)" /Y</Command>
     </PostBuildEvent>
   </ItemDefinitionGroup>
   <!-- Source Files -->
--- a/lib/NativeCode/DynamicLibraryLoaderHelper/DynamicLibraryLoaderHelper/DynamicLibraryLoaderHelper.vcxproj
+++ b/lib/NativeCode/DynamicLibraryLoaderHelper/DynamicLibraryLoaderHelper/DynamicLibraryLoaderHelper.vcxproj
@@ -92,8 +92,10 @@
       <OptimizeReferences>true</OptimizeReferences>
     </Link>
     <PostBuildEvent>
-      <Command>copy $(SolutionDir)$(ProjectName)-$(Platform).dll.meta $(OutputUnityAssetsDirectory)</Command>
-      <Command>xcopy "$(TargetDir)*" "$(OutputUnityAssetsDirectory)" /E /I /Y</Command>
+      <!-- Copy the meta file stored in the solution directory, since it's values have been properly set in such a way that Unity interacts with the DLL in the intended manner. -->
+      <Command>copy $(SolutionDir)$(ProjectName)-$(Platform).dll.meta $(OutputUnityAssetsDirectory) /Y</Command>
+      <!-- Copy just the output .dll file into the appropriate assets directory alongside the aforementioned .dll.meta file.-->
+      <Command>copy "$(TargetDir)$(ProjectName)-$(Platform).dll" "$(OutputUnityAssetsDirectory)" /Y</Command>
     </PostBuildEvent>
   </ItemDefinitionGroup>
   <!-- Source Files -->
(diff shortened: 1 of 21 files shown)


WispyMouse (approved):

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/dllmain.cpp
```diff
@@ -41,13 +40,19 @@ using namespace pew::eos::eos_library_helpers;
...
+ * \brief Forward declaration for function to be called when Unity is loading
+ * the plugin.
+ */
+PEW_EOS_API_FUNC(void) UnityPluginLoad(void* unityInterfaces);
+
+/**
+ * \brief Forward declaration for function to be called when Unity is unloading
+ * the plugin.
+ */
+PEW_EOS_API_FUNC(void) __stdcall UnityPluginUnload();
```
paulhazen: These should actually always be export.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/dllmain.cpp
```diff
@@ -117,43 +122,47 @@ void get_cli_arguments(config::EOSConfig eos_config)
...
+            // Load EOSConfig
+            config_legacy::EOSConfig eos_config;
+            if (!config_legacy::try_get_eos_config(eos_config))
+            {
+                logging::log_error("Could not load EOSConfig.");
+            }
+            else
+            {
+                logging::log_inform("Loaded EOSConfig.");
+            }
```
paulhazen: This should be moved to its original location.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/dllmain.cpp
paulhazen: Fixed.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/dllmain.cpp
paulhazen: Done.

lib/NativeCode/DynamicLibraryLoaderHelper/DynamicLibraryLoaderHelper/DynamicLibraryLoaderHelper.vcxproj
```diff
@@ -92,8 +92,10 @@
       <OptimizeReferences>true</OptimizeReferences>
     </Link>
     <PostBuildEvent>
-      <Command>copy $(SolutionDir)$(ProjectName)-$(Platform).dll.meta $(OutputUnityAssetsDirectory)</Command>
-      <Command>xcopy "$(TargetDir)*" "$(OutputUnityAssetsDirectory)" /E /I /Y</Command>
```
paulhazen, line 96: For context - the reason this is changed is because there are going to be additional projects added to the solution file, so copying _all_ binaries from the target directory is going to bring in things that should only be there in test scenarios.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/NativeRender.vcxproj
```diff
@@ -78,44 +78,47 @@
   <ItemDefinitionGroup Condition="'$(Configuration)'=='Debug'">
     <ClCompile>
       <Optimization>Disabled</Optimization>
-      <PreprocessorDefinitions>OVERLAY_DLL_NAME="$(OverlayDllName)";STEAM_API_DLL="$(SteamApiDll)";SDK_DLL_NAME="$(SdkDllName)";CONFIG_DIRECTORY="$(ConfigDirectory)";_DEBUG;NATIVERENDER_EXPORTS;_WINDOWS;_USRDLL;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions>PEW_EOS_EXPORT;OVERLAY_DLL_NAME="$(OverlayDllName)";STEAM_API_DLL="$(SteamApiDll)";SDK_DLL_NAME="$(SdkDllName)";CONFIG_DIRECTORY="$(ConfigDirectory)";_DEBUG;NATIVERENDER_EXPORTS;_WINDOWS;_USRDLL;%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
paulhazen, line 81: This is where the preprocessor definition exists that determines the behavior of the `PEW_EOS_API` and `PEW_EOS_API_FUNC` macros.

Previously - a manually coded `#define` was utilized to control the behavior.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/string_helpers.cpp
```diff
@@ -25,9 +25,9 @@
 #include <codecvt>
 #include <ctime>
 #include <locale>
-#include <string_helpers.h>
 #include <sstream>
 #include <filesystem>
+#include "string_helpers.h"
```
paulhazen, line 30: This was changed because when this particular file gets included as part of another binary, the angle bracket causes the linker to look in standard include directories instead of in ones that are common to the project - meaning when another project referenced these files and this `.cpp` was absorbed, it was unable to find the header. Changing it to be quotes resolves that issue by telling the linker to look inside the project.

andrew-hirata-playeveryware (approved):

PR #1030, opened 2024-11-27
chore: (Re)introduce `ManagedPluginCode` project.

paulhazen:
This PR introduces a new class library for the Windows native code solution, containing only the essential files needed to replicate the platform creation and initialization steps from the EOS Plugin for Unity. While this library does not perform any tasks in this PR, it lays the groundwork for functionality to be added in a future pull request.

The class library is exclusively intended for use in the Debug configuration of the `DynamicLibraryLoaderHelper` solution. Its purpose is to validate access to test for parity between the native and managed code implementations.

> [!NOTE]
> *HISTORICAL CONTEXT*
> In a prior iteration, a project similar to this one titled `EOSPluginConfig` was added to the native code component of the project. At the time, it's intended purpose was subtly different - instead of _testing parity_ between managed and unmanaged, it was an attempt to execute the managed implementation from within a native context. When we decided against this approach, the project was appropriately removed. 

#EOS-1982


Assets/Plugins/Windows/Core/WindowsPlatformSpecifics.cs
```diff
@@ -58,9 +70,7 @@ public class WindowsPlatformSpecifics : PlatformSpecifics<WindowsConfig>
 #if ENABLE_CONFIGURE_STEAM_FROM_MANAGED
         private static readonly string SteamDllName = 
 #if UNITY_64
-        "steam_api64.dll";
-#else
-        "steam_api.dll";
```
andrew-hirata-playeveryware, line 63: What happened to the steam DLL name?

andrew-hirata-playeveryware (changes requested):

Assets/Plugins/Windows/Core/WindowsPlatformSpecifics.cs
```diff
@@ -58,9 +70,7 @@ public class WindowsPlatformSpecifics : PlatformSpecifics<WindowsConfig>
...
+	@@ -61,22 +79,22 @@ public class WindowsPlatformSpecifics : PlatformSpecifics<WindowsConfig>
```
WispyMouse: This *looks* like a merge conflict.

Assets/Plugins/Windows/Core/WindowsPlatformSpecifics.cs
WispyMouse: This does seem to be an error; under this state the "SteamDllName" might not have a value on set to it.

lib/NativeCode/DynamicLibraryLoaderHelper/DynamicLibraryLoaderHelper.sln
```diff
@@ -39,6 +41,12 @@ Global
 		{D7462EAF-0024-4EA5-9D89-DBBBC7766333}.Debug|x86.Build.0 = Debug|Win32
 		{D7462EAF-0024-4EA5-9D89-DBBBC7766333}.Release|x64.ActiveCfg = Release|x64
 		{D7462EAF-0024-4EA5-9D89-DBBBC7766333}.Release|x86.ActiveCfg = Release|Win32
+		{B3ADEB07-B668-482E-A5AA-9671CE65C5F7}.Debug|x64.ActiveCfg = Debug|x64
+		{B3ADEB07-B668-482E-A5AA-9671CE65C5F7}.Debug|x64.Build.0 = Debug|x64
+		{B3ADEB07-B668-482E-A5AA-9671CE65C5F7}.Debug|x86.ActiveCfg = Debug|x86
```
WispyMouse, line 46: I don't understand these configuration things, so you may already have considered this; it looks like "x86" builds use "Win32" above, but use "x86". Is that intentional?

lib/NativeCode/DynamicLibraryLoaderHelper/ManagedPluginCode/Application.cs
```diff
@@ -0,0 +1,44 @@
...
+    /// so that files compiled outside of the Unity Editor that reference that
+    /// class can still be compiled properly.
+    /// </summary>
+    internal static class Application
+    {
+        /// <summary>
+        /// This is the path of the streaming assets directory relative to the
+        /// output directory of this class library.
+        /// </summary>
+        public static readonly string streamingAssetsPath = @"..\..\..\..\Assets\StreamingAssets\";
```
WispyMouse, line 36: I wish I could come up with a better solution than this... but I can't at this moment. Probably fine to keep.

lib/NativeCode/DynamicLibraryLoaderHelper/ManagedPluginCode/Application.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        /// <summary>
+        /// For the purposes of this class library, this field member must be
+        /// present - but it's value is not utilized, so it is being left empty.
+        /// </summary>
+        public static readonly string temporaryCachePath = string.Empty;
```
WispyMouse, line 42: If this should never be _utilized_, is there a world where you turn this into a {get} accessor that throws a not-implemented-exception?

WispyMouse (changes requested):
There appears to be a merge conflict and possible #ifdef error on `WindowsPlatformSpecifics`. Can those be addressed?

lib/NativeCode/DynamicLibraryLoaderHelper/DynamicLibraryLoaderHelper.sln
paulhazen: It is indeed intentional.

Assets/Plugins/Windows/Core/WindowsPlatformSpecifics.cs
paulhazen: Ah - yes good call. Combination of merge conflict and accidental removal of the definition here. It has now been restored and resolved.

Assets/Plugins/Windows/Core/WindowsPlatformSpecifics.cs
paulhazen: Marking as resolved - since it was addressed in another conversation.

lib/NativeCode/DynamicLibraryLoaderHelper/DynamicLibraryLoaderHelper.sln
paulhazen: The reasons are weird and complicated - but there are portions of the project that are Win32 - but need to be in a directory marked 'x86'.

lib/NativeCode/DynamicLibraryLoaderHelper/ManagedPluginCode/Application.cs
paulhazen: Me too. Although I don't believe this value is even utilized.

lib/NativeCode/DynamicLibraryLoaderHelper/ManagedPluginCode/Application.cs
paulhazen: The reason I am not concerned about this is twofold:

1. This particular field member only needs to exist so that the code can compile
2. The _meaning_ of this value is irrelevant, since this is one of the _only two_ code files that this project owns - and this project will only be used to test the verisimilitude of the config values parsed by managed code and unmanaged code (as in - this file doesn't exist outside of a unit test).

WispyMouse (approved):

andrew-hirata-playeveryware (approved):

PR #1036, opened 2024-11-30
fix: Make `GfxPluginNativeRender` function properly with new config.

paulhazen:
This PR fixes the native component of the plugin to function properly with the new config files.

It has been tested and is functioning properly, but this PR does not include the unit tests that prove that.

I recommend that because `development` is in a broken state because of the native code component when in standalone on Windows, we get these changes in and supplement them with the unit testing.

One area that still needs to be tested is the steam integration component. I have been limited in my ability to test that due to a lack of understanding all the mechanics of that process.

#EOS-2231


lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/ConfigPaths.props
```diff
@@ -1,11 +1,11 @@
 <Project>
   <PropertyGroup Condition="'$(Configuration)'=='Debug'">
     <!-- This is the filepath relative to the executable when debugging -->
-    <ConfigDirectory>$([System.String]::Copy('$(MSBuildThisFileDirectory)').Replace('\', '/'))../../../../Assets/StreamingAssets/EOS/</ConfigDirectory>
+    <ConfigDirectory>../../../../Assets/StreamingAssets/EOS/</ConfigDirectory>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)'=='Release'">
     <!-- This is the filepath relative to the executable (dll) when in release mode -->
-    <ConfigDirectory>$([System.String]::Copy('$(MSBuildThisFileDirectory)').Replace('\', '/'))../../StreamingAssets/EOS/</ConfigDirectory>
+    <ConfigDirectory>../../StreamingAssets/EOS/</ConfigDirectory>
```
paulhazen, line 8: I think these can be removed entirely - the values are determined entirely at runtime I think.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/ProductConfig.hpp
```diff
@@ -0,0 +1,78 @@
...
+
+namespace pew::eos::config
+{
+    struct ProductConfig final : Config
+    {
+        std::string product_name;
+        std::string product_id;
+        std::string product_version;
+        ProductionEnvironments environments;
+        std::vector<ClientCredentials> clients;
```
paulhazen: Add comments to describe what each field member does.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/ProductionEnvironments.hpp
```diff
@@ -0,0 +1,150 @@
+#ifndef PRODUCTION_ENVIRONMENTS_H
+#define PRODUCTION_ENVIRONMENTS_H
```
paulhazen: Change this to be `PRODUCTION_ENVIRONMENTS_HPP`

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/Serializable.hpp
```diff
@@ -0,0 +1,221 @@
...
+        }
+
+        template <typename T, typename = std::enable_if_t<
+            std::is_same_v<T, int> || 
+            std::is_same_v<T, double> || 
+            std::is_same_v<T, long> ||
+            std::is_same_v<T, float> ||
+            std::is_same_v<T, uint64_t>
+        >>
+        static T parse_number(json_value_s& json_value)
```
paulhazen: Function comments are needed.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/Serializable.hpp
```diff
@@ -0,0 +1,221 @@
...
+            else if constexpr(std::is_same_v<T, uint64_t>)
+            {
+                char* end;
+                number_value = (number == nullptr) ? 0 : std::strtoull(number->number, &end, 10);
+            }
+
+            return number_value;
+        }
+
+        static bool parse_bool(const json_value_s& json_value)
```
paulhazen, line 161: Function comments are needed.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/Serializable.hpp
```diff
@@ -0,0 +1,221 @@
...
+            }
+        }
+
+        template<typename T, typename = std::enable_if_t< 
+            std::is_same_v<T, int> ||
+            std::is_same_v<T, EOS_EAuthScopeFlags> ||
+            std::is_same_v<T, EOS_EIntegratedPlatformManagementFlags> ||
+            std::is_same_v<T, EOS_UI_EInputStateButtonFlags>
+            >>
+        static T parse_flags(const std::map<std::string, T>* strings_to_enum_values, T default_value, json_value_s* value)
```
paulhazen: Function comments are needed.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/SteamConfig.h
```diff
@@ -0,0 +1,64 @@
+#ifndef STEAM_CONFIG_H
+#define STEAM_CONFIG_H
```
paulhazen: Convert to `.hpp` file and change these defines accordingly.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/Version.hpp
```diff
@@ -0,0 +1,120 @@
+#ifndef VERSION_H
+#define VERSION_H
```
paulhazen: Rename macros accordingly.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/config_legacy.h
```diff
@@ -255,7 +323,7 @@ namespace pew::eos::config_legacy
      * \brief Parses an EOS configuration object from a JSON structure.
      * \return An `EOSConfig` object populated with settings from the JSON structure.
      */
-    CONFIG_API bool try_get_eos_config(EOSConfig& config);
+    //CONFIG_API bool try_get_eos_config(EOSConfig& config);
```
paulhazen: More properly remove this.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/config_legacy.h
```diff
@@ -118,6 +87,105 @@ namespace pew::eos::config_legacy
...
+        { "Special_Left",      EOS_UI_EInputStateButtonFlags::EOS_UISBF_Special_Left      },
+        { "SpecialLeft",       EOS_UI_EInputStateButtonFlags::EOS_UISBF_Special_Left      },
+
+        { "Special_Right",     EOS_UI_EInputStateButtonFlags::EOS_UISBF_Special_Right     },
+        { "SpecialRight",      EOS_UI_EInputStateButtonFlags::EOS_UISBF_Special_Right     },
+
+        { "LeftThumbstick",    EOS_UI_EInputStateButtonFlags::EOS_UISBF_LeftThumbstick    },
+        { "RightThumbstick",   EOS_UI_EInputStateButtonFlags::EOS_UISBF_RightThumbstick   },
+    };
+
```
paulhazen: Reconsider adding these to the `config_legacy` namespace - since they are new and clearly not legacy.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/pch.h
```diff
@@ -31,10 +31,12 @@
...
 
-#define EOS_SERVICE_CONFIG_FILENAME "EpicOnlineServicesConfig.json"
+#define EOS_WINDOWS_CONFIG_FILENAME "eos_windows_config.json"
+#define EOS_PRODUCT_CONFIG_FILENAME "eos_product_config.json"
+
 #define EOS_STEAM_CONFIG_FILENAME "eos_steam_config.json"
 #define EOS_LOGLEVEL_CONFIG_FILENAME "log_level_config.json"
 
-#define RESTRICT __restrict
+//#define RESTRICT __restrict
```
paulhazen, line 46: Figure out why this was originally needed, and if there isn't a problem than remove it properly.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/config_legacy.cpp
```diff
@@ -175,28 +175,43 @@ namespace pew::eos::config_legacy
...
         }
 
+        // Set the config reference by parsing the json values into the object
         config = eos_config_from_json_value(eos_config_as_json);
+
+        // Indicate successful parsing of the eos config.
         return true;
     }
-
+    */
```
paulhazen: More properly remove this unused code.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/dllmain.cpp
```diff
@@ -125,21 +126,15 @@ void get_cli_arguments(config_legacy::EOSConfig eos_config)
 #if PLATFORM_32BITS
 #pragma comment(linker, "/export:UnityPluginLoad=_UnityPluginLoad@4")
 #endif
-PEW_EOS_API_FUNC(void) UnityPluginLoad(void*)
+PEW_EOS_API_FUNC(void) UnityPluginLoad(void* arg)
 {
+    // TODO: Only display this dialog when not running in debug from within
+    //       Visual Studio.
```
paulhazen: Remove comment - since this has been accomplished, as is clear from the changes below.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/dllmain.cpp
```diff
@@ -161,9 +156,14 @@ PEW_EOS_API_FUNC(void) UnityPluginLoad(void*)
         {
             logging::log_inform("start eos init");
 
-            eos_init(eos_config);
+            const auto product_config = config::Config::get<config::ProductConfig>();
+            const auto windows_config = config::Config::get<config::WindowsConfig>();
+
+            // TODO: Apply CLI arguments
```
paulhazen: Do the thing.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/eos_helpers.cpp
```diff
@@ -243,28 +289,32 @@ namespace pew::eos
...
-        platform_options.ProductId = eos_config.productID.c_str();
-        platform_options.SandboxId = eos_config.sandboxID.c_str();
-        platform_options.DeploymentId = eos_config.deploymentID.c_str();
-        platform_options.ClientCredentials.ClientId = eos_config.clientID.c_str();
-        platform_options.ClientCredentials.ClientSecret = eos_config.clientSecret.c_str();
+        platform_options.EncryptionKey = platform_config.client_credentials.encryption_key.c_str();
+
+        // TODO: Determine why now this cannot be empty.
+        platform_options.OverrideCountryCode = "US";// platform_config.overrideCountryCode.c_str();
+        platform_options.OverrideLocaleCode = "en-US"; // platform_config.overrideLocaleCode.c_str();
```
paulhazen: Again - do the thing. Maybe they can't be empty strings, but they can be set to `nullptr`? This would be consistent with the prior implementation.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/json_helpers.cpp
```diff
@@ -113,24 +113,38 @@ namespace pew::eos::json_helpers
...
-        const size_t config_file_size = static_cast<size_t>(file_size(path_to_config_json));
+        logging::log_inform("Reading json from file \"" + path_to_config_json.string()  + "\".");
+        const auto config_file_size = file_size(path_to_config_json);
         if (config_file_size > SIZE_MAX)
         {
             throw std::filesystem::filesystem_error("File is too large", std::make_error_code(std::errc::file_too_large));
         }
 
+        // TODO: Avoid explicit calls to calloc - leverage RAII for automatic
+        //       memory management instead.
```
paulhazen: +1

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/string_helpers.cpp
```diff
@@ -101,7 +101,7 @@ namespace pew::eos::string_helpers
     }
 
     // wide_str must be null terminated if wide_str_len is passed
-    bool copy_to_utf8_str_from_wide_str(char* RESTRICT utf8_str, size_t utf8_str_len, const wchar_t* RESTRICT wide_str, int wide_str_len)
```
paulhazen, line 104: Again (as mentioned in earlier comment) determine why these were originally here, and if it's a problem that they are now removed.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/ConfigPaths.props
paulhazen: This is not correct - they are in fact still being used.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/dllmain.cpp
paulhazen: Did the thing - CLI arguments are now being applied appropriately.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/string_helpers.cpp
paulhazen, line 104: The `RESTRICT` macro was used here (and here only).

It expands to `__restrict`, which communicates to the compiler that the pointers `utf8_str` and `wide_str` will never point to overlapping memory regions. This "promise" is purely a contract between the developer and the compiler, with no runtime checks to enforce it. Based on this assumption, the compiler may perform additional optimizations, such as avoiding redundant memory loads or enabling faster loop unrolling.

However, the performance benefits in this case are negligible. This function is only used with very small strings, where such memory optimizations are unlikely to make a measurable difference.

By removing `__restrict`, the code becomes safer and more maintainable. If a mistake is ever made and the pointers actually overlap, the absence of `__restrict` ensures that the compiler will generate correct, albeit slightly less optimized, code. In contrast, leaving `__restrict` could lead to undefined behavior that would be difficult to diagnose.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/eos_helpers.cpp
paulhazen: Solution was implemented so that if either of the strings for `overrideLocaleCode` or `overrideCountryCode` is empty, the respective parameters given to create the platform are `nullptr`. This restores original implementation prior to this PR - so this conversation is being marked as resolved.

Assets/Plugins/Source/Editor/Configs/SteamConfig.cs
```diff
@@ -94,7 +94,7 @@ static SteamConfig()
             RegisterFactory(() => new SteamConfig());
         }
 
-        protected SteamConfig() : base("eos_plugin_steam_config.json")
+        protected SteamConfig() : base("eos_steam_config.json")
```
WispyMouse, line 105: Why are we changing the name on this? Will users need to migrate their settings from the previous config to this new one?

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/ClientCredentials.hpp
```diff
@@ -0,0 +1,67 @@
...
+    struct ClientCredentials final : Serializable
+    {
+        std::string client_id;
+        std::string client_secret;
+        std::string encryption_key;
+
+    protected:
+        void parse_json_element(const std::string& name, json_value_s& value) override
+        {
+            if (name == "Value")
```
WispyMouse, line 47: notfeedback; I wish that this version of C++ had a similar built in function to C#'s `nameof`. I feel a bit stressed comparing values directly to strings in such a fragile way. As far as I can tell though, JSON is case sensitive at the relevant parts, so this has to be valued exactly as mentioned here... but I'm afraid of what happens if for some reason these values have their name changed.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/Config.hpp
```diff
@@ -0,0 +1,181 @@
...
+            // Read the values from the file
+            config->read();
+
+            return config;
+        }
+
+    private:
+
+        // Depending on the configuration (debug or release) these are the possible relative paths to the config directory
+        static constexpr std::array<std::string_view, 2> s_possible_config_directories = {
```
WispyMouse: notfeedback; This is somewhat horrifying to see in code, but I can't think of a better solution.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/Config.hpp
```diff
@@ -0,0 +1,181 @@
...
+         */
+        Version _schema_version;
+
+        /**
+         * \brief Create a new Config class.
+         * \param file_name The fully qualified path to the config file.
+         */
+        Config(const std::filesystem::path& file_name)
+        {
+            _file_path = std::filesystem::path(get_config_directory()) / file_name;
```
WispyMouse: notfeedback; Learned something about c++ from this! I was like, "why are you dividing a file path by a name?" but apparently "path" implements "/" as a way of combining paths. Wild.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/Config.hpp
```diff
@@ -0,0 +1,181 @@
...
+            std::ostringstream buffer;
+            buffer << file.rdbuf();
+            const std::string json_content = buffer.str();
+
+            json_value_s* json_value = json_parse(json_content.c_str(), json_content.length());
+            const auto json_object = static_cast<json_object_s*>(json_value->payload);
+
+            from_json(*json_object);
+
+            free(json_value);
```
WispyMouse, line 199: I don't know `free`; it looks like a way of intentionally signaling to let go of resources. Should we `free` the `json_object`, after it has been used in `from_json`?

WispyMouse: What about the stringstream, `buffer`?

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/EnumMappings.h
```diff
@@ -0,0 +1,131 @@
...
+#include <map>
+
+namespace pew::eos
+{
+    /**
+     * \brief Maps string values to specific auth scope flags defined within the
+     * EOS SDK.
+     */
+    const std::map<std::string, EOS_EAuthScopeFlags> AUTH_SCOPE_FLAGS_STRINGS_TO_ENUM = {
+        { "NoFlags",           EOS_EAuthScopeFlags::EOS_AS_NoFlags           },
```
WispyMouse, line 36: I noticed that `PLATFORM_CREATION_FLAGS_STRINGS_TO_ENUMS` specifies both shorthand names and longform enum names. Should this also represent longform enum names? E.x. `{ "EOS_AS_NoFlags", EOS_EAuthScopeFlags::EOS_AS_NoFlags}`?

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/SteamConfig.hpp
```diff
@@ -0,0 +1,146 @@
...
+                for (char currentCharacter : currentFullValue)
+                {
+                    steamApiInterfaceVersionsAsCharArray.push_back(currentCharacter);
+                }
+
+                steamApiInterfaceVersionsAsCharArray.push_back('\0');
+            }
+            steamApiInterfaceVersionsAsCharArray.push_back('\0');
+
+            return steamApiInterfaceVersionsAsCharArray.data();
```
WispyMouse, line 86: I don't know if this is significant, but I get a warning from this:

C26816: The pointer points to memory allocated on the stack.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/SteamConfig.hpp
```diff
@@ -0,0 +1,146 @@
...
+    private:
+        std::filesystem::path _library_path;
+        std::filesystem::path _override_library_path;
+        std::vector<std::string> _steam_api_interface_versions_array;
+
+        explicit SteamConfig() : Config("eos_steam_config.json"),
+            steam_sdk_major_version(0),
+            steam_sdk_minor_version(0)
+        {
+            // TODO: Restore STEAM_SDK_DLL_NAME
```
WispyMouse, line 110: Does this need to be done? It appears as though the override path isn't available either.

WispyMouse (changes requested):
Requesting changes for reasons mentioned in Slack.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/SteamConfig.hpp
paulhazen: It does need to be done - but restoring the functionality of the steam component is a task that is intended to take place after the restoration of the Windows Standalone Build.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/EnumMappings.h
paulhazen: No currently - this map is built from the implementation that exists within the plugin - if it needs to be changed it should be changed in both places at the same time. Because the plugin code serializes values taken from a C# wrapper enum provided by the EOS SDK's C# wrapper - the names written to file are different than the names used in the C++.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/Config.hpp
paulhazen: Not according to the documentation provided by the JSON library that we utilize. See [the repository](https://github.com/sheredom/json.h) for that JSON library for details about why `free` is utilized here and not on `json_object`.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/PlatformConfig.hpp
```diff
@@ -0,0 +1,264 @@
...
+#include "eos_ui_types.h"
+#include "json_helpers.h"
+#include "ProductionEnvironments.hpp"
+#include "include/Config/ClientCredentials.hpp"
+#include "include/Config/Config.hpp"
+
+namespace pew::eos::config
+{
+    
+
```
paulhazen: Remove extra spaces.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/Config.hpp
paulhazen: Yep! One of the handy things about being able to define operator overloads in C++. You can do the same thing in C#, but it's not done as frequently as it seems to be done in C++.

Assets/Plugins/Source/Editor/Configs/SteamConfig.cs
paulhazen: The name was changed in error in another PR. The thought process was that the configuration values in `eos_steam_config.json` are more _plugin related_ than _platform related_, so I had tinkered with the idea of moving this configuration value to `etc/config/`, and had changed the name of the file to match the pattern used in that directory, and to indicate that it was a _plugin_ setting, not an _EOS_ or _platform_ setting.

This line of thinking was incorrect for a variety of reasons, and the original location and name of the file was restored. 

This is evidenced by the fact that in `stable` this file is still named `eos_steam_config.json`, and is still placed in `Assets/StreamingAssets/EOS`.

Good to have noticed this - but this is changing it _back_ to what it was - so there is no need to perform migration or worry about it.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/PlatformConfig.hpp
paulhazen: Done.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/SteamConfig.hpp
paulhazen: It _is_ significant! And good catch. I updated the implementation so that instead of returning a `const char*` that was allocated on the stack (and is therefore not guaranteed to be valid after returning), I changed the function to return a `std::string` value, which is handled appropriately. 

The place where the return value of this function is applied to the EOS SDK was also updated to append the `.c_str()` method so that when the value is assigned it's done correctly.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/Config.hpp
paulhazen: I agree completely. Part of the reason that the first entry in this array is the one that is utilized during release mode - so that in release the correct one is used automatically. A strong argument could be made for removing it altogether, since there are only two entries.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/ClientCredentials.hpp
paulhazen: I _completely_ agree with you on this one. What should be of some comfort when it comes to this PR is that the previous implementation was fragile in the exact same manner.

WispyMouse (approved):

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/Config.hpp
```diff
@@ -0,0 +1,181 @@
...
+         * parameter given.
+         * \tparam T The Config-type-derived class that is being retrieved.
+         * \return An instance of a class derived from Config that contains all
+         * values.
+         */
+        template <typename T>
+        static std::enable_if_t<std::is_base_of_v<Config, T>, std::unique_ptr<T>> get()
+        {
+            // Create the config class
+            auto config = std::unique_ptr<T>(new T());
```
andrew-hirata-playeveryware, line 70: If we are going to use `unique_ptr`, I would recommend using std::make_unique instead of manually calling `new`.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/Config.hpp
```diff
@@ -0,0 +1,181 @@
...
+            "../../StreamingAssets/EOS/",
+            // This is the relative path for the config files when in debug mode
+            // (specifically when the library is loaded in the Build\Debug
+            // directory).
+            "../../../../../../Assets/StreamingAssets/EOS/",
+        };
+
+        static inline std::filesystem::path s_config_directory;
+
+        static std::filesystem::path get_config_directory()
```
andrew-hirata-playeveryware, line 98: Unless I'm mistaken, this code doesn't handle the same cases that the previous code did. The other code would fallback to checking inside of a `Data` subdirectory and this does not. I don't remember why this change occurred (it might be a Unity 2019 based hierarchy). Why is that directory not checked now?

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/Config.hpp
andrew-hirata-playeveryware: This looks like it's creating a duplicate `std::filesystem::path` object, is it not? The `get_config_directory` method in this class returns one already; doing this will create duplicate object.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/Config.hpp
andrew-hirata-playeveryware: I'm not 100% certain, but I think this code will break on user's machines that have Kanji characters in their path: is this something that could be tested? 

I seem to remember needing to rewrite the original version of code once to use the win32 `_wfopen_s` API to handle non-latin script character sets.

And to confirm, `json_parse` returns a `malloc`'d object that has to be `free`'d.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/PlatformConfig.hpp
```diff
@@ -0,0 +1,262 @@
...
+         * combination is selected.
+         */
+        EOS_UI_EInputStateButtonFlags toggle_friends_button_combination = EOS_UI_EInputStateButtonFlags::EOS_UISBF_None;
+
+        /**
+         * \brief The country code to use when creating the eos platform. Note
+         * that the Unity project does not currently expose this property for
+         * the user to edit. TODO: Expose property in Unity project.
+         */
+        std::string overrideCountryCode;
```
andrew-hirata-playeveryware, line 130: I feel like these are options that are supposed to come from the command line, not something that the user specifies?

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/PlatformConfig.hpp
```diff
@@ -0,0 +1,262 @@
...
+                    {"P2PIo", &thread_affinity.P2PIo},
+                    {"HttpRequestIo", &thread_affinity.HttpRequestIo},
+                    {"RTCIo", &thread_affinity.RTCIo},
+                    {"EmbeddedOverlayMainThread", &thread_affinity.EmbeddedOverlayMainThread},
+                    {"EmbeddedOverlayWorkerThreads", &thread_affinity.EmbeddedOverlayWorkerThreads},
+                };
+
+                while(thread_affinity_iterator)
+                {
+                    auto element = *(thread_affinity_iterator->value);
```
andrew-hirata-playeveryware: I would move this fetch of the element into the if statement.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/PlatformConfig.hpp
```diff
@@ -0,0 +1,262 @@
...
+            }
+            else if (name == "threadAffinity")
+            {
+                const auto thread_affinity_json_object = *static_cast<json_object_s*>(value.payload);
+                auto thread_affinity_iterator = thread_affinity_json_object.start;
+
+                // TODO: Figure out how to implement this clean approach with
+                //       all the properties of any class deriving from
+                //       Serializable.
+                std::unordered_map<std::string, uint64_t*> affinity_map = {
```
andrew-hirata-playeveryware, line 233: In general, I would not use a map for this sort of thing, if only because doing the additional allocations / find feels unnecessary here. That being said, this is not a really high performance part of code so I'll also accept this.

andrew-hirata-playeveryware (changes requested):

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/PlatformConfig.hpp
paulhazen: That's a good point. As a non-performance component of code meant to run once I do prefer the readability - and I _believe_ that the compiler is capable of and may in fact optimize this considerably - perhaps even more so if it's marked `static` and/or `const`.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/PlatformConfig.hpp
paulhazen: I assume you mean the `while` conditional clause?

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/PlatformConfig.hpp
paulhazen: Our current implementation within the native code is to _attempt_ to read `overrideCountryCode` and `overrideLocaleCode` from the JSON. In the event that the field cannot be found in the file (which in our case is always because we never write it), a `nullptr` value is assigned. We do not currently provide a mechanism to override this from the command line - whether that be through the native or managed components of the plugin.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/Config.hpp
paulhazen: `Config` class enforces the factory pattern such that deriving objects can only be instantiated from within the `get` function. This means that private constructors are not accessible to `std::make_unique`. That means that if (in this case) you attempt to utilize `std::make_unique<T>()`, you will get the following compiler error:

`C2248: Cannot access private member declared in class.`

Using `new` with `std::unique_ptr` still retains the core benefits of `std::unique_ptr`, mainly automatic scope management that guarantees proper deletion once the `std::unique_ptr` goes out of scope or is reset, and it ensures that there is only one owner of the dynamically allocated memory. The _only_ component that is lost in this approach is that the **exception handling** that `std::make_ptr` provides is not employed.

The alternative would be to create a custom version of `std::make_unique` that does some fanciness:

```cpp
template <typename T, typename... Args>
std::unique_ptr<T> make_unique_private(Args&&... args)
{
    return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
}
```

Then utilizing it like this:

```cpp
auto config = make_unique_private<T>();
```

But that solution felt a little bit convoluted since it gets around the private constructor access limitation through clever use of templating combined with `std::forward` - and since the solution _still_ doesn't utilize `std::make_unique`, it seemed that this approach would be more difficult to maintain.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/Config.hpp
paulhazen: The current implementation in `development` is as follows:

https://github.com/PlayEveryWare/eos_plugin_for_unity/blob/88e6e1b972a2f4ee4649480b88c1adf608422ef6/lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/config_legacy.cpp#L248-L251

and `CONFIG_DIRECTORY` is provided as a compiler preprocessing directive.

The implementation that exists within `stable` is as follows (note that you'll need to scroll to see the full function):

https://github.com/PlayEveryWare/eos_plugin_for_unity/blob/e8856b2c2349e2d7e66be225800e6f46d2f640eb/lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/dllmain.cpp#L1171-L1186

So, it appears as though this was a regression that slipped through a previous PR without notice.

Do we need to restore this functionality? I did not test with `Unity 2019` - doing so did not seem necessary but were we to restore that as a possible path, that would be the first step I would take toward confirming that was the reason for the path to fall back to that.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/PlatformConfig.hpp
andrew-hirata-playeveryware: I mean around line 223, where it uses the element.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/ProductConfig.hpp
```diff
@@ -0,0 +1,99 @@
...
+                product_version = json_value_as_string(&value)->string;
+            }
+            else if (name == "Clients")
+            {
+                clients = parse_json_array<ClientCredentials>(value);
+            }
+            else if (name == "Environments")
+            {
+                // Parse environments
+                auto parsed_environments = ProductionEnvironments();
```
andrew-hirata-playeveryware, line 92: I assume this is supposed to be reading into `environments`, and not `parsed_environments`? Otherwise I believe that `environments` will be it's default value.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/ProductionEnvironments.hpp
```diff
@@ -0,0 +1,149 @@
...
+                {
+                    from_json(value);
+                }
+                else if (name == "DeploymentId")
+                {
+                    id = json_value_as_string(&value)->string;
+                }
+                else if (name == "SandboxId")
+                {
+                    auto sandbox_temp = Sandbox();
```
andrew-hirata-playeveryware, line 87: Why does this use a temp variable instead of directly using the member variable?

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/ProductionEnvironments.hpp
```diff
@@ -0,0 +1,149 @@
...
+
+        /**
+         * \brief Determines if a given deployment id is defined in the product
+         * config or not.
+         * \param deployment_id The deployment to search for.
+         * \return True if the deployment is defined, false otherwise.
+         */
+        bool is_deployment_defined(const std::string& deployment_id) const
+        {
+            return std::any_of(deployments.begin(), deployments.end(), [deployment_id](const Deployment& d)
```
andrew-hirata-playeveryware: Why is the `deployment_id` being captured by value instead of by reference ?

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/ProductionEnvironments.hpp
```diff
@@ -0,0 +1,149 @@
...
+         * \brief Determines if a given sandbox is defined in the product config
+         * or not.
+         * \param sandbox_id The sandbox id to search for.
+         * \return True if the sandbox is defined, false otherwise.
+         */
+        bool is_sandbox_defined(const std::string& sandbox_id) const
+        {
+            return std::any_of(sandboxes.begin(), sandboxes.end(), [sandbox_id](const Sandbox& s)
+                {
+                    return s.id == sandbox_id;
```
andrew-hirata-playeveryware, line 128: Same here: I don't believe that it needs to capture the `sandbox_id` by value.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/PlatformConfig.hpp
paulhazen: Ah - I see - you mean declaring `element` closer to where it is utilized, instead of having it up here.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/Config.hpp
paulhazen: Per conversations - yes: the ideal thing would be to restore original functionality.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/Config.hpp
paulhazen: Per conversation - this makes sense - but a comment should be added where the `new` is used to clarify why it is needed.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/PlatformConfig.hpp
paulhazen: It has been moved. Marking this conversation as resolved.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/Config.hpp
paulhazen: Comment has been added as discussed, marking this conversation as resolved.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/Config.hpp
paulhazen: Functionality compatible with the previous path determination was added to the current implementation - so effectively original functionality was restored. Marking this conversation as resolved.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/Serializable.hpp
```diff
@@ -0,0 +1,275 @@
...
+                      std::is_same_v<T, int> || 
+                      std::is_same_v<T, double> || 
+                      std::is_same_v<T, long> ||
+                      std::is_same_v<T, float> ||
+                      std::is_same_v<T, uint64_t> ||
+                      std::is_same_v<T, uint32_t>
+                  >>
+        static T parse_number(json_value_s& json_value)
+        {
+            T number_value;
```
andrew-hirata-playeveryware: I think it might be impossible for this to happen,  but I notice that `number_value` isn't assigned a value unless it matches one of the if statements.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/Config.hpp
paulhazen: Apparently, the correct way to handle Kanji characters is to do the following:

Replace line `173`:
```cpp
const std::ifstream file(_file_path);
```

with the following:

```cpp

    std::ifstream file;

    // Platform-specific handling of file opening
#ifdef _WIN32
    // On Windows, use wide-character API for Unicode support
    file.open(_file_path.wstring(), std::ios::in);
#else
    // On Linux/macOS, open the file normally (UTF-8 is native)
    file.open(_file_path, std::ios::in);
#endif
```

Testing by doing this demonstrates that it works correctly:

```cpp
int main() {
    // Example path with Kanji characters
    std::filesystem::path file_path = "config/例.json";

    read(file_path);

    return 0;
}
```

@andrew-hirata-playeveryware Can you confirm that implementing these changes would resolve the issue?

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/Config.hpp
paulhazen: Good catch. I have updated it to not do that. Marking conversation as resolved.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/Config.hpp
paulhazen: Implemented changes as discussed above - marking conversation as resolved.

lib/NativeCode/DynamicLibraryLoaderHelper/DynamicLibraryLoaderHelper/DynamicLibraryLoaderHelper.vcxproj
```diff
@@ -78,6 +78,18 @@
       <Optimization>Disabled</Optimization>
       <PreprocessorDefinitions>_DEBUG;_LIB;%(PreprocessorDefinitions)</PreprocessorDefinitions>
     </ClCompile>
+    <PostBuildEvent>
+      <!-- Copy the meta file stored in the solution directory, since it's values have been properly set in such a way that Unity interacts with the DLL in the intended manner. -->
+      <Command>
+      @echo Copying "$(SolutionDir)$(ProjectName)-$(NormalizedPlatform).dll.meta" to "$(OutputUnityAssetsDirectory)"
```
andrew-hirata-playeveryware, line 84: Does this change the dll name when it uses NormalizedPlatform?

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/SteamConfig.hpp
andrew-hirata-playeveryware: Not a required change: I believe since C++11 you can use `string::insert` to add a range of chars. So if you change `steamApiInterfaceVersionsAsCharArray` from a `std::vector<char>` to a `std::string` I believe you should be able to change the code to something like this, and avoid the copy or potentially as many allocations (depending on how often `push_back` causes an alloc to occur): 

```cpp
std::string steamApiInterfaceVersionsAsCharArray;

for (const auto& currentFullValue : _steam_api_interface_versions_array) {
   steamApiInterfaceVersionsAsCharArray.insert(steamApiInterfaceVersionsAsCharArray.end,
                                                                            currentFullValue.begin(),
                                                                            currentFullValue.end()
   );
   steamApiInterfaceVersionsAsCharArray.push_back('\0');
}
steamApiInterfaceVersionsAsCharArray.push_back('\0');

return steamApiInterfaceVersionsAsCharArray.data();
```

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/Version.hpp
```diff
@@ -0,0 +1,120 @@
...
+#include <sstream>
+#include <string>
+
+namespace pew::eos::config
+{
+    struct Version
+    {
+        int major = 0;
+        int minor = 0;
+        int part = 0;
```
andrew-hirata-playeveryware: I would prefer the usage of 'patch' so that this matches the names in semver.

andrew-hirata-playeveryware (changes requested):

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/SteamConfig.hpp
WispyMouse: I think this could be a worthwhile change, but FWIW this isn't new code. It's moved from another place. I'll create a small task for this.

WispyMouse: #EOS-2279

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/Version.hpp
paulhazen: Makes sense. I've switched it to that.

lib/NativeCode/DynamicLibraryLoaderHelper/DynamicLibraryLoaderHelper/DynamicLibraryLoaderHelper.vcxproj
paulhazen: This change does not affect the DLL name. The concept of `NormalizedPlatform` - which translates `Win32` to `x86` - already exists in `development` and is defined here:
https://github.com/PlayEveryWare/eos_plugin_for_unity/blob/5b1577e82f9ac5f3bc4e72a42d676230cca11dd5/lib/NativeCode/DynamicLibraryLoaderHelper/Directory.Build.props#L13-L16

The purpose of `NormalizedPlatform` is to ensure consistency when the `$(Platform)` macro evaluates to `Win32` instead of `x86` in Windows-specific C# and C++ projects. This avoids having to use different platform values across various projects.

That said, this change does not introduce `NormalizedPlatform` or define the DLL output name. It only ensures that the correct `.dll.meta` file is copied to the appropriate directory.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/Serializable.hpp
paulhazen: The reason it does not need to be assigned a value is because the template parameter constraints require it to be a value-type - so it has a default value of `0`. That said, explicitly initializing the value will prevent potential issues if the function is extended or modified in the future - so I changed it to initialize to 0.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/ProductionEnvironments.hpp
paulhazen: I felt that capturing by value here was appropriate for a couple of convention reasons:
1. The lambda function should not alter the value - and if copied by value that isn't a risk
2. Copying the value guarantees that there are no dangling references.

For the simple implementation here - capturing by reference might be acceptable - but copying by value is far safer in my opinion.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/ProductionEnvironments.hpp
paulhazen: It's mostly for exception safety - the value of the field member is only updated if the `from_json` function executes completely.

It's also an artifact of debugging if memory serves - it was useful to determine that the sandbox was correctly parsed before assigning it to the field member.

Also, thanks to the implementation of `Sandbox`, the C++ compiler will optimize this with move semantics.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/ProductConfig.hpp
paulhazen: Good catch! The reason this didn't break during testing is that the `environments` field member on `ProductConfig` is not directly used - if the deployment information is defined within the `PlatformConfig` as it is in our case - then there is no need to look at this field member.

I've updated the implementation to mirror the flow of the sandbox parsing mentioned earlier - such that `parsed_environments` is assigned to `environments` after parsing compelets.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/dllmain.cpp
```diff
@@ -27,119 +27,75 @@
...
-    }
+    auto argument_stream = std::stringstream(GetCommandLineA());
+    const std::istream_iterator<std::string> argument_stream_begin(argument_stream);
+    const std::istream_iterator<std::string> argument_stream_end;
+    const std::vector argument_strings(argument_stream_begin, argument_stream_end);
 
-    //check if a deployment id override exists for sandbox id
-    for (unsigned i = 0; i < eos_config.sandboxDeploymentOverrides.size(); ++i)
+    std::string sandbox_id_override;
+    if (io_helpers::try_get_command_line_argument(argument_strings, sandbox_id_override, "epicsandboxid", "eossandboxid"))
```
andrew-hirata-playeveryware, line 62: I'm not certain why this was changed: the new version appears to iterate the command line arguments twice instead of going though it just once.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/dllmain.cpp
```diff
@@ -27,119 +27,75 @@
...
+        platform_config.deployment.id = deployment_id_override;
     }
 }
 
 // Called by unity on load. It kicks off the work to load the DLL for Overlay
 #if PLATFORM_32BITS
 #pragma comment(linker, "/export:UnityPluginLoad=_UnityPluginLoad@4")
 #endif
-PEW_EOS_API_FUNC(void) UnityPluginLoad(void*)
+PEW_EOS_API_FUNC(void) UnityPluginLoad(void* arg)
```
andrew-hirata-playeveryware, line 106: Adding `arg` here doesn't cause a warning if unused?

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/eos_helpers.cpp
```diff
@@ -190,39 +183,103 @@ namespace pew::eos
...
-        std::string steam_dll_path_as_string = steam_dll_path.string();
-        eos_call_steam_init(steam_dll_path_as_string);
-    }
-    
-    static void eos_call_steam_init(const std::string& steam_dll_path)
+    void eos_call_steam_init(const std::string& steam_dll_filename)
     {
-        auto steam_dll_path_string = io_helpers::get_basename(steam_dll_path);
-        HANDLE steam_dll_handle = GetModuleHandleA(steam_dll_path_string.c_str());
+        // Default (fallback) name of the steam dll to load.
```
andrew-hirata-playeveryware, line 189: I notice that this is now employing a different algorithm to find the DLL: it shouldn't stop if can't find it in the first place.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/eos_helpers.cpp
```diff
@@ -190,39 +183,103 @@ namespace pew::eos
...
+            // use load_library_at_path to load the dll - because _it_ uses the
+            // LoadLibrary method, different from the GetModuleHandleA functions
+            // used above.
+            //
+            // TODO: This seems very confusing and incorrect - it seems that all
+            //       functionality above this line should be included into the
+            //       "load_library_at_path" method - that way every time a
+            //       library load is attempted, both methods (GetModuleHandleA
+            //       and LoadLibrary) can be used if one of them fails.
+            if (steam_dll_handle == nullptr)
```
andrew-hirata-playeveryware, line 227: The difference between them is that GetModuleHandleA won't increase the ref count and doesn't load the library; load_library_at_path (or the original version) only loads the library from disk. The first reason for doing this is to ensure that we don't _have_ to leak a DLL reference / do a redundant load if one of the calls to GetModuleHandleA returns true i.e. Unity already loaded the library.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/eos_helpers.cpp
```diff
@@ -243,170 +300,166 @@ namespace pew::eos
         return s_tempPathBuffer;
     }
 
-    void eos_create(config_legacy::EOSConfig eos_config)
+    void apply_rtc_options(EOS_Platform_Options& platform_options)
```
andrew-hirata-playeveryware: I don't like the way this code has been split up which now requires it to use statics.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/eos_helpers.cpp
```diff
@@ -243,170 +300,166 @@ namespace pew::eos
...
+        const auto product_config = Config::get<ProductConfig>();
+
+        const auto create_options = get_create_options(*platform_config, *product_config);
+
+        // Allocate a new pointer so that it can be returned instead of the
+        // pointer to the object created on the stack above.
+        const auto platform_options_ptr = new EOS_Platform_Options();
+        *platform_options_ptr = create_options;
+
+        return platform_options_ptr;
```
andrew-hirata-playeveryware: I think this `platform_options_ptr` now refers to values that exist in `platform_config` and `product_config` that won't live outside of the scope of this function call; if I'm correct on this that means this will sometimes cause a crash due to stack memory being reused.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/eos_helpers.cpp
```diff
@@ -243,170 +300,166 @@ namespace pew::eos
...
+    }
+
+    PEW_EOS_API_FUNC(EOS_InitializeOptions*) PEW_EOS_Get_InitializeOptions()
```
andrew-hirata-playeveryware: I believe this function has the same issue: `get_initialize_options` returns an object that has a lifetime that's tied to platform_config and product_config, both of which only live in this scope.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/string_helpers.cpp
andrew-hirata-playeveryware: That is true, but I don't think the change is needed to get the gfx code to read the config file.

andrew-hirata-playeveryware (changes requested):

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/dllmain.cpp
paulhazen: It does not - just gives the parameter a name.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/eos_helpers.cpp
paulhazen: Cool - that makes sense. I removed the `TODO:` comment. Marking conversation as resolved.

lib/NativeCode/DynamicLibraryLoaderHelper/DynamicLibraryLoaderHelper/DynamicLibraryLoaderHelper.vcxproj
andrew-hirata-playeveryware: Okay that sounds good

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/eos_helpers.cpp
paulhazen: You are correct, this was a problem I discovered and subsequently fixed in another PR where I make use of these functions to test equivalency with managed implementation. Since these functions should only be used in testing, I corrected the implementation _and_ enclosed both their definitions and implementations in a `_DEBUG` flag to make sure they don't get included in any released binaries.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/ProductionEnvironments.hpp
andrew-hirata-playeveryware: Unless I'm mistaken, because it's a `const&` already, it should already not be mutable if captured by reference.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/eos_helpers.cpp
paulhazen: I address this problem in [your other conversation](https://github.com/PlayEveryWare/eos_plugin_for_unity/pull/1036/files/534a884be3a245cb1297a78b54077272687d57a4#r1890625922) on the same topic, so I am going to mark this conversation as resolved.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/ProductionEnvironments.hpp
paulhazen: I address this concern in [your other conversation](https://github.com/PlayEveryWare/eos_plugin_for_unity/pull/1036/files#r1884312879) on the same topic, so I'm adding a note to that conversation regarding the issue being in this place as well and am going to subsequently mark this conversation as resolved.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/ProductionEnvironments.hpp
paulhazen: Note: Similar issue was discussed WRT to the implementation of the `is_sandbox_defined` method, but that conversation was marked as resolved in order to consolidate discussion.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/Serializable.hpp
andrew-hirata-playeveryware: It's my understanding that the default value for builtin types on the stack (i.e. non-static objects) is that they are not initialized in C / C++ so it's value would be what happens to be on that part of the stack; at least as of C++11 / C11.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/dllmain.cpp
paulhazen: The previous implementation also went through it twice - this change just consolidates the looping logic into a helper function. 

The previous implementation can be found here: (https://github.com/PlayEveryWare/eos_plugin_for_unity/blob/5b1577e82f9ac5f3bc4e72a42d676230cca11dd5/lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/dllmain.cpp#L58

And is as follows:

```cpp
void get_cli_arguments(config_legacy::EOSConfig eos_config)
{
  //support sandbox and deployment id override via command line arguments
  std::stringstream argument_stream = std::stringstream(GetCommandLineA());
  std::istream_iterator<std::string> argument_stream_begin(argument_stream);
  std::istream_iterator<std::string> argument_stream_end;
  std::vector<std::string> argument_strings(argument_stream_begin, argument_stream_end);
  std::string egsArgName = "-epicsandboxid=";
  std::string sandboxArgName = "-eossandboxid=";
  for (unsigned i = 0; i < argument_strings.size(); ++i)
  {
      std::string* match = nullptr;
      if (argument_strings[i]._Starts_with(sandboxArgName))
      {
          match = &sandboxArgName;
      }
      else if (argument_strings[i]._Starts_with(egsArgName))
      {
          match = &egsArgName;
      }
      if (match != nullptr)
      {
          std::string sandboxArg = argument_strings[i].substr(match->length());
          if (!sandboxArg.empty())
          {
              logging::log_inform(("Sandbox ID override specified: " + sandboxArg).c_str());
              eos_config.sandboxID = sandboxArg;
          }
      }
  }
  
  //check if a deployment id override exists for sandbox id
  for (unsigned i = 0; i < eos_config.sandboxDeploymentOverrides.size(); ++i)
  {
      if (eos_config.sandboxID == eos_config.sandboxDeploymentOverrides[i].sandboxID)
      {
          logging::log_inform(("Sandbox Deployment ID override specified: " + eos_config.sandboxDeploymentOverrides[i].deploymentID).c_str());
          eos_config.deploymentID = eos_config.sandboxDeploymentOverrides[i].deploymentID;
      }
  }
  
  std::string deploymentArgName = "-eosdeploymentid=";
  std::string egsDeploymentArgName = "-epicdeploymentid=";
  for (unsigned i = 0; i < argument_strings.size(); ++i)
  {
      std::string* match = nullptr;
      if (argument_strings[i]._Starts_with(deploymentArgName))
      {
          match = &deploymentArgName;
      }
      else if (argument_strings[i]._Starts_with(egsDeploymentArgName))
      {
          match = &egsDeploymentArgName;
      }
      if (match != nullptr)
      {
          std::string deploymentArg = argument_strings[i].substr(match->length());
          if (!deploymentArg.empty())
          {
              logging::log_inform(("Deployment ID override specified: " + deploymentArg).c_str());
              eos_config.deploymentID = deploymentArg;
          }
      }
  }
}
```

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/eos_helpers.cpp
paulhazen: Very good observation. I've updated the implementation to use dependency injection instead of lazily leaning on use of the `static` keyword.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/eos_helpers.cpp
paulhazen: Excellent observation. I updated the algorithm to not stop when the first steam dll could not be found - restoring the original algorithm.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/ProductionEnvironments.hpp
paulhazen: Ah - good point. I have updated the implementation accordingly in both the `is_deployment_defined` and `is_sandbox_defined` functions. Feel free to mark this conversation as resolved.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/PlatformConfig.hpp
```diff
@@ -0,0 +1,303 @@
...
+     * \brief Describes the configuration options for a platform that wants to
+     * interact with the EOS SDK.
+     */
+    struct PlatformConfig : Config
+    {
+    public:
+        /**
+         * \brief Used to statically store the cache directory once it has been determined.
+         */
+        static inline std::string s_cache_directory;
```
andrew-hirata-playeveryware: Why are these static?

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/WindowsConfig.hpp
```diff
@@ -0,0 +1,99 @@
...
+                GetTempPathW(buffer_size, lpTempPathBuffer);
+
+                s_cache_directory = string_helpers::create_utf8_str_from_wide_str(lpTempPathBuffer);
+                free(lpTempPathBuffer);
+            }
+
+            return s_cache_directory.c_str();
+        }
+
+        void set_platform_specific_rtc_options() const override
```
andrew-hirata-playeveryware: I don't see why this code was moved into Windows Config: it doesn't appear to get rid of the statics, it just moves into this class.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/dllmain.cpp
andrew-hirata-playeveryware: Ah my mistake; I misread the original code.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/eos_helpers.cpp
```diff
@@ -190,223 +184,249 @@ namespace pew::eos
...
 
-#if PLATFORM_WINDOWS
-        auto path_to_steam_config_json = config_legacy::get_path_for_eos_service_config(EOS_STEAM_CONFIG_FILENAME);
+    void apply_steam_settings(EOS_Platform_Options& platform_options) 
+    {
+        // =================== START APPLY STEAM OPTIONS =======================
+
+        // This struct is used to guarantee the lifetime of the string field 
+        // members without introducing dangling pointers.
+        struct SteamSettingsContext
```
andrew-hirata-playeveryware, line 284: This is _almost_ correct: the Context object needs to be something passed into this function for it to live beyond the scope of this function.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/eos_helpers.cpp
```diff
@@ -190,223 +184,249 @@ namespace pew::eos
...
+        {
+            logging::log_error(
+                "SteamAPI_Init returned false. Steam was not able "
+                "to be initialized.");
+        }
+    }
 
-#if PLATFORM_WINDOWS
-        auto path_to_steam_config_json = config_legacy::get_path_for_eos_service_config(EOS_STEAM_CONFIG_FILENAME);
+    void apply_steam_settings(EOS_Platform_Options& platform_options)
```
andrew-hirata-playeveryware, line 278: Not for this review, but I have a style nitpick that I'd like to discuss with you about how to use references in the future.

andrew-hirata-playeveryware (changes requested):

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/eos_helpers.cpp
paulhazen: Yes - this is correct. However, I've waited to test the Steam Integration stuff for a subsequent review, and in this PR the function is never called.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/PlatformConfig.hpp
paulhazen: This issue is the same as a comment you made in another conversation - so I am marking this one as resolved and we will discuss it there.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/WindowsConfig.hpp
paulhazen: Yes - I did this in a confusing way at first. All the `statics` should be removed now.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/eos_helpers.cpp
andrew-hirata-playeveryware: Okay, I'll let this count as "fine for now".

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/eos_helpers.cpp
```diff
@@ -190,223 +184,249 @@ namespace pew::eos
...
-            platform_options.IntegratedPlatformOptionsContainerHandle = integrated_platform_options_container;
+        platform_options.TickBudgetInMilliseconds = platform_config.tick_budget_in_milliseconds;
 
-            EOS_IntegratedPlatformOptionsContainer_AddOptions addOptions = { EOS_INTEGRATEDPLATFORMOPTIONSCONTAINER_ADD_API_LATEST };
-            addOptions.Options = &steam_integrated_platform_option;
-            eos_library_helpers::EOS_IntegratedPlatformOptionsContainer_Add_ptr(integrated_platform_options_container, &addOptions);
+        if (platform_config.task_network_timeout_seconds > 0)
+        {
+            double task_network_timeout_seconds_dbl = platform_config.task_network_timeout_seconds;
+            platform_options.TaskNetworkTimeoutSeconds = &task_network_timeout_seconds_dbl;
```
andrew-hirata-playeveryware: This also is reference to a stack value that won't exist outside of this call.

andrew-hirata-playeveryware (changes requested):

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/eos_helpers.cpp
paulhazen: I fixed this to call new - and added proper `delete` calls when `platform_options` is no longer needed.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/Serializable.hpp
paulhazen: ~This is not true for C++17 in the types that the template is restricted to - but regardless - and for clarity I changed the implementation to initialize the value to `0`.~

Reminded myself by reading up on it - you are absolutely right about the initialization - but there are no code paths within the function where the variable is not assigned a value. Regardless - and in case the function's implementation changes and introduces such a path - the variable has been initialized to `0`.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/eos_helpers.cpp
andrew-hirata-playeveryware: I mean line 371 creates a pointer to a stack value, `task_network_timeout_seconds_dbl`, which won't survive outside of this `if` statement.

andrew-hirata-playeveryware (changes requested):

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/eos_helpers.cpp
paulhazen: I believe that it instantiates a new block of memory and uses a copy operator to copy the value from `platform_config.task_network_timeout_seconds`. Is that not the case?

paulhazen: Testing confirms my assumption about the persistence of the memory.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/Config/Serializable.hpp
andrew-hirata-playeveryware: The default value for int, float, and double, I believe is still uninitialized data.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/eos_helpers.cpp
andrew-hirata-playeveryware: line 370 copies the value from from `platform_config` and stores it in a double that exists on the stack for just frame. i.e. the value `&task_network_timeout_seconds_dbl`  refers to a stack address, however that is only valid till this if statement terminates. On most C++ implementations (that I've seen, at least), it will probably live till at least line `378`; however once something needs more stack space, i.e. the calling function calls another function that needs more stack space than it did previously, the value stored at that address could get overwritten.

andrew-hirata-playeveryware (changes requested):

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/eos_helpers.cpp
paulhazen: I think I see the confusion - this comment thread is showing outdated implementation - but because the comment thread was started on that version of the implementation - it still shows it for context.

The updated implementation can be seen here:

https://github.com/PlayEveryWare/eos_plugin_for_unity/blob/c7b4dd3f197df212a60cf1e28ce9b5e23ec4291e/lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/eos_helpers.cpp#L368-L372

andrew-hirata-playeveryware: Ah okay I see. I'll accept this for now but in the future we should work to change the way init is called so we don't need to allocate any memory.

andrew-hirata-playeveryware (approved):

PR #1037, opened 2024-12-02
feat/chore: Enable mirroring logging to console application.

paulhazen:
This change allows the functions in the `pew::eos::logging` namespace to have their output be mirrored to `std::cout` and enables that option in the `ConsoleApplication` project used for testing.


arthur740212 (approved):

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/include/logging.h
```diff
@@ -38,6 +38,13 @@ namespace pew::eos::logging
 {
     typedef void (*log_flush_function_t)(const char* str);
 
+    /**
+     * \brief Allows an option to be set that will mirror the log messages to
```
WispyMouse, line 42: I don't recognize this commenting syntax, but judging by below, maybe this should be `@brief`?

paulhazen: Doxygen comment formatting accepts either option. This is the format that Visual Studio will auto-insert when you type `/**` and hit enter (if the feature is turned on). Visual Studio and other IDEs will also correctly interpret these into intellisense.

andrew-hirata-playeveryware (approved):

PR #1041, opened 2024-12-04
refactor(editor): Move editor code to new upm layout directory

andrew-hirata-playeveryware:
This is a change to move editor changes to the UPM layout.

Since it has been a while, I'm going to write a little bit of a refresher on why this is being done:

The goal of doing this is to allow us to deprecate the UPM github repo, and to change our development style to more closely match what users use i.e. our plugin as an 'external' UPM package. 

Additionally, these changes in particular only move Editor code that is supposed to be shipped with the UPM, allowing tools that we write for the maintenance or generation of the UPM package to stay separate from the UPM layout.


arthur740212:
We will have to remember to update the package description.

andrew-hirata-playeveryware:
@arthur740212, currently, a wildcard is used to include everything in the UPM directory, so I believe that should suffice.

WispyMouse (approved):

paulhazen (approved):

PR #1042, opened 2024-12-04
fix: Achievements Only Refresh on Connect Login

WispyMouse:
The Achievements Service was attempting to refresh before connect logins were done. This commonly causes errors to appear relating to user account lookup failing. With this small change, we only refresh after a Connect login, which is when this'll actually be available.

#EOS-2199

### DIFF
--- a/Assets/Scripts/StandardSamples/Services/AchievementsService.cs
+++ b/Assets/Scripts/StandardSamples/Services/AchievementsService.cs
@@ -109,7 +109,11 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
 
         protected async override void OnLoggedIn(AuthenticationListener.LoginChangeKind changeType)
         {
-            await RefreshAsync();
+            // Achievements require Epic Game Services to function, which requires a Connect login
+            if (changeType == AuthenticationListener.LoginChangeKind.Connect)
+            {
+                await RefreshAsync();
+            }
         }
 
         protected override void Reset()


arthur740212 (approved):

Assets/Scripts/StandardSamples/Services/AchievementsService.cs
```diff
@@ -109,7 +109,11 @@ private static string ProductUserIdToString(ProductUserId productUserId)
 
         protected async override void OnLoggedIn(AuthenticationListener.LoginChangeKind changeType)
         {
-            await RefreshAsync();
+            // Achievements require Epic Game Services to function, which requires a Connect login
+            if (changeType == AuthenticationListener.LoginChangeKind.Connect)
+            {
+                await RefreshAsync();
+            }
```
paulhazen, line 116: The implementation of `RefreshAsync()` is as follows:

```csharp
// Check to see if authentication is required, if it's not then 
// continue. If it is, then make sure a user is authenticated before
// refreshing.
if (!RequiresAuthentication || 
    (RequiresAuthentication && AuthenticationListener.Instance.IsAuthenticated))
{
    await InternalRefreshAsync();
}
```

So it seems incorrect to make a double check on this - is there a reason that the check that's done inside the `RefreshAsync` function is insufficient? If so - we should figure out how to resolve the issue so that the conditional already within the base implementation of `RefreshAsync` resolves the way we expect it to.

If the conditional _within_ `RefreshAsync` is not behaving properly, that indicates that there is a bug within the value being set for the property `RequiresAuthentication`, or within the value being set for the property `AuthenticationListener.Instance.IsAuthenticated`.

paulhazen (changes requested):

Assets/Scripts/StandardSamples/Services/AchievementsService.cs
andrew-hirata-playeveryware: Wouldn't the refresh in the RefreshAsync also fire if _only_ a EGS aka auth login had occurred?

andrew-hirata-playeveryware (approved):

Assets/Scripts/StandardSamples/Services/AchievementsService.cs
paulhazen: The refresh in the `RefreshAsync` base implementation will only fire if an EGS aka auth login occurred _AND_ such an auth login is required for the service in question.

paulhazen: I am working on a fix that clarifies the distinct roles of Auth Interface authentication vs Connect Interface authentication - as the current implementation's limitation of having a `boolean` value that simply determines if _any_ authentication is the crux of the issue that this PR seeks to resolve. 

I've coordinated with @WispyMouse regarding how to proceed with these changes on their behalf. To be clear - the changes proposed in this PR as they currently exist are insufficient to address the problem and serve to compound an already incorrect implementation.

paulhazen:
Closing this PR per conversations and the resulting [new PR](https://github.com/PlayEveryWare/eos_plugin_for_unity/pull/1051) that addresses the issue as discussed.

PR #1043, opened 2024-12-04
fix: `UIDebugLog` enforces the maximum length of a log

WispyMouse:
If a `Text` element has a `.text` that is sufficiently large, then when trying to render it there will be a failure regarding "too many vertices", capping at 65000. One character does not necessarily map to one vertice. This change imposes a limit of 10000 printed characters. If the log goes over this limit, log messages are dequeued until there is enough room. This number was chosen after some light experimentation; I wasn't able to get a precise value that it exactly needed to be, and this seems to serve well enough.

This does recalculate the "length" of the log cache every time the log is updated. Is that problematic?

#EOS-2245


arthur740212 (approved):

Assets/Scripts/StandardSamples/UI/Common/UIDebugLog.cs
```diff
@@ -366,5 +386,33 @@ private void ScrollToBottom()
...
+                if (totalLength + nextLength > MaximumLogStringLength)
+                {
+                    numberOfMessagesToDequeueToReachThreshold++;
+                }
+
+                totalLength += nextLength;
+            }
+
+            return totalLength;
+        }
```
paulhazen: Instead of calculating the length of the log cache every time a log is added (which has a complexity of `O(n)`), we could store the total log length within a private `uint` value `_logCacheLength`. Each time a log is added, we would increment that field member with the length of the new log.

This would make the complexity of adding to the log `O(1)`. 

We would also need to make sure that when (and if) an item is removed from the queue, the `_logCacheLength` field member is decremented by the length of the log removed.

paulhazen (changes requested):

Assets/Scripts/StandardSamples/UI/Common/UIDebugLog.cs
paulhazen: I've updated the implementation per our discussions, and with permission from @WispyMouse. Marking this conversation as resolved and changing my review to approval.

paulhazen (approved):

PR #1046, opened 2024-12-06
doc(mobile,features) : add that mobile overlay isn't supported yet

arthur740212:
As of EOSSDK 1.16.4 overlay on mobile is not supported yet. That includes Social Overlay and Store Overlay


paulhazen (approved):

paulhazen (changes requested):
The table included in the `README.md` file is actually from the `eos_features.md` document. If it is to be updated in one place, then that change should be reflected in both files.

WispyMouse (approved):

andrew-hirata-playeveryware (approved):

paulhazen (approved):

PR #1051, opened 2024-12-12
fix: Implement distinction between authentication on different interfaces within EOS SDK.

paulhazen:
## Clarifying Authentication Flow in `AchievementsService`

This change addresses an issue where `AchievementsService` attempted to refresh before authentication was complete, specifically when using the Connect Interface provided by the EOS SDK.

## Key Changes:

1. **`LoginInterfaces` Enum**
    - A new `[Flags]` enum, `LoginInterfaces`, was introduced to represent combinations of authentication interfaces (Connect and Auth) defined within the EOS SDK.
3. **`AuthenticationObserver` Class**
    - An `abstract` class, `AuthenticationObserver`, was added to handle specific authentication scenarios effectively.
5. **Class Renaming for Clarity**
    - The existing `AuthenticationListener` was renamed to `AuthenticationEventDispatcher` to reduce confusion with the new `AuthenticationObserver`. The old class name was marked `[Obsolete]`.
6. **`Disposable` Class**
    - A new class, `Disposable`, was added to manage proper disposal for `AuthenticationObserver` and its derived classes.

These changes ensure clear differentiation between Connect and Auth interface events, improving the reliability of authentication-dependent services.

#EOS-2199


WispyMouse:
I opened the Achievements sample, and it didn't show any achievements until I pressed "Refresh Data". Could this suggest that it's not set up to listen to the correct events or that the correct events aren't firing? Same for PlayerDataStorage.

com.playeveryware.eos/Runtime/Core/Authentication/AuthenticationEventObserver.cs
```diff
@@ -0,0 +1,77 @@
...
+            // Subscribe to the authentication change event
+            AuthenticationEventDispatcher.Instance.AuthenticationChanged += ObserveAuthenticationChange;
+        }
+
+        private void ObserveAuthenticationChange(object sender, AuthenticationChangedEventArgs args)
+        {
+            // If the the interface for the authentication event does not 
+            // match the login interfaces that should be observed, then stop
+            // processing the event.
+            if ((ObservedLoginInterfaces & args.Interface) != ObservedLoginInterfaces)
```
WispyMouse: With the Achievement Service and Player Data Storage set to "AuthOrConnect", it appears as though this doesn't fire because "AuthOrConnect" is never equal to exactly "Auth" or "Connect". Should we adjust this bitwise logic so that this would be true in those cases?

WispyMouse:
Is it within the scope of this PR to set each of the services that exist to use the new parameter, or should that be done separately? Achievement for example should be a Connect rather than an AuthAndConnect. We could do a sweep of services and align them with their intended logins.

com.playeveryware.eos/Runtime/Core/Authentication/AuthenticationEventObserver.cs
paulhazen: I have addressed this issue by removing the `AuthOrConnect` value - and instead just having individual values for each type - and having the `AuthenticationEventObserver` store an array of acceptable interfaces to listen to.

If the fix is acceptable to you, please mark this conversation as resolved :)

com.playeveryware.eos/Runtime/Core/Authentication/AuthenticationEventObserver.cs
```diff
@@ -0,0 +1,78 @@
...
+            // Subscribe to the authentication change event
+            AuthenticationEventDispatcher.Instance.AuthenticationChanged += ObserveAuthenticationChange;
+        }
+
+        private void ObserveAuthenticationChange(object sender, AuthenticationChangedEventArgs args)
+        {
+            // If the the interface for the authentication event does not 
+            // match the login interfaces that should be observed, then stop
+            // processing the event.
+            if (ObservedLoginInterfaces.Contains(args.Interface))
```
WispyMouse: I think this has a boolean logic problem; it's preventing things that _contain_ the interface.

com.playeveryware.eos/Runtime/Core/Authentication/AuthenticationEventObserver.cs
WispyMouse: This approach is satisfactory, though please see new comment in regards to the boolean logic.

com.playeveryware.eos/Runtime/Core/Authentication/AuthenticationEventObserver.cs
paulhazen: I have updated the implementation to avoid use of bitwise operations altogether - instead the services now accept a `params` value for the constructor that allows multiple `LoginInterfaces` values to be provided to indicate what the required login stuff is.

com.playeveryware.eos/Runtime/Core/Authentication/AuthenticationEventObserver.cs
```diff
@@ -0,0 +1,78 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#if !EOS_DISABLE
+
+namespace PlayEveryWare.EpicOnlineServices
+{
+    using Common;
+    using System.Linq;
```
paulhazen: This is suspicious - does linq need to be used?

com.playeveryware.eos/Runtime/Core/Common/Disposable.cs
```diff
@@ -0,0 +1,99 @@
...
+namespace PlayEveryWare.Common
+{
+    using System;
+
+    /// <summary>
+    /// Disposable provides a base implementation for the disposal pattern,
+    /// requiring deriving classes to define behavior for disposal of both
+    /// managed and unmanaged resources.
+    /// </summary>
+    public abstract class Disposable : IDisposable
```
paulhazen, line 33: This feels like a refactor - perhaps this work would be better suited to be outside of this PR.

com.playeveryware.eos/Runtime/Core/Authentication/AuthenticationEventObserver.cs
```diff
@@ -0,0 +1,78 @@
...
+
+    public abstract class AuthenticationEventObserver : Disposable
+    {
+        protected bool NeedsAuthentication { get; private set; }
+
+        protected LoginInterfaces[] ObservedLoginInterfaces { get; }
```
paulhazen: This is a strange approach to take - we should re-evaluate use of bitwise operations on a flags enum.

com.playeveryware.eos/Runtime/Core/Common/Disposable.cs
paulhazen: Not a refactor - because existing implementations of the `IDisposable` interface were not changed to use this - instead just the newly introduced class that requires the [disposable pattern](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/dispose-pattern) uses it.

com.playeveryware.eos/Runtime/Core/Authentication/AuthenticationEventObserver.cs
paulhazen: Nope - it has been removed.

com.playeveryware.eos/Runtime/Core/Authentication/AuthenticationEventObserver.cs
paulhazen: Implementation was restored to make use of bitflags - as is appropriate.

WispyMouse (changes requested):
Setting to "Request changes" while pending fixes.

WispyMouse (approved):

andrew-hirata-playeveryware (approved):

PR #1052, opened 2024-12-12
fix: Restore functionality of the `EOS_DISABLE` scripting define.

paulhazen:
This PR resolves an issue that currently exists within `development`, where the functionality of the `EOS_DISABLE` scripting define is broken.

As is often the case with correcting this scripting define, the need to apply it balloons quickly, requiring multiple files to have the scripting define compilation guard introduced.


Assets/Plugins/Source/Editor/Build/BuildRunner.cs
```diff
@@ -20,6 +20,8 @@
  * SOFTWARE.
  */
 
+#if !EOS_DISABLE
+
 namespace PlayEveryWare.EpicOnlineServices.Editor.Build
```
paulhazen, line 25: Add tasking to address any potential issues that may arise when applied to other platforms.

Assets/Plugins/Source/Editor/EditorWindows/EOSSettingsWindow_DEPRECATED.cs
```diff
@@ -1,469 +0,0 @@
-/*
```
paulhazen: This file should not be removed - even though the scripting define conditional is weird, making it work here is preferable to including this change in this PR.

andrew-hirata-playeveryware (approved):

WispyMouse (approved):

Assets/Plugins/Source/Editor/Build/BuildRunner.cs
paulhazen: Tasking has been added - marking conversation as resolved.

Assets/Plugins/Source/Editor/EditorWindows/EOSSettingsWindow_DEPRECATED.cs
paulhazen: File has been restored - the solution was to change the multi-line string literal that was conflicting with the use of the `EOS_DISABLE` scripting define so that it's properly escaped.

File has been re-introduced with proper application of `EOS_DISABLE`, and this conversation is being marked as resolved.

Issue #975, opened 2024-10-25
`Internal_OpenConnection` within `EOSTransportManager.cs` appears to send packet before client is ready.


pewcworrell:
**Describe the bug**
When connecting a client to the server using `StartClient` with `EOSTransportManager`, the host sends back a packet within 'Internal_OpenConnection'. However, it appears the client does not properly receive this packet. I get the following warning on the client:
> 10/25/2024 15:14:04 LogEOSP2P(Warning): Received packet with unknown remote socket mapping. LocalUserId=[000...901] RemoteUserId=[000...c76] RemoteSessionRouteId=[1] Channel=[255]
After this warning the client disconnects after a few seconds.

**To Reproduce**
[PartyGameSampleWIP.zip](https://github.com/user-attachments/files/17528091/PartyGameSampleWIP.zip)
1. Download the above Unity project
2. Run 2 instances of `PartyGameLobby` scene.
3. Continue through lobby sample as you would for the regular version of the sample, connecting the two clients.
4. Click the "Play Game" button.
5. See error and disconnect after a few seconds on the client instance.

**Expected behavior**
This packet should go through without triggering a warning and disconnect on the Client.

**Screenshots**
![EOSPacketError](https://github.com/user-attachments/assets/4158cb3b-2d12-4ebb-b5e7-bdca9f00e5dd)
Screenshot shows the client (top right and attached to the editor) getting the error triggered by the host (standalone build in the bottom right).

**Desktop**
 - OS: Windows x64
 - Editor
 - Unity Version 2022.3.50f1
 - Plugin Version 3.3.3

**Additional context**
This packet is sent from an if statement that checks `connection.IsFullyOpened`. Given the context, I think that this bool may not be set correctly. If I place a `Task.Delay` within this check, this packet reliably goes through without triggering a warning on the client and the subsequent disconnect. If the issue is not the fault of EOS code, but rather something within my code, I would at least expect a more detailed error or warning from the `EOSTransportManager` of what is set up and/or called incorrectly.

PR #1058, opened 2024-12-14
fix: Properly wait for client to accept connection

WispyMouse:
Before: When a host requests a client to start a connection, clients were using `AcceptConnection` as if it were synchronous. In the event that a host opens a connection sufficiently faster than a client, the client may be messaged with the "READY" before it has actually accepted the remote connection. The result is that the client fails to receive the packet, does not respond to the host's connection attempt, and thus the connection cannot be made.

Now: Subscribing to the `AddNotifyPeerConnectionEstablished` message and handling the results in there allows the client to properly time when they're ready for communication.

`Internal_OpenConnection` has been split into synchronous part (`Internal_StartOpenIncomingConnection`) and asynchronous part (`Internal_ContinueOpenOutgoingConnection`).

> [!WARNING]
> This has barely been tested at the point where this is put up to PR. This pr is in draft to start with until it has been tested. Please give feedback on the approach and naming conventions. Please provide scenarios you want to see covered in tests, if you can think of anything in particular.
> The testing process for this will include:
> - Recreating #975 's code on top of this to reproduce the original issue and ensure it is fixed.
> - Testing the P2P samples.
> - Creating a multiple-client scenario, in addition to one:one client communications.

#EOS-2255

### DIFF
--- a/Assets/Scripts/P2PNetcodeSample/Networking/EOSTransportManager.cs
+++ b/Assets/Scripts/P2PNetcodeSample/Networking/EOSTransportManager.cs
@@ -117,6 +117,12 @@ namespace PlayEveryWare.EpicOnlineServices.Samples.Network
             public bool IsValid = false;
 
             /// <summary>
+            /// A callback that runs after this particular connection has become fully opened properly.
+            /// Alternatively, listeners can subscribe to <see cref="EOSTransportManager.OnConnectionOpenedCb"/>.
+            /// </summary>
+            public OnConnectionOpenedCallback AfterConnectionCallback = null;
+
+            /// <summary>
             /// Creates a Connection with no initial information.
             /// </summary>
             public Connection() { }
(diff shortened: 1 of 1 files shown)


Assets/Scripts/P2PNetcodeSample/Networking/EOSTransportManager.cs
```diff
@@ -1279,6 +1289,7 @@ public bool TryReceivePacket(out ProductUserId remoteUserId, out string socketNa
         //
 
         private ulong ConnectionRequestNotificationsId = 0;
+        private ulong ConnectionEstablishedNotificationsId = 0;
```
paulhazen: Field members should always be placed above any functions in a class - just as by convention.

paulhazen: They should also have comments that sufficiently explain what they are.

I understand that in _this PR_ you just renamed the field member - but I think it is appropriate in this circumstance to take the opportunity to move the field member to the correct place and add documentation - even if that is every so slightly beyond the scope of the PR.

Assets/Scripts/P2PNetcodeSample/Networking/EOSTransportManager.cs
```diff
@@ -1349,6 +1396,59 @@ private void OnConnectionClosedNotification(ref OnRemoteConnectionClosedInfo dat
...
+            bool shouldCreateIncomingConnection = false;
+            Connection connection = null;
+
+            if (!Connections.TryGetValue(remoteUserId, out List<Connection> foundConnections))
+            {
+                shouldCreateIncomingConnection = true;
+            }
+            else
+            {
+                connection = foundConnections.Find(x => x.SocketName == socketName);
```
WispyMouse, line 1419: Note to self: Determine without a doubt that this is not a `using System.Linq` related function call.

WispyMouse:
Retargeted to `release-3.3.6`. This is now ready for review and approval.

arthur740212 (approved):

WispyMouse:
This was intentionally changed to target development, as it is not going in to the 3.3.6 release.

andrew-hirata-playeveryware (approved):

Assets/Scripts/P2PNetcodeSample/Networking/EOSTransportManager.cs
WispyMouse: Confirmed; this is from `System.Collections.Generic.List`. `Find` is a part of the List definition.

PR #1059, opened 2024-12-16
test: Native Render

paulhazen:
The following PR adds unit tests that confirm the equivalency of config values as read from the native and managed components of the plugin.

> ~[!IMPORTANT]~
> ~This PR has been moved to draft pending an internal discussion regarding the best testing framework to utilize.~

> [!NOTE]
> Discussion was had internally, and it was agreed that having tests in _some_ framework was preferable to having no tests - and that changing testing frameworks would be something trivial to do at a later date if it was deemed necessary.

#EOS-1982


lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/eos_helpers.cpp
```diff
@@ -405,39 +405,29 @@ namespace pew::eos
         }
 
         apply_rtc_options(platform_options);
-        apply_steam_settings(platform_options);
+        // TODO: Re-enable apply_steam_settings(platform_options);
```
WispyMouse: Is it necessary to have this commented out to unblock this pr?

lib/NativeCode/DynamicLibraryLoaderHelper/ManagedPluginCode/Application.cs
```diff
@@ -33,7 +33,7 @@ internal static class Application
         /// This is the path of the streaming assets directory relative to the
         /// output directory of this class library.
         /// </summary>
-        public static readonly string streamingAssetsPath = @"..\..\..\..\Assets\StreamingAssets\";
+        public static readonly string streamingAssetsPath = @"..\..\..\..\..\..\Assets\StreamingAssets\";
```
WispyMouse, line 36: Nervousness about relative paths intensify. Why did this change?

WispyMouse (approved):

lib/NativeCode/DynamicLibraryLoaderHelper/ManagedPluginCode/Application.cs
paulhazen: This changed because the `ManagedPluginCode` project is _only ever_ deployed to
```
$(SolutionDir)Build\$(Configuration)\$(Platform)
``` 
for testing purposes. And the relative path to the `Assets\StreamingAssets\` directory from _that_ directory which typically evaluates to:
```
[PathToRepo]\lib\NativeCode\DynamicLibraryLoaderHelper\Build\Debug\x64
```
is 
```
..\..\..\..\..\..\Assets\StreamingAssets
```

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/eos_helpers.cpp
paulhazen: Not certain what you mean by to unblock this pr - but the functionality was turned off because it (the steam integration) wasn't in a test-able state. This will be resolved in a subsequent PR.

arthur740212:
> Not so comfortable that the solution requires the test to compile.

The solution only requires the tests to compile if in `Debug` mode - otherwise the `Test` project is not required.

> I feel like a test like this is better in a separate solution.

It's standard practice to put test projects in the same solution as the projects that they are designed to test.

> Also there is an extra .net 4.8 dependency introduced somewhere that I'm not quite sure why.

That dependency is only for `ManagedPluginCode`, which is an _existing dependency_ not introduced by this PR.

arthur740212 (approved):

andrew-hirata-playeveryware (changes requested):
I believe we need to discuss testing frameworks.

paulhazen:
> I believe we need to discuss testing frameworks.

We've now had the discussion internally - and I believe that the consensus was to include these tests as it would be easy to change unit testing frameworks later, and would be good to have tests in _some_ framework rather than no tests at all.

andrew-hirata-playeveryware (approved):

PR #1063, opened 2024-12-17
chore: Organize files within `DynamicLibraryLoaderHelper` for Windows.

paulhazen:
At some point in the history of the project - different `DynamicLibraryLoaderHelper` solutions were introduced - one for each platform for which it is required. Source and header files for those projects were appropriately placed into subdirectories - but the files specific to Windows were never put into the subdirectory for the windows version of the project.

This PR accomplishes that move without changing any implementation. 

The result is a clearer organization for the files within `lib/NativeCode/DynamicLibraryLoaderHelper`

Untitled


andrew-hirata-playeveryware (changes requested):
I specifically want the files to be in path external to the directory that has the solutions.

WispyMouse (approved):

paulhazen:
> I specifically want the files to be in path external to the directory that has the solutions.

That's not consistent with all the other projects - including NativeRender:

The current layout for NativeRender is as follows:
`lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/*`

and the solution file is here:
`lib/NativeCode/DynamicLibraryLoaderHelper/DynamicLibraryLoaderHelper.sln`

The same is true of all the other projects within the solution.

PR #1066, opened 2024-12-18
fix: If `definition.StatThresholds` is null, don't iterate on it for selectedDescription

WispyMouse:
Not all achievements have StatThresholds. This fixes an issue with the achievements sample where a null reference exception is thrown on trying to display an achievement with a null statthresholds.

#EOS-2299

### DIFF
--- a/Assets/Scripts/StandardSamples/UI/Achievements/UIAchievementsMenu.cs
+++ b/Assets/Scripts/StandardSamples/UI/Achievements/UIAchievementsMenu.cs
@@ -292,9 +292,12 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
                 "Id: {0}\nUnlocked Display Name: {1}\nUnlocked Description: {2}\nLocked Display Name: {3}\nLocked Description: {4}\nHidden: {5}\n",
                 definition.AchievementId, definition.UnlockedDisplayName, definition.UnlockedDescription, definition.LockedDisplayName, definition.LockedDescription, definition.IsHidden);
 
-            foreach (StatThresholds st in definition.StatThresholds)
+            if (definition.StatThresholds != null)
             {
-                selectedDescription += string.Format("Stat Thresholds: '{0}': {1}\n", st.Name, st.Threshold);
+                foreach (StatThresholds st in definition.StatThresholds)
+                {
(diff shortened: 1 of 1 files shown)


Assets/Scripts/StandardSamples/UI/Achievements/UIAchievementsMenu.cs
```diff
@@ -292,9 +292,12 @@ void DisplayAchievementDefinition(DefinitionV2 definition)
                 "Id: {0}\nUnlocked Display Name: {1}\nUnlocked Description: {2}\nLocked Display Name: {3}\nLocked Description: {4}\nHidden: {5}\n",
                 definition.AchievementId, definition.UnlockedDisplayName, definition.UnlockedDescription, definition.LockedDisplayName, definition.LockedDescription, definition.IsHidden);
 
-            foreach (StatThresholds st in definition.StatThresholds)
+            if (definition.StatThresholds != null)
```
paulhazen, line 295: Nice catch. Not a suggestion to change - but food for thought is that you can do this in one line by utilizing the null-coalescing operator `??` to provide an empty collection if the value being iterated on is null:

```csharp
foreach (StatThresholds st in definition.StatsThresholds ?? Enumerable.Empty<StatThresholds>)
{
    /* ... */
}
```

But I think your approach in this case is more straightforward and (more importantly) easier to understand.

Feel free to mark this conversation as resolved - I have approved the review but wanted you to see my input.

paulhazen (approved):
Nice catch! Approving because this resolves the issue identified - but I wonder whether it would also be wise to add a check just inside the function definition to check to make sure that the provided `definition` itself is not `null`. Unless I'm mistaken - if a `null` parameter is provided an uncaught exception will occur.

Assets/Scripts/StandardSamples/UI/Achievements/UIAchievementsMenu.cs
WispyMouse: Thank you for sharing! I didn't consider that you could do something like this after the "in" statement as part of a foreach.

A null definition would have other issues here, yea. I _think_ this is the only nullable field inside it.

arthur740212 (approved):

andrew-hirata-playeveryware (approved):

PR #1069, opened 2025-01-02
fix: Parse Enums as Integers for Comparison in `ListOfStringsToEnumConverters.FromNumberValue`

WispyMouse:
`ListOfStringsToEnumConverters.FromNumberValue` converts a string to an enum, and then compares that enum to find the "lowest" value. If the enum fails to parse, it will set itself properly to the lowest value. This was trying to compare an Enum to an integer, and was previously failing when the integer was 0 but the Enum didn't have a "0" defined. This fix explicitly compares the underlying value instead of comparing the enum, which covers this scenario.

#EOS-2337


andrew-hirata-playeveryware (approved):

paulhazen (approved):

paulhazen (changes requested):
PR needs to merge into `development` instead of `fix/native-use-new-config`

paulhazen (approved):

PR #1070, opened 2025-01-02
Fix/nullcheck migration

WispyMouse:
While migrating, do some nullchecking to ensure that values are present before trying to import them.
`platformConfig.threadAffinity` can be null, so do a null check before unwrapping.

#EOS-2338


WispyMouse:
Force pushed this a couple of times while fixing a silly persistent binary error. Shouldn't cause any problems with merging.

paulhazen (changes requested):
PR should be merging into `development` instead of `fix/native-use-new-config`

paulhazen (approved):

andrew-hirata-playeveryware (approved):

PR #1071, opened 2025-01-02
Fix: `RenderSetOfNamed` can render empty set

WispyMouse:
Before: `RenderSetOfNamed` required at least one entry, or it wouldn't render.
Now: If the list is empty, only the + to add things will be displayed.

#EOS-2339


paulhazen (changes requested):
PR needs to merge into `development` instead of `fix/native-use-new-config`.

paulhazen (approved):

andrew-hirata-playeveryware (approved):

PR #1075, opened 2025-01-07
fix:  Add config editors for all platforms.

paulhazen:
This PR enables all platforms to have configuration editors available.

#EOS-2342


WispyMouse:
Internally discussed that we might want a different approach, since users without all platforms won't be able to compile in this way. Will review next approach.

paulhazen:
Closing this PR in favor of [this PR](https://github.com/PlayEveryWare/eos_plugin_for_unity/pull/1079).

PR #1079, opened 2025-01-08
fix: Add config editors for all platforms

paulhazen:
This PR enables all platforms to have configuration editors available.

#EOS-2342


com.playeveryware.eos/Runtime/Core/Common/Utility/ReflectionUtility.cs
```diff
@@ -20,15 +20,35 @@
  * SOFTWARE.
  */
 
-namespace PlayEveryWare.EpicOnlineServices.Editor.Utility
+namespace PlayEveryWare.Common.Utility
 {
     using System.Collections.Generic;
     using System;
     using System.Linq;
     using System.Reflection;
```
WispyMouse, line 29: I'm wary of any of our tools relying on Reflection and Linq. It looks like this was moved from the Editor folder. What kind of discussion can we have to either decide against this implementation detail, or assuaging my fears of things? I get that this is only meant to run at the start of the program, but it's concerning.

paulhazen: Couple of things. First of all - the using Linq and using Reflection lines were not introduced in this PR, so if they are to change they should be changed in a different PR.

Secondly - you will notice that the `using PlayEveryWare.Common.Utility;` statement, and all references to functions within the class are surrounded by compile time definitions of `#if UNITY_EDITOR`.

paulhazen: Instead of wrapping the places that the class is used with `#if UNITY_EDITOR` conditionals - it should probably be done _inside_ the `ReflectionUtility.cs` file. However, this would be out of scope for the PR.

WispyMouse: It's true that this PR doesn't add Reflection or Linq to this file, _but_ this was moved from an Editor-only thing to being included in runtime now. This does appear to only be called in one place at the moment, the `InitializePlatformConfigs`, which does wrap it around in `#if UNITY_EDITOR`.

If this isn't a significant point of concern for you/other engineers on the team, then this can go through as is. I'd feel better if we did go with the strategy of wrapping this class in the `#if UNITY_EDITOR`, such that Reflection and Linq aren't brought in as namespaces in those environments... but I also don't know if there's any inherent harm in including the namespace.

paulhazen: You make a very good point about it being moved from the editor directory. In light of that, I will add compile conditionals around the class definition to guarantee things.

As for "including the namespace," I believe that during compilation only namespaces whose members are utilized are actually included.

paulhazen: I have just pushed a change that encloses the class definition for `ReflectionUtility` in a compile conditional.

WispyMouse (approved):

arthur740212 (approved):

PR #1080, opened 2025-01-09
feat: Enable PlatformConfigs to Provide Default `IntegratedPlatformManagementFlags`

WispyMouse:
While migrating, if the old config had no IntegratedPlatformManagementFlags, or the previous value was `Disabled`, try to set the value to a provideable set of default flags. Otherwise, use the old config's IPMFs. This set up currently always provides `Disabled` as the default, but it enables flexibility in platform management of default flags.

#EOS-2344

### DIFF
--- a/com.playeveryware.eos/Runtime/Core/Config/PlatformConfig.cs
+++ b/com.playeveryware.eos/Runtime/Core/Config/PlatformConfig.cs
@@ -412,8 +412,15 @@ namespace PlayEveryWare.EpicOnlineServices
 
             MigratePlatformFlags(overrideValuesFromFieldMember, mainNonOverrideableConfig);
 
-            integratedPlatformManagementFlags = IntegratedPlatformManagementFlags.Disabled;
-            integratedPlatformManagementFlags |= mainNonOverrideableConfig.integratedPlatformManagementFlags;
+            // If there are no Integrated Platform Management Flags in the original config, apply a set of default per-platform IMPFs
+            if ((int)mainNonOverrideableConfig.integratedPlatformManagementFlags == 0 || mainNonOverrideableConfig.integratedPlatformManagementFlags == IntegratedPlatformManagementFlags.Disabled)
+            {
+                integratedPlatformManagementFlags = GetDefaultIntegratedPlatformManagementFlags();
+            }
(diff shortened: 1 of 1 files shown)


paulhazen (approved):
Solid choice. My only suggested change is to not use the fully qualified name for the `GetDefaultIntegratedPlatformManagementFlags()` return value.

WispyMouse:
> Solid choice. My only suggested change is to not use the fully qualified name for the `GetDefaultIntegratedPlatformManagementFlags()` return value.

Do you have a preferred alternative name we should consider?

paulhazen:
> > Solid choice. My only suggested change is to not use the fully qualified name for the `GetDefaultIntegratedPlatformManagementFlags()` return value.
> 
> Do you have a preferred alternative name we should consider?

I think you misunderstand - the _name_ of the function I think is fine - sure it's long but it's descriptive and anything shorter could be misleading. I meant that the return value is written as 
```csharp
Epic.OnlineServices.IntegratedPlatform.IntegratedPlatformManagementFlags
```
Instead of having
```csharp
using Epic.OnlineServices;
```

at the top, and just having
```csharp
IntegratedPlatformManagementFlags
```

written as the return type.

andrew-hirata-playeveryware (approved):

arthur740212 (approved):

WispyMouse:
> > > Solid choice. My only suggested change is to not use the fully qualified name for the `GetDefaultIntegratedPlatformManagementFlags()` return value.
> > 
> > 
> > Do you have a preferred alternative name we should consider?
> 
> I think you misunderstand - the _name_ of the function I think is fine - sure it's long but it's descriptive and anything shorter could be misleading. I meant that the return value is written as
> 
> ```cs
> Epic.OnlineServices.IntegratedPlatform.IntegratedPlatformManagementFlags
> ```
> 
> Instead of having
> 
> ```cs
> using Epic.OnlineServices;
> ```
> 
> at the top, and just having
> 
> ```cs
> IntegratedPlatformManagementFlags
> ```
> 
> written as the return type.

Oh now I see, you're entirely right. I'm going to use the namespace as you suggested, and then merge this in.

PR #1081, opened 2025-01-10
fix: Restore Steam Integration

paulhazen:
This PR restores the functionality of steam integration at the native layer - specifically for Windows Standalone.

#EOS-2265


lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/eos_helpers.cpp
```diff
@@ -250,19 +250,19 @@ namespace pew::eos
 
         // Get a pointer to the SteamAPI_Init function within the steam
         // library.
-        const auto SteamAPI_Init = pew::eos::eos_library_helpers::load_function_with_name<SteamAPI_Init_t>(steam_dll_handle, "SteamAPI_Init");
+        const auto SteamAPI_Init_Fn = pew::eos::eos_library_helpers::load_function_with_name<SteamAPI_Init_t>(steam_dll_handle, "SteamAPI_InitSafe");
```
paulhazen: It was curious to me that this needed to be changed from `SteamAPI_Init`, because Steamworks documentation still indicates that `SteamAPI_Init` should be called, however, that function does not exist in the version of `steam_api64.dll` that is included with the steamworks plugin.

lib/NativeCode/DynamicLibraryLoaderHelper/Directory.Build.props
```diff
@@ -15,12 +15,19 @@
...
+    <STEAM_API_DLL_NAME Condition="'$(NormalizedPlatform)' == 'x86'">steam_api.dll</STEAM_API_DLL_NAME>
+
+      <!-- 
       Define a property with the SolutionDir path but with backslashes replaced 
       by forward slashes 
       -->
     <SolutionDirForwardSlashes>$(SolutionDir.Replace('\', '/'))</SolutionDirForwardSlashes>
     <UnityStreamingAssetsDirectory>$(SolutionDirForwardSlashes)../../../Assets/StreamingAssets/</UnityStreamingAssetsDirectory>
-    <OutputUnityAssetsDirectory>$(SolutionDir)..\..\..\Assets\Plugins\Windows\$(NormalizedPlatform)\</OutputUnityAssetsDirectory>
+    <SteamLibraryDirectory>$(SolutionDirForwardSlashes)../../../Library/PackageCache/com.rlabrecque.steamworks.net*/Plugins/</SteamLibraryDirectory>
```
WispyMouse, line 30: It seems strange to reach into the /Library folder for things. If I have a clean build of the repo, and go to Native code first, will this still correctly build? Or is this the output directory?

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/NativeRender.vcxproj
```diff
@@ -81,23 +81,16 @@
       <PreprocessorDefinitions>PEW_EOS_EXPORT;OVERLAY_DLL_NAME="$(OverlayDllName)";STEAM_API_DLL="$(SteamApiDll)";SDK_DLL_NAME="$(SdkDllName)";CONFIG_DIRECTORY="$(ConfigDirectory)";_DEBUG;NATIVERENDER_EXPORTS;_WINDOWS;_USRDLL;%(PreprocessorDefinitions)</PreprocessorDefinitions>
     </ClCompile>
     <PreBuildEvent>
-      <!-- When debugging, make sure to copy the EOSSDK Dll to the output directory -->
-      <Command>
-        @echo Copying "$(OutputUnityAssetsDirectory)$(SdkDllName)" to "$(TargetDir)"
-        copy "$(OutputUnityAssetsDirectory)$(SdkDllName)" "$(TargetDir)" /Y
-      </Command>
+      <!-- Execute the prebuild powershell command -->
+      <Command>powershell -NoProfile -ExecutionPolicy Bypass -File "$(ProjectDir)scripts\prebuild.ps1" -Configuration "$(Configuration)" -OutputUnityAssetsDirectory "$(OutputUnityAssetsDirectory)" -EOS_SDK_DLL_NAME "$(EOS_SDK_DLL_NAME)" -STEAM_API_DLL_NAME "$(STEAM_API_DLL_NAME)" -TargetDir $(TargetDir) -SteamLibraryDirectory "$(SteamLibraryDirectory)" -SolutionDir "$(SolutionDir)"</Command>
```
WispyMouse, line 85: I feel like this will fail to run on some computers... but probably not our own! Powershell often has to be installed or agreed to, in order to work, but I'm willing to deal with that for our own native libs.

WispyMouse: If Steam is missing, will this still compile?

lib/NativeCode/DynamicLibraryLoaderHelper/Directory.Build.props
paulhazen: This is only used in debug mode. What's happening is that the console application that can be used to run all the native code without unity needs access to the `steam_api64.dll` (or `steam_api.dll`) file. 

When making a build from within Unity - the appropriate file is placed in the built output directory. When making a test build that just checks to see if the code that hooks into the dll is working - it needs to know where that file is.

In summary - it is strange - and if you're building a debug build of the native code from a fresh repository it will still succeed - but will not be able to execute the component relating to steam integration. It will behave as if there is no steam integration - which is actually how we have it in our project, so it would behave as expected.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/NativeRender.vcxproj
paulhazen: > I feel like this will fail to run on some computers... but probably not our own! Powershell often has to be installed or agreed to, in order to work, but I'm willing to deal with that for our own native libs.

PowerShell comes with Windows by default since Windows 8 or perhaps even earlier. Specifying the `Bypass` value for the `ExecutionPolicy` flag should make this work properly.

> If Steam is missing, will this still compile?

Good question - I think it might fail because it cannot copy the steam dll if it does not exist. I'll go fix that.

paulhazen: I have updated the implementation to warn that the steam dll could not be copied if it doesn't exist - which will make the build succeed on systems that do not have steamworks setup.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/eos_helpers.cpp
paulhazen: We should add code that _falls back on_ the previous function name if the one does not exist.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/string_helpers.cpp
```diff
@@ -151,4 +151,22 @@ namespace pew::eos::string_helpers
     {
         return to_utf8_str(path.native());
     }
+
+    bool is_empty_or_whitespace(const char* str)
```
paulhazen, line 155: Does this work for different text encodings?

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/eos_helpers.cpp
paulhazen: Such a fall back has now been implemented - marking this conversation as resolved.

lib/NativeCode/DynamicLibraryLoaderHelper/NativeRender/src/string_helpers.cpp
paulhazen: It works appropriately for all encodings that are going to be encountered. If the string is a wide string this would fail - but that can't really happen because the parameter type isn't `wchar`. In summary: this function works in all circumstances for our purposes.

Marking this conversation as resolved.

WispyMouse (approved):
Steam integration appears to work. Able to login with External Auth without an issue.

andrew-hirata-playeveryware (approved):

PR #1082, opened 2025-01-13
feat: ExecuteQueuedMainThreadTasks

WispyMouse:
Some Actions need to be executed on the main thread. This brings over a change from another branch to allow for the enqueing of such tasks, and subsequent execution on main thread.

#EOS-2350

### DIFF
--- a/com.playeveryware.eos/Runtime/Core/EOSManager.cs
+++ b/com.playeveryware.eos/Runtime/Core/EOSManager.cs
@@ -184,6 +184,17 @@ namespace PlayEveryWare.EpicOnlineServices
         static private bool s_isConstrained = true;
         static public bool ApplicationIsConstrained { get => s_isConstrained; }
 
+        /// <summary>
+        /// Actions that need to be executed on the main thread.
+        /// Lazy allocated in <see cref="DispatchAsync"/>.
+        /// </summary>
+        static private List<Action> s_enqueuedTasks;
+
+        /// <summary>
+        /// Locak object used for <see cref="s_enqueuedTasks"/>, such that it can
+        /// be executed thread-safe way.
+        /// </summary>
+        static private System.Object s_enqueuedTasksLock = new System.Object();
         //private static List
 
         //-------------------------------------------------------------------------
@@ -1553,6 +1564,7 @@ namespace PlayEveryWare.EpicOnlineServices
             //-------------------------------------------------------------------------
             public void Tick()
             {
+                ExecuteQueuedMainThreadTasks();
                 if (GetEOSPlatformInterface() != null)
                 {
                     // Poll for any application constrained state change that didn't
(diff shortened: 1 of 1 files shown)


com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -184,6 +184,17 @@ enum EOSState
         static private bool s_isConstrained = true;
         static public bool ApplicationIsConstrained { get => s_isConstrained; }
 
+        /// <summary>
+        /// Actions that need to be executed on the main thread.
+        /// Lazy allocated in <see cref="DispatchAsync"/>.
+        /// </summary>
+        static private List<Action> s_enqueuedTasks;
```
paulhazen: Nit: typical convention is to have access modifiers precede other modifiers:
```csharp
private static List<Action> s_enqueuedTasks;
```

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -1891,5 +1903,46 @@ void IEOSCoroutineOwner.StartCoroutine(IEnumerator routine)
...
+                {
+                    s_enqueuedTasks = new List<Action>();
+                }
+                s_enqueuedTasks.Add(action);
+            }
+        }
+
+        static private void ExecuteQueuedMainThreadTasks()
+        {
+            // Lock the enqued tasks list, and hold refrence to the enqueued tasks.
```
paulhazen: nit: typo "refrence" => "reference"

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -1891,5 +1903,46 @@ void IEOSCoroutineOwner.StartCoroutine(IEnumerator routine)
...
+                if (actionsToRun == null)
+                {
+                    return;
+                }
+            }
+
+            foreach (Action action in actionsToRun)
+            {
+                action.Invoke();
+            }
```
paulhazen, line 1995: I _could_ be wrong about this - but don't we want the lock to enclose the invocation of the actions as well?

paulhazen (approved):
Provisionally approving - pending the resolution of opened conversations.

com.playeveryware.eos/Runtime/Core/EOSManager.cs
WispyMouse: This was copied from elsewhere, but looking at it... Yea I can't see why the lock is outside of the actual execution.

WispyMouse: Ah, I see now. This is outside of it in case anything wants to run `DispatchAsync` during something that is executing in `ExecuteQueuedMainThreadTasks`. If there was a lock around the whole thing, then there'd be a deadlock when executing the tasks. Instead;

- Create `actionsToRun`
- Lock on the `s_enqueuedTasksLock`
- Set `actionsToRun` to the current `s_enqueuedTasks`
- End lock
- If there was anything in `actionsToRun`, then execute them

I think this is the correct implementation, as-is.

andrew-hirata-playeveryware (approved):

WispyMouse:
Converting this to draft to signal how it isn't making it in to the next release.

WispyMouse:
Retargeted to `release-3.3.6`. Ready for review and approval.

WispyMouse:
Which this already has! So I'm going to merge it in.

PR #1085, opened 2025-01-16
fix: Set default deployment and sandbox values for platforms for first-time user scenarios

paulhazen:
This PR addresses an issue discovered during the testing of the upcoming release - wherein first-time users would have difficulty properly setting the Deployment information for the platform of their choice.

With this change - when a deployment is set for the first time (by defining a DeploymentId and a SandboxId in EOS Plugin -> EOS Configuration) - that deployment will be selected for each of the platform configurations.

Prior to this change it was possible to get into a state where the deployment set for the platform configuration would be incompletely defined. It was not initially caught because emphasis was placed on testing the most common use scenario - where a user already had deployment values defined. Testing for the release revealed the gap in testing that occluded the first-time user scenario. Such testing is now complete, and this PR addresses the identified shortcomings.

#EOS-2355


com.playeveryware.eos/Runtime/Core/Config/ProductConfig.cs
```diff
@@ -114,6 +141,86 @@ protected override bool NeedsMigration()
...
+            // If there is one deployment, and one sandbox, then make sure they
+            // are linked to each other.
+            // But only do this if they have been newly added
+            if (!_deploymentDefined && Environments.Deployments.Count ==1 && Environments.Sandboxes.Count == 1)
+            {
+                Environments.Deployments[0].Value.SandboxId = Environments.Sandboxes[0].Value;
+            }
+        }
+
+        protected override void OnWriteCompleted()
```
paulhazen, line 168: Within this PR, this is the meat of the change. All other changes are in service of accomplishing what this function accomplishes. It would be sensible to review the PR starting by reading this function.

Assets/Plugins/Source/Editor/EditorWindows/EOSSettingsWindow.cs
```diff
@@ -178,6 +216,9 @@ private void Save()
             // Save the product config editor
             _productConfigEditor.Save();
 
+            // reload the product config editor
+            await _productConfigEditor.LoadAsync();
```
WispyMouse: This is an `await` inside a non async method.

com.playeveryware.eos/Runtime/Core/Config/ProductConfig.cs
```diff
@@ -114,6 +141,86 @@ protected override bool NeedsMigration()
...
+#if UNITY_EDITOR
+
+        protected override void BeforeWrite()
+        {
+            // If there is one deployment, and one sandbox, then make sure they
+            // are linked to each other.
+            // But only do this if they have been newly added
+            if (!_deploymentDefined && Environments.Deployments.Count ==1 && Environments.Sandboxes.Count == 1)
+            {
+                Environments.Deployments[0].Value.SandboxId = Environments.Sandboxes[0].Value;
```
WispyMouse, line 164: Should this only assign if `Environments.Sandboxes[0]` is not Empty?

com.playeveryware.eos/Runtime/Core/Config/ProductConfig.cs
```diff
@@ -114,6 +141,86 @@ protected override bool NeedsMigration()
...
+            // or there is not now one defined - then there is no need to try
+            // and set the deployment values for each platform config.
+            if (_deploymentDefined || !Environments.IsDeploymentDefined)
+            {
+                return;
+            }
+
+            // Select the first defined deployment as the deployment to set
+            // platform configs to use.
+            Named<Deployment> deploymentToSetPlatformsTo = Environments.Deployments[0];
```
paulhazen: This should be fixed so that it is not assumed that the correctly defined deployment is the first one. Perhaps have a "GetFirstDefinedDeployment" function?

com.playeveryware.eos/Runtime/Core/Config/SandboxId.cs
```diff
@@ -70,6 +71,22 @@ readonly get
...
+        /// value is equal to the string value for an empty Guid with the dashes
+        /// removed.
+        /// </summary>
+        [JsonIgnore]
+        public readonly bool IsEmpty
+        {
+            get
+            {
+                return String.IsNullOrEmpty(_value) || 
+                       Guid.Empty.ToString().Replace("-", "").Equals(_value);
```
paulhazen: Instead of doing this - we should utilize the `.ToString("N")` function - since that does the appropriate stripping.

paulhazen: Also look into whether this property can or should be utilized prior to applying the value to start up parameters of the EOS SDK.

com.playeveryware.eos/Runtime/Core/Config/ProductConfig.cs
paulhazen: Yep - good catch. I've updated the implementation. Marking this conversation as resolved accordingly.

com.playeveryware.eos/Runtime/Core/Config/SandboxId.cs
paulhazen: Property was updated to correctly use the `ToString("N")` function, and (since doing so would be outside the scope of the PR) a comment was added to `IsEmpty` suggesting that the function be utilized prior to being used to initialize the EOS SDK.

Marking conversation as resolved accordingly.

com.playeveryware.eos/Runtime/Core/Config/ProductConfig.cs
paulhazen: Implementation was updated to use a newly defined function called `TryGetFirstNamedDeployment`.

Accordingly, I am going to mark this conversation as resolved.

Assets/Plugins/Source/Editor/EditorWindows/EOSSettingsWindow.cs
paulhazen: I resolved this issue - marking conversation as resolved.

Assets/Plugins/Source/Editor/EditorWindows/EOSSettingsWindow.cs
```diff
@@ -92,10 +93,47 @@ public static void ShowWindow()
...
+        protected override void OnDestroy()
+        {
+            ProductConfig.PlatformConfigsDeploymentUpdatedEvent -= ReloadDeploymentSettingsForPlatformConfigEditors;
+            base.OnDestroy();
+        }
+
+        private void ReloadDeploymentSettingsForPlatformConfigEditors(object sender, ProductConfig.PlatformConfigsDeploymentUpdatedEventArgs e)
+        {
+            // For each of the platform config editors
+            foreach (IPlatformConfigEditor platformConfigEditor in _platformConfigEditors)
```
WispyMouse, line 111: Is it ever possible for `_platformConfigEditors` to become added to as a result of running this method? I'm guessing by this point, all platform config editors are already subscribed.

com.playeveryware.eos/Runtime/Core/Config/ProductConfig.cs
```diff
@@ -114,6 +141,86 @@ protected override bool NeedsMigration()
...
+                if (config.deployment.IsComplete)
+                {
+                    continue;
+                }
+
+                // Add to the list of platform configs that have been updated
+                platformConfigsUpdated.Add(platform);
+
+                // Set the deployment.
+                config.deployment = deploymentToSetPlatformsTo.Value;
```
WispyMouse, line 205: This fails if `deploymentToSetPlatformsTo` is null.

com.playeveryware.eos/Runtime/Core/Config/ProductionEnvironments.cs
```diff
@@ -50,6 +50,33 @@ public ProductionEnvironments()
...
+            {
+                // If the deployment is complete then stop
+                if (dep.Value.IsComplete)
+                {
+                    deployment = dep;
+                    break;
+                }
+            }
+
+            return (deployment == null);
```
WispyMouse: This has a boolean logic error; I think it should return true if deployment is not null.

com.playeveryware.eos/Runtime/Core/Config/ProductConfig.cs
paulhazen: If `deploymentToSetPlaytformsTo` is not defined - then the function will not get to that point.

Assets/Plugins/Source/Editor/EditorWindows/EOSSettingsWindow.cs
paulhazen: No - it is not possible for `_platformConfigEditors` to be added as a result of this function that particular field member is never altered - it is only enumerated.

com.playeveryware.eos/Runtime/Core/Config/ProductionEnvironments.cs
paulhazen: Good catch - I've corrected it.

andrew-hirata-playeveryware (approved):

WispyMouse (approved):

com.playeveryware.eos/Runtime/Core/Config/ProductConfig.cs
WispyMouse: `deploymentToSetPlatformsTo` was null due to a boolean logic error; that has been fixed in another discussion.

PR #1086, opened 2025-01-16
fix: PlatformManager Internals

WispyMouse:
Setting `PlatformManager.PlatformInfo` and `PlatformInformation` to be `internal`, such that other code can reference it and add to it.


paulhazen (changes requested):
Looking at these changes in isolation - it's not clear to me why these need to be made `internal`. I _do_ recall there was a reason to make them internal - but that change does not seem to be reflected in this PR.

Additionally, I believe the PR should be opened into `release-3.3.5`.

paulhazen (approved):
Changing to approval per offline conversation - and being reminded why it's an important fix.

arthur740212 (approved):

andrew-hirata-playeveryware (approved):

PR #1087, opened 2025-01-16
fix: Make first time user experience clearer.

paulhazen:
In the process of debugging another issue, I discovered that our current implementation that handles the scenario where a set of named is empty leads to a little bit of a confusing user experience. The difference that this PR makes is illustrated below - and I think is a critical improvement to help onboard new users.

image


WispyMouse (approved):
Works, makes sense to me. No user will ever have a reason to have zero entries, and that makes the UI more clear. Approved.

PR #1088, opened 2025-01-16
fix: Prevent generation of empty `EpicOnlineServicesConfig.json` file.

paulhazen:
This PR corrects a problem that exists within the current release branch where an empty `EpicOnlineServicesConfig.json` file is always being generated. 

It is critical that this be avoided, since the generation and inclusion of the file in the project despite it being deprecated will otherwise be a major source of confusion for many users.

### DIFF
--- a/com.playeveryware.eos/Runtime/Core/Config/PlatformConfig.cs
+++ b/com.playeveryware.eos/Runtime/Core/Config/PlatformConfig.cs
@@ -268,6 +268,10 @@ namespace PlayEveryWare.EpicOnlineServices
 
         #region Logic for Migrating Override Values from Previous Structure
 
+        // This warning is suppressed because while EOSConfig is marked as 
+        // obsolete - it is important that it remain and be used within this
+        // section of code so that things can be properly migrated.
+#pragma warning disable CS0618 // Type or member is obsolete
 #if !EOS_DISABLE
 
         protected sealed class NonOverrideableConfigValues : Config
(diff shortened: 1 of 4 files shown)


com.playeveryware.eos/Runtime/Core/Config/PlatformConfig.cs
```diff
@@ -268,6 +268,10 @@ protected PlatformConfig(PlatformManager.Platform platform) :
 
         #region Logic for Migrating Override Values from Previous Structure
 
+        // This warning is suppressed because while EOSConfig is marked as 
+        // obsolete - it is important that it remain and be used within this
+        // section of code so that things can be properly migrated.
+#pragma warning disable CS0618 // Type or member is obsolete
```
WispyMouse, line 274: Needs a pragma restore

WispyMouse (approved):
Add pragma restore as noted in comment. Otherwise approved.

com.playeveryware.eos/Runtime/Core/Config/PlatformConfig.cs
paulhazen: Running a find all text within `PlatformConfig.cs` reveals that there is a restore directive and that it comes after the disable directive. Can you confirm?

image

paulhazen: I'm going to err on the side of moving fast and breaking things - since this isn't a PR into `development`.  Leaving conversation unresolved - which does not block merging.

PR #1095, opened 2025-01-21
fix: Client Credential auto-selection

paulhazen:
This PR addresses an issue nearly identical to the previous PR - but instead of applying to first time users defining their first deployment and having that deployment selected for each platform - this does the same thing for the first time client credentials are defined.

Notes are heavily added to emphasize the importance of reducing the duplicated code that this introduces, and the intent is to have that refactor be part of the planning for a future release.

#EOS-2367

### DIFF
--- a/com.playeveryware.eos/Runtime/Core/Config/ProductConfig.cs
+++ b/com.playeveryware.eos/Runtime/Core/Config/ProductConfig.cs
@@ -296,7 +296,7 @@ namespace PlayEveryWare.EpicOnlineServices
                 // If the config already has a completely defined set of client
                 // credentials, then do not override, and move to the next
                 // platform config.
-                if (config.clientCredentials.IsComplete)
+                if (config.clientCredentials != null && config.clientCredentials.IsComplete)
                 {
                     continue;
                 }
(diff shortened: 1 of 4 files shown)


com.playeveryware.eos/Runtime/Core/Config/ProductConfig.cs
```diff
@@ -217,8 +257,80 @@ protected override void OnWriteCompleted()
...
+                // next.
+                if (!PlatformManager.TryGetConfig(platform, out PlatformConfig config))
+                {
+                    continue;
+                }
+
+                // If the config already has a completely defined set of client
+                // credentials, then do not override, and move to the next
+                // platform config.
+                if (config.clientCredentials.IsComplete)
```
WispyMouse: My `config.clientCredentials` was null here, resulting in a nullreference.

paulhazen: Good catch. I've updated the conditional to check first if the value is not null.

arthur740212 (approved):

WispyMouse (approved):

PR #1098, opened 2025-01-21
fix: `WrappedInitializeThreadAffinity` `null` case

paulhazen:
This PR corrects an issue where after a new `PlatformConfig` is loaded `WrappedInitializeThreadAffinity` is null and subsequently errors in the config editor are generated.

The fix checks the value for null inside `OnReadComplete`, and if it is null, then a new instance is made. This works properly because `WrappedInitializeThreadAffinity` is a reference type that encloses a struct - so it is appropriate for it to have default values when first created.

#EOS-2369

### DIFF
--- a/com.playeveryware.eos/Runtime/Core/Config/PlatformConfig.cs
+++ b/com.playeveryware.eos/Runtime/Core/Config/PlatformConfig.cs
@@ -266,6 +266,14 @@ namespace PlayEveryWare.EpicOnlineServices
             Platform = platform;
         }
 
+        protected override void OnReadCompleted()
+        {
+            base.OnReadCompleted();
+
+            // If thread affinity is null then instantiate it.
+            threadAffinity = new();
+        }
+
         #region Logic for Migrating Override Values from Previous Structure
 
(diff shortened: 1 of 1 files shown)


com.playeveryware.eos/Runtime/Core/Config/PlatformConfig.cs
```diff
@@ -266,6 +266,14 @@ protected PlatformConfig(PlatformManager.Platform platform) :
             Platform = platform;
         }
 
+        protected override void OnReadCompleted()
+        {
+            base.OnReadCompleted();
+
+            // If thread affinity is null then instantiate it.
+            threadAffinity = new();
```
WispyMouse: This should nullcheck / default check. As-is it doesn't do what this comment says it does.

paulhazen: 🤦 yeah, you'd think I'd have caught that 😆 

I've corrected it now.

WispyMouse (approved):
Seems to be in a working state, now. I'm able to save thread affinity and load it up properly.

PR #1099, opened 2025-01-21
fix: Restore functionality of Unit Tests

paulhazen:
This PR corrects namespace references and assembly linkages to restore functionality of the unit tests. This PR _does not_ seek to alter any tests - merely restores the ability to run them via the Test Runner in the Unity Editor.

#EOS-2370


WispyMouse (approved):
It appears to me that all tests are discovered, and all tests run as expected. They don't all succeed, but they do run.

PR #1101, opened 2025-01-22
fix: Complete application of command line arguments from the Epic Games Launcher

paulhazen:
This PR completes the implementation of the application of deployment and sandbox values from the Epic Games Launcher.

Prior to this fix, only the sandbox would be applied - and only if the sandbox was also defined in `ProductConfig`.

#EOS-2371

### DIFF
--- a/com.playeveryware.eos/Runtime/Core/EOSManager.cs
+++ b/com.playeveryware.eos/Runtime/Core/EOSManager.cs
@@ -78,7 +78,6 @@ namespace PlayEveryWare.EpicOnlineServices
     using Extensions;
     using System.Diagnostics;
     using System.Globalization;
-    using System.IO.Pipes;
     using UnityEngine.Assertions;
     using AddNotifyLoginStatusChangedOptions = Epic.OnlineServices.Auth.AddNotifyLoginStatusChangedOptions;
     using Credentials = Epic.OnlineServices.Auth.Credentials;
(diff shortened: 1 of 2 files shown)


com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -491,40 +491,97 @@ private void InitializeOverlay(IEOSCoroutineOwner coroutineOwner)
...
+                        // Check for equality regardless of case - and
+                        // regardless of whether the dashes are included in the
+                        // Guid for the purposes of comparison.
+                        if (namedDeployment.Value.DeploymentId.ToString().Equals(epicArgs.epicDeploymentID,
+                                StringComparison.OrdinalIgnoreCase) ||
+                            namedDeployment.Value.DeploymentId.ToString("N").Equals(epicArgs.epicDeploymentID,
+                                StringComparison.OrdinalIgnoreCase))
+                        {
+                            Debug.Log($"{namedDeployment} selected as deployment.");
+                            deploymentDefined = true;
```
WispyMouse, line 551: nit; If we've found a match, then we should `break` the foreach loop.

paulhazen: Excellent observation - I've corrected the implementation. Marking the conversation as resolved.

WispyMouse (approved):

PR #1105, opened 2025-01-27
feat: Display & Make editable the Encryption Key for Client Credentials

paulhazen:
This PR allows the Encryption Key for Client Credentials to be displayed and updated.

image

image

#EOS-2379


Assets/Plugins/Source/Editor/Utility/GUIEditorUtility.cs
```diff
@@ -539,6 +540,11 @@ public static float MeasureLongestLabelWidth(List<string> labels)
             return MeasureLabelWidth(longestString);
         }
 
+        public static float MeasureLabelWidth(int characters)
+        {
+            return MeasureLabelWidth(new string('0', characters));
```
paulhazen: Should probably use `M`, since typically (get it? typically?) this is the longest character in a typeset.

com.playeveryware.eos/Runtime/Core/EOS_SDK_Additions/EOSClientCredentials.cs
```diff
@@ -45,13 +45,18 @@ static EOSClientCredentials()
         }
 
         public EOSClientCredentials()
+        {
+            EncryptionKey = GenerateEncryptionKey();
```
paulhazen, line 49: Does this override the serialized value? Make sure it does not.

Assets/Plugins/Source/Editor/Utility/GUIEditorUtility.cs
paulhazen: Done. Marking conversation as resolved.

com.playeveryware.eos/Runtime/Core/EOS_SDK_Additions/EOSClientCredentials.cs
paulhazen: It does not override the serialized value - because there is a specific constructor that is used for deserializing from JSON, it's specified using the `[JsonConstructor]` attribute.

Marking conversation as resolved.

WispyMouse:
Retargeted to `release-3.3.6`.

Assets/Plugins/Source/Editor/Utility/GUIEditorUtility.cs
```diff
@@ -1004,9 +1010,9 @@ private static SetOfNamed<EOSClientCredentials> RenderInput(ConfigFieldAttribute
                 "Enter your client information here as it appears in the Epic Dev Portal.",
                 "https://dev.epicgames.com/docs/dev-portal/product-management#clients",
                 clientCredentialsCopy,
-                (rect, item, nameAsLabel) =>
+                (rect, item, nameAsLabel) => // Things function renders input for a EOSClientCredential item.
                 {
-                    float remainingWidth = rect.width;
+                    float remainingWidth = rect.width - 5f;
```
WispyMouse, line 1015: nit: I'm wondering if we want to assign '5' to a variable representing 'X button width' or something. I see '5' used in 4 places in this PR.

Assets/Plugins/Source/Editor/EditorWindows/ModalEOSEditorWindow.cs
```diff
@@ -0,0 +1,245 @@
...
+            window.ShowModalUtility(); // Makes it modal-like
+            window.SetIsEmbedded(false);
+
+            s_currentWindow = window;
+        }
+
+        private void OnLostFocus()
+        {
+            // Refocus if the window loses focus
+            Focus();
```
WispyMouse, line 162: I suspect this is overzealously applied. The symptom of this is really simple and small. To see what I mean, open this modal overlay, and then mouse over the text field. I expect text fields to have a Windows Cursor indicating that I'm selecting things, but instead the cursor doesn't have that contextual knowledge. I'm guessing that's because it's impossible to focus this field, but everything kind of works out.

I'm not really sure if this actually causes a problem. It feels like one of those things that _someone's_ screen is going to handle poorly. Can we somehow ask if a child element has focus before trying to re-focus?

WispyMouse: Hmm.... I put a Debug Log on this and it didn't ever fire. So I wonder why the text cursor isn't working? I guess it's not related to the above, though, so I'm going to resolve this conversation and make a general comment instead.

WispyMouse (commented):
Tiny aesthetic problem: Open the key window, and mouse over the text field. I expect there would be a text cursor there indicating I'm mousing over a field, but instead it retains the previous cursor contextual state. I do not know why it does this. I suspect that might play poorly with some people's screens, but I have no repro case of this being a problem.

WispyMouse (approved):
Approved. I was able to get it into a bad state _somehow_, but I wasn't able to reproduce it again. I tried throwing all sorts of stuff at this and it seemed to hold up.

paulhazen:
> Approved. I was able to get it into a bad state _somehow_, but I wasn't able to reproduce it again. I tried throwing all sorts of stuff at this and it seemed to hold up.

I recall having a similar problem - I suspect it has to do with how the Unity Editor handles scenarios where it has loaded a new window that it has never rendered before - I've noticed this sort of behavior a lot when adding new windows. Might we worth filing a bug with Unity....

PR #1106, opened 2025-01-27
fix(google,login) : default to google login disabled

arthur740212:
google login feature require extra dependencies in gradle file to compile.
This fix disables the feature by default and can be toggled on `login.java.meta`

In summary, file moved, and meta file is set to disabled on android


WispyMouse (approved):

WispyMouse:
Taking executive decision to merge this in now.

PR #1109, opened 2025-01-29
fix: Smarter Enum Conversion

paulhazen:
This PR improves the way in which a number value is cast into an `enum` type via `EnumUtility`.

Currently, when an attempt is made to convert a value that does not correspond to a _named value_ within the `enum`, the lowest value in the `enum` is selected.

This change checks not just for if the value is out of range by being too low, but also checking for if the value is out of range by being too high. If either is the case, then instead of selecting the highest or lowest value accordingly, a value of 0 is selected.

Despite not _necessarily_ being a _named value_ within an `enum`, `0` is _always_ a valid value for any `enum`, and is used to indicate a value that is **not set**.

#EOS-2381

### DIFF
--- a/Assets/Tests/PlayMode/Config/JsonConverters/ListOfStringsToEnumConverterTests.cs
+++ b/Assets/Tests/PlayMode/Config/JsonConverters/ListOfStringsToEnumConverterTests.cs
@@ -30,7 +30,7 @@ namespace PlayEveryWare.EpicOnlineServices.Tests.Config
     using Newtonsoft.Json.Linq;
     public static partial class ListOfStringsToEnumConverterTests
     {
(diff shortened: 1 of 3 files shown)


WispyMouse:
Retargeted to `release-3.3.6`.

WispyMouse (commented):
I ran the `ListOfStringsToEnumConverterTests` set, which seem relevant here.

`ListOfStringsToEnumConverterTestClass_ReadJson_InvalidTokenType_ThrowsException` no longer throws an exception, instead setting the value to 0. Should we remove the test?

WispyMouse (approved):
Added test for Flags being set to 0. This looks good to me and functions correctly.

PR #1116, opened 2025-02-05
feat: User experience for entering invalid product version.

paulhazen:
This PR introduces a feature within the Config system that allows attributes to be added to the `productVersion` field member of the `ProductConfig` class. 

When the configuration editor window renders the input for this field, if the value is invalid a helpbox will appear above the input, and the input field itself will be highlighted.

image

Currently, this _does not prevent_ the user from saving an invalid value. Extending implementation to include that is recommended - but is beyond the scope of the task as it is currently described.

> [!IMPORTANT]
> It should be noted that implementing the task as described has a side effect that was not considered: When a first-time user loads the editor - they will see a warning indicating that the product version is invalid - but they will _not_ see similar warnings for the other fields that they have yet to fill out.
>
> Some thought needs to be put into whether to simply not display these warnings until a user tries to save - or carve out a special case for first-time users, or (ideally) complete the validation implementation to extend to the other fields that can have attributes similar to the ones contained in this PR applied to them so the behavior is consistent.

#EOS-2236


com.playeveryware.eos/Runtime/Core/Config/Attributes/Validators/FieldValidator.cs.meta
```diff
@@ -1,5 +1,5 @@
 fileFormatVersion: 2
-guid: 9209686f48242544997a71f68e50fa07
+guid: 10d2cf32aa930674697799f5f5cf5fb9
```
WispyMouse, line 2: Intentional new guid? Probably doesn't matter.

com.playeveryware.eos/Runtime/Core/Config/Attributes/Validators/LengthValidationAttribute.cs
```diff
@@ -0,0 +1,84 @@
...
+        {
+            _minLength = null;
+            _maxLength = maxLength;
+        }
+
+        public override bool ValidateStringField(string value, out string errorMessage)
+        {
+            // If the minimum length is set and greater than 0, then fail if
+            // string is either null or is too short.
+            if (_minLength is > 0 && value == null || value.Length < _minLength)
```
WispyMouse: Should we ask if `_minLength.HasValue`?

com.playeveryware.eos/Runtime/Core/Config/Attributes/Validators/LengthValidationAttribute.cs
```diff
@@ -0,0 +1,84 @@
...
+
+        // Constructor for specifying only min length
+        public LengthValidationAttribute(int minLength)
+        {
+            _minLength = minLength;
+            _maxLength = null;
+        }
+
+        // Constructor for specifying only max length
+        public LengthValidationAttribute(int maxLength, bool isMaxOnly)
```
WispyMouse, line 55: I feel a bit strange about using a parameter like this. I'm not really sure how to best resolve it, though, given that you can't have two constructors with identical parameter inputs.

com.playeveryware.eos/Runtime/Core/Config/Attributes/Validators/LengthValidationAttribute.cs
```diff
@@ -0,0 +1,84 @@
...
+        {
+            // If the minimum length is set and greater than 0, then fail if
+            // string is either null or is too short.
+            if (_minLength is > 0 && value == null || value.Length < _minLength)
+            {
+                errorMessage = $"String must be at least {_minLength.Value} characters long.";
+                return false;
+            }
+
+            if (value.Length > _maxLength)
```
WispyMouse: Should `_maxLength.HasValue` be checked?

com.playeveryware.eos/Runtime/Core/Config/Attributes/Validators/LengthValidationAttribute.cs
WispyMouse: I always feel a bit weird about if statements that are combined in this way. Perhaps you could surround it in parens?


```csharp
if ((_minLength is > 0 && value == null) || value.Length < _minLength)
```

com.playeveryware.eos/Runtime/Core/Config/ProductConfig.cs
```diff
@@ -63,7 +70,9 @@ public class ProductConfig : Config
         [ConfigField("Version",
             ConfigFieldType.Text,
             "Use this to indicate to the EOS SDK your game version.",
-            0)]
+            0, "https://dev.epicgames.com/docs/api-ref/structs/eos-initialize-options")]
+        [RegexValidation("^[A-Za-z0-9._ !?()+=:-]+$", "Product version must consist of only the following characters: A-Z, a-z, 0-9, dot, underscore, space, exclamation mark, question mark, sign, hyphen, parenthesis, plus, minus, or colon characters.")]
```
WispyMouse: It appears as though `=` also works.

com.playeveryware.eos/Runtime/Core/Config/Attributes/Validators/FieldValidator.cs.meta
paulhazen: Not intentional - I created the file - then realized it was identical to one that already existed. In the shuffle the guid got exchanged - but yeah it doesn't matter.

com.playeveryware.eos/Runtime/Core/Config/Attributes/Validators/LengthValidationAttribute.cs
paulhazen: > Should we ask if `_minLength.HasValue`?

the `is` statement is a **relational pattern**, and as such it checks if `_minLength` has a value - then compares the value to `0`.

> I always feel a bit weird about if statements that are combined in this way. Perhaps you could surround it in parens?
> 
> ```cs
> if ((_minLength is > 0 && value == null) || value.Length < _minLength)
> ```

That is an astute observation - and I have updated the implementation to include the parenthesis.

com.playeveryware.eos/Runtime/Core/Config/Attributes/Validators/LengthValidationAttribute.cs
paulhazen: Yeah I'm not a huge fan of it either - but could not figure out a way to get around the two constructors with identical parameters problem. I think a cleaner approach _might be_ to actually only have the one constructor that specifies min and max - but that makes applying the attribute to a field look weird and misleading. This seemed like the best option.

com.playeveryware.eos/Runtime/Core/Config/Attributes/Validators/LengthValidationAttribute.cs
paulhazen: It does not need to be - if it has no value, by design C# will evaluate the entire conditional to `false`. 

However, for the sake of clarity I just updated the implementation to include `_maxLength.HasValue` as a `&&` clause before the value comparison.

com.playeveryware.eos/Runtime/Core/Config/ProductConfig.cs
paulhazen: Well-spotted @WispyMouse! I've updated it to exclude that from the regex pattern.

WispyMouse (approved):

andrew-hirata-playeveryware (approved):

PR #1131, opened 2025-02-14
Fixed "Add via Git URL" link, and included instructions on how to target a specific release version.

kdserra:
[The current "Add via Git URL" instructions](https://github.com/PlayEveryWare/eos_plugin_for_unity/blob/stable/com.playeveryware.eos/Documentation~/add_plugin.md#adding-the-package-from-a-git-url) uses an invalid URL, this PR fixes the invalid link with the correct one, and also includes instructions on how to target a specific [release version tag](https://github.com/PlayEveryWare/eos_plugin_for_unity/releases).

The updated URL is `https://github.com/PlayEveryWare/eos_plugin_for_unity_upm.git` with optional version tag `#vX.X.X`

I suspect the reason the original link may have worked is that PlayEveryWare has privileged access to the repo, and the current link is likely leveraging that access, but the general public can't access it through the same means.  It's also possible the current URL is non-functional in general.

I made sure to use the `eos_plugin_for_unity_upm` repo link, and not the main `eos_plugin_for_unity` repo.

I have tested this on Unity 2022.3.57f1, and it works as expected, included is a screenshot of using the fixed link with the version selector in my project.

You can also see the updated URL in use in the "Installed From" section of the package description.

image

Release tags can be located here:
Notice the `v` prefix in the tag, this is important to include in the URL.

release-tag-example


arthur740212 (approved):

andrew-hirata-playeveryware (approved):

paulhazen (approved):
Awesome, thanks for your contribution @kdserra!

PR #1165, opened 2025-08-12
Update EOS 1.17.1.3 CL 44532354

JessTello:
Add P2P connection state tracking and handshake protocol, Multibyte character fix for Windows dllmain.cpp, EAC Fix


billfr-ApexSystems (approved):
Approved for release

EpicCliffHudson (changes requested):
There were a lot of public C# types that look like they got renamed - do we need to communicate these as breaking changes?

EpicCliffHudson (approved):

PR #1166, opened 2025-08-26
FIX: Fixed the error related to Gradle build failures on Android

JessTello:


EpicCliffHudson (approved):
Looks like you need to update to the base branch, but otherwise fine.

billfr-ApexSystems (approved):
Package change out for Android approved

PR #1167, opened 2025-08-27
Reimplement Mac EAC Settings fix and remove legacy config file

DanF-ApexSystems:


EOSPluginJeff:
We were double checking the only changes between the 4.1.0 package and 4.1.1 package were the Android changes and saw these two differences. 

[Config.cs ](https://github.com/EOS-Contrib/eos_plugin_for_unity/blob/stable/com.playeveryware.eos/Runtime/EOS_SDK/Core/Config.cs) was replaced by [common.cs a](https://github.com/EOS-Contrib/eos_plugin_for_unity/blob/stable/com.playeveryware.eos/Runtime/EOS_SDK/Core/Common.cs) and is no longer needed. It was left undeleted due to the revert we did originally when the 1.17.1.0 release was delayed.

The Mac fix was reverted incorrectly and we're placing it back.

EpicCliffHudson (approved):

billfr-ApexSystems (approved):
Changes approved

PR #1182, opened 2025-12-02
Fix: Use EOS_Auth_CopyIdToken instead of CopyUserAuthToken for Connec…

JessTello:
The plugin was using EOS_Auth_CopyUserAuthToken, which is unnecessary for ConnectLogin and can cause warnings. Updated the implementation to use EOS_Auth_CopyIdToken as recommended by the EOS API.

### DIFF
--- a/com.playeveryware.eos/Runtime/Core/EOSManager.cs
+++ b/com.playeveryware.eos/Runtime/Core/EOSManager.cs
@@ -1061,83 +1061,45 @@ namespace PlayEveryWare.EpicOnlineServices
             /// Also contains the information needed to set ProductUserId.
             /// <see cref="s_localProductUserId"/>
             /// </param>
-            public async void StartConnectLoginWithEpicAccount(EpicAccountId epicAccountId,
-                OnConnectLoginCallback onConnectLoginCallback)
+            public async void StartConnectLoginWithEpicAccount(EpicAccountId epicAccountId, OnConnectLoginCallback onConnectLoginCallback)
             {
-                var EOSAuthInterface = GetEOSPlatformInterface().GetAuthInterface();
-                var copyUserTokenOptions = new CopyUserAuthTokenOptions();
-                var result =
-                    EOSAuthInterface.CopyUserAuthToken(ref copyUserTokenOptions, epicAccountId, out Token? authToken);
-                var connectLoginOptions = new Epic.OnlineServices.Connect.LoginOptions();
+                var authInterface = GetEOSPlatformInterface().GetAuthInterface();
 
-                if (result == Result.NotFound)
+                var idOpts = new Epic.OnlineServices.Auth.CopyIdTokenOptions
                 {
-                    Log("No User Auth tokens found to login");
-                    if (onConnectLoginCallback != null)
+                    AccountId = epicAccountId
+                };
+
+                var result = authInterface.CopyIdToken(ref idOpts, out Epic.OnlineServices.Auth.IdToken? idToken);
+
+                if (result != Result.Success || !idToken.HasValue)
+                {
+                    Debug.LogError("[EOS] CopyIdToken failed" + result);
+                    var dummy = new Epic.OnlineServices.Connect.LoginCallbackInfo
                     {
-                        var dummyLoginCallbackInfo = new Epic.OnlineServices.Connect.LoginCallbackInfo();
-                        dummyLoginCallbackInfo.ResultCode = Result.ConnectAuthExpired;
-                        onConnectLoginCallback(dummyLoginCallbackInfo);
-                    }
-
+                        ResultCode = Result.InvalidAuth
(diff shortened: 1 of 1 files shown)


com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -1061,83 +1061,45 @@ public void ConnectLinkExternalAccountWithContinuanceToken(ContinuanceToken toke
...
                     return;
                 }
 
-                // If the GetUserLoginInfo delegate is set, the UserLoginInfo can
-                // be provided here for platforms that require it in this scenario.
-                if (EOSManager.GetUserLoginInfo != null)
-                {
-                    connectLoginOptions.UserLoginInfo = await EOSManager.GetUserLoginInfo();
-                }
+                Debug.Log("[EOS] Using ID Token for Connect Login");
```
matt-clarke: Please use the `Log()` function in this file, as it's conditionally compiled out when `ENABLE_DEBUG_EOSMANAGER` is not defined.

Also I see some other logging in this file which is including the class and functions names, which is a nice format. It's a minor thing but this is being changed anyway and it's nice to have extra context
```suggestion
                Log($"{nameof(EOSManager)} {nameof(StartConnectLoginWithEpicAccount)}: Using ID Token for Connect Login");
```

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -1061,83 +1061,45 @@ public void ConnectLinkExternalAccountWithContinuanceToken(ContinuanceToken toke
...
-                if (!authToken.HasValue)
+                if (result != Result.Success || !idToken.HasValue)
                 {
-                    Log("authToken was not found, unable to login");
-
-                    var dummyLoginCallbackInfo = new Epic.OnlineServices.Connect.LoginCallbackInfo();
-                    dummyLoginCallbackInfo.ResultCode = Result.InvalidAuth;
-                    onConnectLoginCallback(dummyLoginCallbackInfo);
-
+                    Debug.LogError("[EOS] CopyIdToken failed" + result);
```
matt-clarke: As below, minor change but having this additional context is handy
```suggestion
                    Debug.LogError($"{nameof(EOSManager)} {nameof(StartConnectLoginWithEpicAccount)}: CopyIdToken failed with result: {result}");
```

com.playeveryware.eos/Runtime/Core/EOSManager.cs
EpicCliffHudson: Are we not able to use the `[CallerMemberName]` attribute and the other `[Caller...]` attributes on a log function to annotate this automatically?

EpicCliffHudson (changes requested):

com.playeveryware.eos/Runtime/Core/EOSManager.cs
JessTello: > Are we not able to use the `[CallerMemberName]` attribute and the other `[Caller...]` attributes on a log function to annotate this automatically?

Bill and I discussed, and we’re considering implementing a centralized logging utility as shown below. This approach uses CallerMemberName and CallerFilePath to automatically include class and method names, so developers only need to provide the message. For example:
`EOSLog.Log("Using ID Token for Connect Login");`
Here’s what we’re thinking:

```
internal static class EOSLog
{
    public static void Log(
        string message,
        [CallerMemberName] string memberName = "",
        [CallerFilePath] string filePath = "",
        [CallerLineNumber] int lineNumber = 0)
    {
        string className = Path.GetFileNameWithoutExtension(filePath);
        Debug.Log($"[EOS][{className}::{memberName} @ {lineNumber}] {message}");
    }

    public static void LogWarning(
        string message,
        [CallerMemberName] string memberName = "",
        [CallerFilePath] string filePath = "",
        [CallerLineNumber] int lineNumber = 0)
    {
        string className = Path.GetFileNameWithoutExtension(filePath);
        Debug.LogWarning($"[EOS][{className}::{memberName} @ {lineNumber}] {message}");
    }

    public static void LogError(
        string message,
        [CallerMemberName] string memberName = "",
        [CallerFilePath] string filePath = "",
        [CallerLineNumber] int lineNumber = 0)
    {
        string className = Path.GetFileNameWithoutExtension(filePath);
        Debug.LogError($"[EOS][{className}::{memberName} @ {lineNumber}] {message}");
    }
}
```
We’ll need to add this to our backlog for full implementation across all logging. Before that, I’ll test it locally to validate the approach.
If you agree, I can include the new logging utility in a new pull request, and we can handle the full migration in a follow-up to avoid expanding the scope of this change.
Let me know if this direction looks good to you.

matt-clarke (approved):

EpicCliffHudson (approved):

PR #1184, opened 2025-12-04
Feat: achivements scene user feedback  (EOSU-653)

JessTello:
- Display updated achievement image right after unlocking for better user feedback
- Disable unlock button when achievement is already unlocked

### DIFF
--- a/Assets/Scripts/StandardSamples/Services/AchievementsService.cs
+++ b/Assets/Scripts/StandardSamples/Services/AchievementsService.cs
@@ -493,6 +493,12 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
             return achievementDefinitionCount;
         }
 
+        public struct AchievementUnlockInfo
+        {
+            public Utf8String AchievementId;
+            public string DisplayName;
+            public string Description;
+        }
         /// <summary>
         /// Unlocks the achievement for the current player.
         /// TODO: Create a callback version of this method
@@ -500,16 +506,16 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
         /// <param name="achievementId">
         /// The id of the achievement to unlock for the current player.
         /// </param>
-        public Task UnlockAchievementAsync(string achievementId)
+        public Task<PlayerAchievement> UnlockAchievementAsync(AchievementUnlockInfo info)
         {
             var localUserId = EOSManager.Instance.GetProductUserId();
             var eosAchievementOption = new UnlockAchievementsOptions
             {
                 UserId = localUserId,
-                AchievementIds = new Utf8String[] { achievementId }
+                AchievementIds = new Utf8String[] { info.AchievementId }
             };
 
-            TaskCompletionSource<object> tcs = new();
+            var tcs = new TaskCompletionSource<PlayerAchievement>();
 
             GetEOSAchievementInterface().UnlockAchievements(ref eosAchievementOption, null,
                 (ref OnUnlockAchievementsCompleteCallbackInfo data) =>
@@ -520,7 +526,16 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
                     }
                     else
                     {
-                        tcs.SetResult(null);
+                        var achievement = new PlayerAchievement()
+                        {
+                            AchievementId = info.AchievementId,
+                            DisplayName = info.DisplayName,
+                            Description = info.Description,
+                            Progress = 1.0,
+                            UnlockTime = DateTime.UtcNow,
+                            StatInfo = null
+                        };
+                        tcs.SetResult(achievement);
                     }
                 });
 
--- a/Assets/Scripts/StandardSamples/Services/AchievementsService.cs
+++ b/Assets/Scripts/StandardSamples/Services/AchievementsService.cs
@@ -493,12 +493,6 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
             return achievementDefinitionCount;
         }
 
-        public struct AchievementUnlockInfo
-        {
-            public Utf8String AchievementId;
-            public string DisplayName;
-            public string Description;
-        }
         /// <summary>
         /// Unlocks the achievement for the current player.
         /// TODO: Create a callback version of this method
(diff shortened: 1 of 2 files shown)


Assets/Scripts/StandardSamples/UI/Achievements/UIAchievementsMenu.cs
```diff
@@ -87,20 +87,48 @@ public async void IncrementLoginStat()
...
+
+                var newPlayerAchievement = new PlayerAchievement()
+                {
+                    AchievementId = definition.AchievementId,
+                    DisplayName = definition.UnlockedDisplayName,
+                    Description = definition.UnlockedDescription,
+                    Progress = 1.0,
+                    UnlockTime = DateTime.UtcNow,
+                    StatInfo = null
+                };
```
matt-clarke: The menu creating the PlayerAchievement here is a bit backwards, it shouldn't have to own that. I would update `UnlockAchievementAsync()` so the Task result is the `PlayerAchievement`

Assets/Scripts/StandardSamples/UI/Achievements/UIAchievementsMenu.cs
```diff
@@ -87,20 +87,48 @@ public async void IncrementLoginStat()
...
+
+                achievementDataList[displayIndex].PlayerData = newPlayerAchievement;
+                DisplayPlayerAchievement(definition);
+
+                Texture2D unlockedIcon = await AchievementsService.Instance.GetAchievementUnlockedIconTexture(definition.AchievementId);
+                Texture2D lockedIcon = await AchievementsService.Instance.GetAchievementLockedIconTexture(definition.AchievementId);
+
+                achievementUnlockedIcon.texture = unlockedIcon;
+                achievementLockedIcon.texture = lockedIcon;
```
matt-clarke: Is it necessary to retrieve the textures and assign them again? Wouldn't they already be set?

billfr-ApexSystems: @JessTello Validate the need for texture assignment and determine if fixing original assignment should be targeted fix as well.

JessTello: It is not being assigned twice to the same element; the first assignment corresponds to the thumbnail icon, and the second one to the main icon.
image

JessTello: Without this assignment, the main icon does not update and remains in a locked state, which prevents the expected behavior.

image

matt-clarke: With that in mind, one optimisation would be to retrieve the textures from the other icon and assign it to this icon. That eliminates the need to download the textures again.

If there's anything which makes that non-trivial then don't worry about it

Assets/Scripts/StandardSamples/Services/AchievementsService.cs
```diff
@@ -493,23 +493,29 @@ public static uint GetAchievementsCount()
...
+        }
         /// <summary>
         /// Unlocks the achievement for the current player.
         /// TODO: Create a callback version of this method
         /// </summary>
         /// <param name="achievementId">
         /// The id of the achievement to unlock for the current player.
         /// </param>
-        public Task UnlockAchievementAsync(string achievementId)
+        public Task<PlayerAchievement> UnlockAchievementAsync(AchievementUnlockInfo info)
```
matt-clarke: This class has all the data about the achievements, so it should only need the id passed in. Then you can loop over the `_achievements` list to find the data. If the data isn't found then you can set an exception on the Task.

JessTello: I’ve already implemented this comment in the new commit. Thank you for your feedback!

Assets/Scripts/StandardSamples/UI/Achievements/UIAchievementsMenu.cs
JessTello: Implemented this comment in the new commit

Assets/Scripts/StandardSamples/Services/AchievementsService.cs
```diff
@@ -493,23 +493,46 @@ public static uint GetAchievementsCount()
             return achievementDefinitionCount;
         }
 
+        public struct AchievementUnlockInfo
+        {
+            public Utf8String AchievementId;
+            public string DisplayName;
+            public string Description;
+        }
```
matt-clarke: This type is now unused and can be removed
```suggestion
```

Assets/Scripts/StandardSamples/Services/AchievementsService.cs
```diff
@@ -493,23 +493,46 @@ public static uint GetAchievementsCount()
...
         /// <summary>
         /// Unlocks the achievement for the current player.
         /// TODO: Create a callback version of this method
         /// </summary>
         /// <param name="achievementId">
         /// The id of the achievement to unlock for the current player.
         /// </param>
-        public Task UnlockAchievementAsync(string achievementId)
+        public Task<PlayerAchievement> UnlockAchievementAsync(string achivementId)
```
matt-clarke: The name of this param has a spelling mistake

matt-clarke (changes requested):

Assets/Scripts/StandardSamples/Services/AchievementsService.cs
JessTello: Fixed in the latest commit.

Assets/Scripts/StandardSamples/Services/AchievementsService.cs
JessTello: Fixed in the latest commit.

matt-clarke (approved):

billfr-ApexSystems (approved):
Approved with changes requested by Matt

PR #1186, opened 2025-12-04
Fix: Retrieve Sandbox and Deployment IDs from CLI (EOSU-668)

JessTello:
- Added new class to fetch Sandbox ID and Deployment ID from CLI arguments
- Replaced previous logic that only read values from JSON after initialization
- Ensures correct IDs are used when running builds in different environments

### DIFF
--- a/com.playeveryware.eos/Runtime/Core/EOSCommandLineOverrides.cs
+++ b/com.playeveryware.eos/Runtime/Core/EOSCommandLineOverrides.cs
@@ -20,15 +20,15 @@ namespace PlayEveryWare.EpicOnlineServices
     /// </summary>
     public static class EOSCommandLineOverrides
     {
-        private const string DeploymentKey = "-epicdeploymentid";
-        private const string SandboxKey = "-epicsandboxid";
+        private const string DeploymentKey = "-epicdeploymentid=";
+        private const string SandboxKey = "-epicsandboxid=";
     
-        [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.SubsystemRegistration)]
-        private static void ApplyCliOverrides()
+        [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.BeforeSceneLoad)]
+        private static void SaveCliOverrides()
         {
             try
             {
-                Debug.Log($"[{nameof(EOSCommandLineOverrides)}::{nameof(ApplyCliOverrides)}] Initializing CLI overrides");
+                Debug.Log($"[{nameof(EOSCommandLineOverrides)}::{nameof(SaveCliOverrides)}] Initializing CLI overrides");
     
                 // Get command line arguments
                 string[] args = Environment.GetCommandLineArgs();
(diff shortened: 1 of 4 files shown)


com.playeveryware.eos/Runtime/Core/EOSCommandLineOverrides.cs
```diff
@@ -0,0 +1,91 @@
...
+{
+    private const string DeploymentKey = "-epicdeploymentid";
+    private const string SandboxKey = "-epicsandboxid";
+
+    [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.SubsystemRegistration)]
+    private static void ApplyCliOverrides()
+    {
+        try
+        {
+            Debug.Log($"[{nameof(EOSCommandLineOverrides)}::{nameof(ApplyCliOverrides)}] Initializing CLI overrides�");
```
matt-clarke: Invalid char in message
```suggestion
            Debug.Log($"[{nameof(EOSCommandLineOverrides)}::{nameof(ApplyCliOverrides)}] Initializing CLI overrides");
```

com.playeveryware.eos/Runtime/Core/EOSCommandLineOverrides.cs
```diff
@@ -0,0 +1,91 @@
...
+{
+    /// <summary>
+    /// Holds override values for Deployment and Sandbox IDs.
+    /// </summary>
+    public static class EOSOverrideState
+    {
+        public static string DeploymentIdOverride { get; set; }
+        public static string SandboxIdOverride { get; set; }
+    }
+}
```
matt-clarke: Namespace should apply to all types in the file

com.playeveryware.eos/Runtime/Core/EOSCommandLineOverrides.cs
```diff
@@ -0,0 +1,91 @@
...
+            string sandboxArg = null;
+
+            // Parse CLI arguments for deployment and sandbox IDs
+            foreach (var arg in args)
+            {
+                if (arg.StartsWith(DeploymentKey, StringComparison.OrdinalIgnoreCase))
+                    deploymentArg = arg.Substring(DeploymentKey.Length);
+
+                if (arg.StartsWith(SandboxKey, StringComparison.OrdinalIgnoreCase))
+                    sandboxArg = arg.Substring(SandboxKey.Length);
```
matt-clarke:
```suggestion
                if (arg.StartsWith(DeploymentKey, StringComparison.OrdinalIgnoreCase))
                    deploymentArg = arg.Substring(DeploymentKey.Length);
                else if (arg.StartsWith(SandboxKey, StringComparison.OrdinalIgnoreCase))
                    sandboxArg = arg.Substring(SandboxKey.Length);
```

com.playeveryware.eos/Runtime/Core/EOSCommandLineOverrides.cs
```diff
@@ -0,0 +1,91 @@
...
+            }
+
+            // If no overrides found, exit early
+            if (deploymentArg == null && sandboxArg == null)
```
matt-clarke: Adding `IsNullOrEmpty()` adds a little extra safety
```suggestion
            if (string.IsNullOrEmpty(deploymentArg) && string.IsNullOrEmpty(sandboxArg))
```

com.playeveryware.eos/Runtime/Core/EOSCommandLineOverrides.cs
```diff
@@ -0,0 +1,91 @@
...
+                return;
+            }
+
+            Debug.Log($"[EOSBuildBootstrapper] CLI Override detected: DEPLOYMENT={deploymentArg}, SANDBOX={sandboxArg}");
+
+            // --------------------------------------------------------------------
+            // Save values globally for later access
+            // --------------------------------------------------------------------
+            EOSOverrideState.DeploymentIdOverride = deploymentArg;
+            EOSOverrideState.SandboxIdOverride = sandboxArg;
```
matt-clarke: I'm a bit confused about the need for both these override vars and updating the platform config below. If the override vars are used then why is there a need to update the platform config? 

With that in mind I would get rid of the override vars and just update the config. This eliminates the chance that someone might read the values from the config without checking the overrides.

com.playeveryware.eos/Runtime/Core/EOSCommandLineOverrides.cs
```diff
@@ -0,0 +1,91 @@
...
+
+            // --------------------------------------------------------------------
+            // Regenerate a clean platform config
+            // --------------------------------------------------------------------
```
matt-clarke: This comment isn't accurate and is unnecessary, as there's already the comment below. Also, I'm not a fan of the extra comment lines for just regular comments

matt-clarke (changes requested):

com.playeveryware.eos/Runtime/Core/EOSCommandLineOverrides.cs
JessTello: The main reason is that PlatformManager.GetPlatformConfig() is called in several places, and every time this function runs, it restores and reads the values from the JSON file. I would need to do a deeper analysis to understand why it was implemented this way and whether it’s truly necessary. However, to avoid any risk, I decided to take the safer approach and store the values so that EOSManager reads them directly, instead of using the ones that are constantly overwritten when PlatformManager.GetPlatformConfig() is called.

Additionally, with this approach, if the CLI is not used, the platform still falls back to the values from the platform JSON.

com.playeveryware.eos/Runtime/Core/EOSCommandLineOverrides.cs
JessTello: Comment applied

com.playeveryware.eos/Runtime/Core/EOSCommandLineOverrides.cs
JessTello: By the way, the reason for applying them now is to prepare for a future refactor of PlatformManager.GetPlatformConfig(). Once that method stops overwriting values, this script will already handle applying them, so the overrides could be removed without extra work. That’s why I didn’t remove the function but added an explicit comment for clarity.

com.playeveryware.eos/Runtime/Core/EOSCommandLineOverrides.cs
JessTello: Comment applied

com.playeveryware.eos/Runtime/Core/EOSCommandLineOverrides.cs
JessTello: Comment applied

com.playeveryware.eos/Runtime/Core/EOSCommandLineOverrides.cs
JessTello: Comment applied

com.playeveryware.eos/Runtime/Core/EOSCommandLineOverrides.cs
matt-clarke: That's very strange, `PlatformManager.GetPlatformConfig()` uses the static var `s_platformConfig`. Once it's been called once it should always use that cached version.

Regardless changing the `GetPlatformConfig()` function to apply the cmd line args would be safest, which can be done straight after `s_platformConfig` is assigned. This way whenever `GetPlatformConfig()` is called then these overrides are applied, even if something were to clear `s_platformConfig` since it was last called.

JessTello: You were right, Matt. Thank you so much for pointing that out. It’s now implemented in the new commit.

JessTello:
Closing this PR because the simple implementation has already been approved.

PR #1196, opened 2025-12-15
Bug/eosu 734 sandbox json update and save fix

JessTello:
This pull request addresses multiple issues and improvements related to configuration handling and editor behavior.

**Changes included:**

- Fix: Removed validation that prevented proper JSON updates when sandboxes were deleted.
- Fix: Corrected save behavior so both product and platform configs are saved on the first click.
- Fix: Removed redundant Write() call that was executed constantly while the settings window was open.
- Style: Normalized JSON line endings (CRLF → LF) across Android, iOS, and Linux configuration files for consistency.
- Refactor: Removed unnecessary AssetDatabase.Refresh() and SaveAssets() calls to prevent window initialization errors.


Assets/Plugins/Source/Editor/EditorWindows/EOSSettingsWindow.cs
```diff
@@ -255,7 +255,9 @@ private void Save()
             // Save the product config editor
             _productConfigEditor.Save();
 
-            // reload the product config editor
+            //Update deployment in current platform
+            ProductConfig.Get<ProductConfig>().Environments.TryGetFirstDefinedNamedDeployment(out var namedDep);
+            _platformConfigEditors[_selectedTab].SetDeployment(namedDep.Value);
```
matt-clarke: The `TryGet` should return a bool, which needs checking before you try access the out param

Assets/Plugins/Source/Editor/Utility/GUIEditorUtility.cs
```diff
@@ -993,6 +993,7 @@ private static void RenderSandboxInputs(ref ProductionEnvironments value)
                     if (!productionEnvironmentsCopy.Sandboxes.Remove(item))
                     {
                         // TODO: Tell user why the sandbox could not be removed.
+                        Debug.LogError($"{nameof(GUIEditorUtility)} {nameof(RenderSandboxInputs)}: Could not remove sandbox value .");
```
matt-clarke: Updating the message should cover this TODO

**(untested code)**
```suggestion
                        Debug.LogError($"{nameof(GUIEditorUtility)} {nameof(RenderSandboxInputs)}: Failed to find sandbox with name {item.Name} when trying to remove it.");
```

com.playeveryware.eos/Runtime/Core/Config/Config.cs
```diff
@@ -188,7 +188,7 @@ private void MigrateConfigIfNeeded()
         {
             MigrateConfigIfNeededInternal();
 #if UNITY_EDITOR
-            Write();
+            //Write();
 #endif
```
matt-clarke: `#if` block can be deleted

com.playeveryware.eos/Runtime/Core/Config/Config.cs
```diff
@@ -506,11 +506,6 @@ public virtual async Task WriteAsync(bool prettyPrint = true)
 
             var json = JsonUtility.ToJson(this, prettyPrint);
 
-            // If the json hasn't changed since it was last read, then
-            // take no action.
-            if (json == _lastReadJsonString)
-                return;
```
matt-clarke, line 512: What item in the description is related to this? Is it this one?
> Fix: Corrected save behavior so both product and platform configs are saved on the first click.

Regardless this is more of a case of not fixing the underlying problem and just removing this check. The real issue is why does this evaluate to true and result in aborting the save

matt-clarke (dismissed):

com.playeveryware.eos/Runtime/Core/Config/Config.cs
JessTello: That condition was actually preventing the sandboxes from being deleted. I tried setting it to null, but it still didn’t work. Let me check again today with more time, because that condition definitely causes the changes not to be saved when they are deleted, even if  _lastReadJsonString it’s null.

Assets/Plugins/Source/Editor/EditorWindows/EOSSettingsWindow.cs
JessTello: This has been implemented in the following commit.

Assets/Plugins/Source/Editor/Utility/GUIEditorUtility.cs
JessTello: This has been implemented in the following commit.

com.playeveryware.eos/Runtime/Core/Config/Config.cs
JessTello: This has been implemented in the following commit.

com.playeveryware.eos/Runtime/Core/Config/Config.cs
JessTello: You’re absolutely right. Looking at it now with a different perspective than a few days ago, the only thing missing was updating the _lastReadJsonString variable when writing. Thanks for pointing that out! 
This has been implemented in the following commit.

billfr-ApexSystems (approved):
Approved with changes requested

EpicCliffHudson (approved):

PR #1201, opened 2025-12-30
Bug: Fix Windows build error caused by asmdef changes (EOSU-797)

JessTello:
- Updated .asmdef to build only when UNITY_INCLUDE_TESTS is defined
- Ensured test assemblies are excluded from production builds


billfr-ApexSystems (approved):
Approved with constraint to handle UNITY_INCLUDE_TESTS

PR #1225, opened 2026-01-23
fix\EOSU-857-Lost-UI-Navigation  --> 6.0.0

OswaldoApexSystems:
[ EOSU - 857 ](https://epicgames.atlassian.net/browse/EOSU-857)

In some menus, a selectable object variable was missing, so sometimes the reference could be lost and to recover it, events had to be called in the sample.

To avoid this, all variables in the menus were referenced.


matt-clarke (approved):

billfr-ApexSystems (approved):

OswaldoApexSystems:
@DanF-ApexSystems this needs to be replicated in restricted.

PR #1237, opened 2026-02-18
Eosu 404 auth token reauth p2p disconnects

JessTello:
Added Connect AuthExpiration notify registration with minimal diagnostic logging
- Cached the last Connect LoginOptions and credential type to support safe re-auth attempts
- Implemented PUID change detection to log when the ProductUserId remains stable or changes mid-session
- Introduced a reauth-in-progress guard to prevent concurrent Connect.Login storms during token refresh
- Added a debug hook to force a Connect re-login using a fresh Epic IdToken (CopyIdToken -> Connect.Login)


com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -1173,26 +1177,53 @@ public void StartConnectLoginWithOptions(ExternalCredentialType externalCredenti
...
-                        if (connectLoginData.ResultCode != Result.Success)
+                        try
                         {
-                            Log($"Connect login was not successful. ResultCode: {connectLoginData.ResultCode}", LogType.Error);
-                        }
+                            if (connectLoginData.ResultCode != Result.Success)
+                            {
+                                Log($"Connect login was not successful. ResultCode: {connectLoginData.ResultCode}", LogType.Error);
+                                OnConnectLogin?.Invoke(connectLoginData);
+                                onloginCallback?.Invoke(connectLoginData);
```
matt-clarke: With this here `onLoginCallback` would be called twice (bc it's also in the finally block)
```suggestion
```

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -1173,26 +1177,53 @@ public void StartConnectLoginWithOptions(ExternalCredentialType externalCredenti
             public void StartConnectLoginWithOptions(Epic.OnlineServices.Connect.LoginOptions connectLoginOptions,
                 OnConnectLoginCallback onloginCallback)
             {
+                Debug.Log($"[EOS][ConnectLogin] t={Time.realtimeSinceStartup:F1} puid(prev)={s_lastKnownProductUserId} cred={s_lastConnectCredentialType}");
```
matt-clarke: Logging in this file should use the `Log()` function because it's conditionally compiled out with the `ENABLE_DEBUG_EOSMANAGER` flag

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -1173,26 +1177,53 @@ public void StartConnectLoginWithOptions(ExternalCredentialType externalCredenti
...
+                            }
+
+                            Debug.Log($"[EOS][PUID] unchanged {newPuid}");
+                            s_lastKnownProductUserId = newPuid;
+
+                            SetLocalProductUserId(newPuid);
                             ConfigureConnectStatusCallback();
-                            ConfigureConnectExpirationCallback(connectLoginOptions);
+                            ConfigureConnectExpirationCallback();
                             OnConnectLogin?.Invoke(connectLoginData);
```
matt-clarke: Moving the `OnConnectLogin` call into the `finally` block also cleans up the code above and is a nice way to make sure it's always called

com.playeveryware.eos/Runtime/Core/EOSManager.cs
matt-clarke: We shouldn't log PUID/Epic account ids, created a thread to discuss this further

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -1355,25 +1386,29 @@ private void ConfigureConnectStatusCallback()
...
-                            StartConnectLoginWithOptions(connectLoginOptions, null);
-                        });
+                if (s_notifyConnectAuthExpirationCallbackHandle != null)
+                    return;
 
-                    s_notifyConnectAuthExpirationCallbackHandle = new NotifyEventHandle(callbackHandle, handle =>
+                var EOSConnectInterface = GetEOSConnectInterface();
+                var addNotifyAuthExpirationOptions = new AddNotifyAuthExpirationOptions();
+
+                ulong callbackHandle = EOSConnectInterface.AddNotifyAuthExpiration(
```
matt-clarke: Not your change but we might as well fix the var name whilst here
```suggestion
                var eosConnectInterface = GetEOSConnectInterface();
                var addNotifyAuthExpirationOptions = new AddNotifyAuthExpirationOptions();

                ulong callbackHandle = eosConnectInterface.AddNotifyAuthExpiration(
```

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -1355,25 +1386,29 @@ private void ConfigureConnectStatusCallback()
...
-                }
+                Debug.Log("[EOS][AuthExp] registered");
+                s_notifyConnectAuthExpirationCallbackHandle = new NotifyEventHandle(callbackHandle, handle =>
+                {
+                    GetEOSConnectInterface()?.RemoveNotifyAuthExpiration(handle);
+                });
             }
 
+
+
```
matt-clarke, line 1389:
```suggestion
```

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -1847,6 +2001,9 @@ static public EOSSingleton Instance
         ///     <item><description>Calls <c>Init()</c></description></item>
         /// </list>
         /// </summary>
+        /// 
+      
+
```
matt-clarke:
```suggestion
```

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -1807,6 +1837,130 @@ private static void UpdateNetworkStatus()
...
+                Log($"{nameof(EOSManager)} RefreshConnectLoginWithFreshEpicIdToken: Re-logging Connect with fresh JWT.", LogType.Warning);
+
+                StartConnectLoginWithOptions(refreshedLoginOptions, null);
+            }
+
+            public void Debug_RefreshConnectLoginWithFreshEpicIdToken()
+            {
+                Debug.Log("[EOS][ReauthTest] Manual reauth triggered");
+                RefreshConnectLoginWithFreshEpicIdToken();
+            }
```
matt-clarke, line 1888: Test function needs removing (at least it appears to be unused)

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -1807,6 +1837,130 @@ private static void UpdateNetworkStatus()
...
+
+                    if (s_lastConnectCredentialType == Epic.OnlineServices.ExternalCredentialType.EpicIdToken)
+                    {
+                        RefreshConnectLoginWithFreshEpicIdToken();
+                    }
+                    else
+                    {
+                        Debug.LogWarning(
+                            $"{nameof(EOSManager)} Connect auth expired but last credential type was {s_lastConnectCredentialType}. " +
+                            $"Plugin cannot refresh that token automatically. Game should re-fetch external token and call StartConnectLoginWithOptions again.");
```
matt-clarke: Is there documentation for this? It's a bit of an unusual case where it handles it in certain circumstances

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -1173,26 +1177,53 @@ public void StartConnectLoginWithOptions(ExternalCredentialType externalCredenti
...
+
+                            var newPuid = connectLoginData.LocalUserId;
+
+                            if (s_lastKnownProductUserId != null && !s_lastKnownProductUserId.Equals(newPuid))
+                            {
+                                Debug.LogError(
+                                    $"[EOS][PUID] changed {s_lastKnownProductUserId} => {newPuid}. " +
+                                    "This can break P2P socket mapping. Refusing to overwrite silently.");
+                                OnConnectLogin?.Invoke(connectLoginData);
+                                return;
```
matt-clarke: Let's discuss this in the thread

matt-clarke (changes requested):

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -1807,6 +1837,130 @@ private static void UpdateNetworkStatus()
...
+                    Credentials = creds,
+                    UserLoginInfo = connectLoginOptions.UserLoginInfo
+                };
+            }
+
+
+            private void OnConnectAuthExpiration(ref AuthExpirationCallbackInfo callbackInfo)
+            {
+                Debug.Log($"[EOS][AuthExp] fired t={Time.realtimeSinceStartup:F1} puid={s_lastKnownProductUserId} cred={s_lastConnectCredentialType}");
+                lock (s_connectReauthLock)
```
matt-clarke: This code should be always running on the main thread anyway because all callbacks will be fired from the thread where `PlatformInterface.Tick()` is called. Was this a safety measure or did that not happen?

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -1807,6 +1837,130 @@ private static void UpdateNetworkStatus()
...
+                {
+                    lock (s_connectReauthLock)
+                    {
+                        s_connectReauthInProgress = false;
+                    }
+                }
+            }
+
+
+            private void RefreshConnectLoginWithFreshEpicIdToken()
```
matt-clarke, line 1869: The code in this function overlaps a lot with what's in `StartConnectLoginWithEpicAccount()`.
With a little work you should be able to move most of the logic in that function to a new one and re-use it here? The other function is async but it could call GetUserLoginInfo upfront, so no issues there

com.playeveryware.eos/Runtime/Core/EOSManager.cs
JessTello: Fixed in the next commit. I removed the extra onloginCallback?.Invoke()

com.playeveryware.eos/Runtime/Core/EOSManager.cs
JessTello: Fixed in the next commit. I switched to using the Log() wrapper instead of Debug.Log

JessTello: Updated in the next commit the remaining PUID/EAS identifier logs (including the legacy SetLocalProductUserId() message) so no user IDs are printed.

com.playeveryware.eos/Runtime/Core/EOSManager.cs
JessTello: I centralized onloginCallback into the finally to guarantee single delivery. For OnConnectLogin, I kept the explicit invokes in each outcome branch to preserve the early-return flow and clarity, but I can refactor it into the finally as well (single invoke) if you prefer.

com.playeveryware.eos/Runtime/Core/EOSManager.cs
JessTello: Removed

com.playeveryware.eos/Runtime/Core/EOSManager.cs
JessTello: Added handling for auth-expiration by credential type: we only auto-refresh when the last Connect credential is EpicIdToken (via fresh CopyIdToken -> Connect.Login). For other credential types we log a warning and require the game to re-fetch the external token and re-initiate StartConnectLoginWithOptions.

com.playeveryware.eos/Runtime/Core/EOSManager.cs
JessTello: Agreed on the overlap. Since StartConnectLoginWithEpicAccount() already performs the fresh IdToken + Connect.Login flow, I updated RefreshConnectLoginWithFreshEpicIdToken() to be a thin wrapper that reuses that existing codepath (silent, no callback) and keeps the reauth-in-progress guard.

com.playeveryware.eos/Runtime/Core/EOSManager.cs
JessTello: Removed the lock and replaced it with a simple reauth-in-progress guard to prevent concurrent reauth storms.

com.playeveryware.eos/Runtime/Core/EOSManager.cs
matt-clarke: In this case my thinking is mostly around ensuring that the callback is always called, as people can forget to call callbacks. Moving it into the finally block is the safest way to do that

com.playeveryware.eos/Runtime/Core/EOSManager.cs
matt-clarke: What I mean is that this is only something that developers will see when running the game for long enough that auth expires. Also, in a specific case the plugin is handling re-authing for them, so it would be easy for someone to miss this on other platforms. It would be good to get documentation added that covers this to avoid support tickets

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -190,16 +190,29 @@ enum EOSState
...
             static private NotifyEventHandle s_notifyConnectAuthExpirationCallbackHandle;
 
+            // --- Connect reauth hardening (prevents PUID churn / P2P route resets) ---
+            // NOTE: EOS callbacks are expected to run on the same thread that drives PlatformInterface.Tick().
+            // In typical Unity integrations that's the main thread. We keep this as a simple guard to avoid
+            // reauth/login storms during token refresh.
+            static private bool s_connectReauthInProgress = false;
+
+            // IMPORTANT: LoginOptions is a struct, so store it as nullable
+            static private Epic.OnlineServices.Connect.LoginOptions? s_lastConnectLoginOptions;
```
matt-clarke: This is now unused

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -190,16 +190,29 @@ enum EOSState
...
             static private ProductUserId s_localProductUserId;
 
             static private NotifyEventHandle s_notifyLoginStatusChangedCallbackHandle;
             static private NotifyEventHandle s_notifyConnectLoginStatusChangedCallbackHandle;
             static private NotifyEventHandle s_notifyConnectAuthExpirationCallbackHandle;
 
+            // --- Connect reauth hardening (prevents PUID churn / P2P route resets) ---
+            // NOTE: EOS callbacks are expected to run on the same thread that drives PlatformInterface.Tick().
+            // In typical Unity integrations that's the main thread. We keep this as a simple guard to avoid
+            // reauth/login storms during token refresh.
```
matt-clarke: Can you move this to where the variable is checked in RefreshConnectLoginWithFreshEpicIdToken()? It's more likely to be spotted there

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -1807,6 +1839,84 @@ private static void UpdateNetworkStatus()
...
+                    if (s_lastConnectCredentialType == Epic.OnlineServices.ExternalCredentialType.EpicIdToken)
+                    {
+                        RefreshConnectLoginWithFreshEpicIdToken();
+                    }
+                    else
+                    {
+                        Log(
+                            $"{nameof(EOSManager)} Connect auth expired but last credential type was {s_lastConnectCredentialType}. " +
+                            "Plugin cannot refresh that token automatically. Game should re-fetch external token and call StartConnectLoginWithOptions again.",
+                            LogType.Warning);
```
matt-clarke: In this case it would be good to keep this logging using Debug.LogWarning as it's very infrequent and would be very useful in logs

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -1355,25 +1388,29 @@ private void ConfigureConnectStatusCallback()
...
+                var eosConnectInterface = GetEOSConnectInterface();
+                var addNotifyAuthExpirationOptions = new AddNotifyAuthExpirationOptions();
+
+                ulong callbackHandle = eosConnectInterface.AddNotifyAuthExpiration(
+                    ref addNotifyAuthExpirationOptions, null,
+                    (ref AuthExpirationCallbackInfo callbackInfo) =>
                     {
-                        GetEOSConnectInterface()?.RemoveNotifyAuthExpiration(handle);
+                        OnConnectAuthExpiration(ref callbackInfo);
                     });
```
matt-clarke: Minor thing but the args match, so you should be able to just do this

```suggestion
                    OnConnectAuthExpiration);
```

DanF-ApexSystems:
Once requested changes are resolved, please retarget to [release-6.0.2](https://github.com/EOS-Contrib/eos_plugin_for_unity/tree/release-6.0.2)

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -190,16 +191,17 @@ enum EOSState
...
-            private static float s_nextNetworkStatusUpdateTime = 0.0f;
             static private EpicAccountId s_localUserId;
             static private ProductUserId s_localProductUserId;
 
             static private NotifyEventHandle s_notifyLoginStatusChangedCallbackHandle;
             static private NotifyEventHandle s_notifyConnectLoginStatusChangedCallbackHandle;
             static private NotifyEventHandle s_notifyConnectAuthExpirationCallbackHandle;
 
+            // --- Connect reauth hardening (prevents PUID churn / P2P route resets) ---
+            static private int s_connectReauthInProgress = 0;// Last Connect credential type used. Used to decide whether we can auto-refresh on auth expiration.
```
matt-clarke:
```suggestion
            static private int s_connectReauthInProgress = 0;
            // Last Connect credential type used. Used to decide whether we can auto-refresh on auth expiration.
```

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -259,9 +261,14 @@ private string PUIDToString(ProductUserId puid)
...
-                Log("Changing PUID: " + PUIDToString(s_localProductUserId) + " => " +
-                      PUIDToString(localProductUserId));
+                var previous = s_localProductUserId;
                 s_localProductUserId = localProductUserId;
+
+                // Do not log user identifiers (PUID/EAS ids).
+                if (previous != null && localProductUserId != null && !previous.Equals(localProductUserId))
+                {
+                    Log("[EOS] Local ProductUserId changed during runtime. This likely indicates account switch or login with different credentials. Game should tear down/rebuild P2P as needed.", LogType.Error);
+                }
```
matt-clarke: Comment is a bit unnecessary/hidden here, that info should be in a document shared with the team. Also reduced this to a warning
```suggestion
                if (previous != null && localProductUserId != null && !previous.Equals(localProductUserId))
                {
                    Log("[EOS] Local ProductUserId changed during runtime. This likely indicates account switch or login with different credentials. Game should tear down/rebuild P2P as needed.", LogType.Warning);
                }
```

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -1173,26 +1173,38 @@ public void StartConnectLoginWithOptions(ExternalCredentialType externalCredenti
             public void StartConnectLoginWithOptions(Epic.OnlineServices.Connect.LoginOptions connectLoginOptions,
                 OnConnectLoginCallback onloginCallback)
             {
+                CacheLastConnectCredentialType(connectLoginOptions);
+                // Do not log user identifiers (PUID/EAS ids).
```
matt-clarke: As per the other comment about this info being in a doc that is shared with the team
```suggestion
```

matt-clarke (changes requested):

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -1807,6 +1815,77 @@ private static void UpdateNetworkStatus()
...
+            /// - For other external credential types, the plugin cannot refresh tokens automatically; the game must
+            ///   re-fetch the external token from the platform and call StartConnectLoginWithOptions again.
+            /// </summary>
+            private void OnConnectAuthExpiration(ref AuthExpirationCallbackInfo callbackInfo)
+            {
+                // Do not log user identifiers (PUID/EAS ids).
+                Log($"[EOS][AuthExp] received. CredentialType={s_lastConnectCredentialType}", LogType.Warning);
+
+                // Prevent concurrent re-auth attempts from repeated expiration notifications.
+                if (Interlocked.Exchange(ref s_connectReauthInProgress, 1) == 1)
```
matt-clarke: Asked a question about this in the thread, also if we keep this lets just tweak it for extra safety
```suggestion
                if (Interlocked.Exchange(ref s_connectReauthInProgress, 1) > 0)
```

com.playeveryware.eos/Runtime/Core/EOSManager.cs
JessTello: Fixed in new commit

com.playeveryware.eos/Runtime/Core/EOSManager.cs
JessTello: Fixed

com.playeveryware.eos/Runtime/Core/EOSManager.cs
JessTello: Fixed

com.playeveryware.eos/Runtime/Core/EOSManager.cs
JessTello: Fixed

com.playeveryware.eos/Runtime/Core/EOSManager.cs
JessTello: Fixed

com.playeveryware.eos/Runtime/Core/EOSManager.cs
JessTello: In the new commit on line 1830 I added a very detailed comment. Please let me know if this is enough or if you prefer that I include an XML in the documentation.

com.playeveryware.eos/Runtime/Core/EOSManager.cs
JessTello: Fixed

com.playeveryware.eos/Runtime/Core/EOSManager.cs
JessTello: Fixed

com.playeveryware.eos/Runtime/Core/EOSManager.cs
JessTello: Fixed

com.playeveryware.eos/Runtime/Core/EOSManager.cs
JessTello: Fixed

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -1792,6 +1815,78 @@ private static void UpdateNetworkStatus()
...
+            /// - EpicIdToken:
+            ///   The plugin can recover automatically by obtaining a fresh Epic ID token and re-running Connect.Login.
+            /// - Other external credential types:
+            ///   The plugin cannot refresh the external token automatically because that token must be re-fetched by the game/platform integration. In those cases,
+            ///   the game must obtain a fresh external token and call StartConnectLoginWithOptions again.
+            /// This distinction is important because automatic recovery is only supported for credential flows where the plugin can source a fresh token itself.
+            /// </summary>
+            private void OnConnectAuthExpiration(ref AuthExpirationCallbackInfo callbackInfo)
+            {
+                // Do not log user identifiers (PUID/EAS ids).
```
matt-clarke:
```suggestion
```

com.playeveryware.eos/Runtime/Core/EOSManager.cs
matt-clarke: Yea, adding a markdown file to the documentation is what I mean

matt-clarke (changes requested):

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -1792,6 +1815,77 @@ private static void UpdateNetworkStatus()
...
+            /// - For other external credential types, the plugin cannot refresh tokens automatically; the game must
+            ///   re-fetch the external token from the platform and call StartConnectLoginWithOptions again.
+            /// </summary>
+            private void OnConnectAuthExpiration(ref AuthExpirationCallbackInfo callbackInfo)
+            {
+                // Do not log user identifiers (PUID/EAS ids).
+                Log($"[EOS][AuthExp] received. CredentialType={s_lastConnectCredentialType}", LogType.Warning);
+
+                // Prevent concurrent re-auth attempts from repeated expiration notifications.
+                if (Interlocked.Exchange(ref s_connectReauthInProgress, 1) == 1)
```
matt-clarke: Discussion in the thread about whether locking is required is still ongoing in the thread

PR #1255, opened 2026-03-11
Egpi Redaction Utility

JessTello:
- Prevent EGPI (identifiers beyond PII) from being written to logs.
- Ensure remote player identifiers are not leaked by plugin logging paths while preserving developer debugging ability.
- Provide a single, reusable redaction utility that follows the requested masking rules and can be disabled in debug builds.


com.playeveryware.eos/Runtime/Core/Utility/EgpiRedactionUtility.cs
```diff
@@ -0,0 +1,98 @@
...
+ */
+
+#if !EOS_DISABLE
+
+namespace PlayEveryWare.EpicOnlineServices.Utility
+{
+    using System;
+    using System.Text.RegularExpressions;
+
+    public static class EgpiRedactionUtility
```
matt-clarke: Can you rename this class/file to `LoggingUtils`? It's better to have it be more general than refer to EGPI

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -736,14 +737,15 @@ private static void SimplePrintCallback(ref LogMessage message)
...
+                string sanitizedMessage = EgpiRedactionUtility.RedactKnownKeyValueIdentifiers(message.Message);
                 Debug.LogFormat(
                     type,
                     LogOption.NoStacktrace,
                     null, "{0:O} {1}({2}): {3}",
                     dateTime.ToString(DateTimeFormatInfo.InvariantInfo),
                     messageCategory,
                     message.Level,
-                    message.Message);
+                    sanitizedMessage);
```
matt-clarke: The SDK will be redacting info itself so you don't need to do anything here

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -853,7 +855,8 @@ private static void SimplePrintCallbackWithCallstack(LogMessage message)
                 var dateTime = DateTime.Now;
                 var messageCategory = message.Category.Length == 0 ? new Utf8String() : message.Category;
 
-                Log(string.Format("{0:O} {1}({2}): {3}", dateTime, messageCategory, message.Level, message.Message));
+                string sanitizedMessage = EgpiRedactionUtility.RedactKnownKeyValueIdentifiers(message.Message);
+                Log(string.Format("{0:O} {1}({2}): {3}", dateTime, messageCategory, message.Level, sanitizedMessage));
```
matt-clarke: As per comment in SimplePrintCallback about the SDK already redacting logs

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -241,7 +241,7 @@ private string PUIDToString(ProductUserId puid)
                 string toReturn = null;
                 if (puid != null)
                 {
-                    toReturn = puid.ToString();
+                    toReturn = EgpiRedactionUtility.RedactIdentifier(puid.ToString());
```
matt-clarke: There's reasons to want a PUID as a string which aren't just for logging, for that reason I wouldn't redact it here and would instead redact wherever it's logged

com.playeveryware.eos/Runtime/Core/Utility/EgpiRedactionUtility.cs
```diff
@@ -0,0 +1,98 @@
...
+    {
+        private static readonly Regex s_keyValueRegex = new Regex(
+        @"(?<key>Puid|ProductUserId|EpicAccountId|AccountId|LocalUserId|RemoteUserId|PlatformId|PlatformName|DisplayName)\b\s*[:=]\s*(?<quote>[""']?)(?<value>[A-Za-z0-9_\-\.]+)\k<quote>",
+        RegexOptions.Compiled);
```
matt-clarke: Was the main reason for adding this regex because of the logging that comes from the SDK? As pointed out in EOSManager.cs that doesn't need to be redacted by the plugin. 

As far as logging from the plugin I also prefer that any redaction is done at the point of logging too. The regex may be convenient but It makes it easier to miss if something is redacted.

e.g.
```csharp
// regex - must refer to the keys above to know if this is redacted
Debug.Log($"ProductUserId: {puid}");

// deliberate - not automatic but when checking code it's visible
Debug.Log($"ProductUserId: {LoggingUtils.Redact(puid)}";
```

com.playeveryware.eos/Runtime/Core/Utility/EgpiRedactionUtility.cs
```diff
@@ -0,0 +1,98 @@
...
+#if !EXTERNAL_TO_UNITY
+                return !UnityEngine.Debug.isDebugBuild;
+#else
+                return true;
+#endif
+#endif
+            }
+        }
+
+        public static string RedactIdentifier(string identifier, int preserveChars = 3)
```
matt-clarke: With the above suggestion we can just keep this function and pass it the value string
`public static string Redact(string value, int preserveChars = 3)`

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -347,7 +347,8 @@ public bool IsOverlayOpenWithExclusiveInput()
             [Conditional("ENABLE_DEBUG_EOSMANAGER")]
             internal static void Log(string toPrint, LogType type = LogType.Log)
             {
-                Debug.LogFormat(type, LogOption.None, null, toPrint);
+                string sanitizedLog = EgpiRedactionUtility.RedactKnownKeyValueIdentifiers(toPrint);
+                Debug.LogFormat(type, LogOption.None, null, sanitizedLog);
```
matt-clarke: As per the other file I'd prefer that the values are redacted where they're passed to the logging function

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -347,7 +347,8 @@ public bool IsOverlayOpenWithExclusiveInput()
...
             }
```
matt-clarke, line 352: _(this doesn't apply to the code here but it's relevant to the comment above)_

There needs to be a pass to check all logging in the codebase to make sure values are redacted appropriately. I'm not sure if you were planning a follow up PR but this needs to be done.

As an example here's one where a remote user's PUID is logged: https://github.com/EOS-Contrib/eos_plugin_for_unity/blob/e71e562b697d6c2d9d664ecce6c86f351595261e/Assets/Scripts/EOSPeer2PeerManager.cs#L535

There's a lot of logging via Unity's `Debug` class, both directly and some classes may also have their own Log functions like this one. That means the messages passed into them need to be checked too.

com.playeveryware.eos/Runtime/Core/Utility/EgpiRedactionUtility.cs
```diff
@@ -0,0 +1,98 @@
...
+    using System;
+    using System.Text.RegularExpressions;
+
+    public static class EgpiRedactionUtility
+    {
+        private static readonly Regex s_keyValueRegex = new Regex(
+        @"(?<key>Puid|ProductUserId|EpicAccountId|AccountId|LocalUserId|RemoteUserId|PlatformId|PlatformName|DisplayName)\b\s*[:=]\s*(?<quote>[""']?)(?<value>[A-Za-z0-9_\-\.]+)\k<quote>",
+        RegexOptions.Compiled);
+
+        public static bool ShouldRedactIdentifiers
```
matt-clarke: With that simplification there's no handling identifiers anymore, so `ShouldRedactValues` makes more sense

matt-clarke (changes requested):

Assets/Scripts/P2PNetcodeSample/Networking/EOSTransport.cs
```diff
@@ -138,7 +138,7 @@ public override void Send(ulong clientId, ArraySegment<byte> payload, NetworkDel
 
             ProductUserId userId = GetUserId(clientId);
 
-            Log($"EOSP2PTransport.Send: [ClientId='{clientId}', UserId='{userId}', PayloadBytes='{payload.Count}', SendTimeSec='{Time.realtimeSinceStartup}']");
+            Log($"EOSP2PTransport.Send: [ClientId='{clientId}', UserId='{PlayEveryWare.EpicOnlineServices.Utility.LoggingUtils.Redact(userId?.ToString())}', PayloadBytes='{payload.Count}', SendTimeSec='{Time.realtimeSinceStartup}']");
```
matt-clarke: Can you add the following to each of these files, including the full namespace here is quite long
`using PlayEveryWare.EpicOnlineServices.Utility;`

Assets/Scripts/P2PNetcodeSample/Networking/EOSTransport.cs
```diff
@@ -196,7 +196,7 @@ public override NetworkEvent PollEvent(out ulong clientId, out ArraySegment<byte
                 payload = new ArraySegment<byte>();
                 receiveTime = Time.realtimeSinceStartup;
                 NetworkEvent networkEventType = evntIsConnectionEvent ? NetworkEvent.Connect : NetworkEvent.Disconnect;
-                Log($"EOSP2PTransport.PollEvent: [{networkEventType}, ClientId='{clientId}', UserId='{evntUserId}', PayloadBytes='{payload.Count}', RecvTimeSec='{receiveTime}']");
+                Log($"EOSP2PTransport.PollEvent: [{networkEventType}, ClientId='{clientId}', UserId='{PlayEveryWare.EpicOnlineServices.Utility.LoggingUtils.Redact(evntUserId?.ToString())}', PayloadBytes='{payload.Count}', RecvTimeSec='{receiveTime}']");
```
matt-clarke: If you add a generic overload then this can become much cleaner:
`LoggingUtils.Redact(evntUserId);`

```csharp
public static string Redact<T>(T value, int preserveChars = 3)
    where T : class
{
    Redact(value != null ? value.ToString() : null, preserveChars);
}
```

Structs will still need to use the main Redact function and have .ToString() called but that's fine

matt-clarke (changes requested):

Assets/Scripts/P2PNetcodeSample/Networking/EOSTransport.cs
JessTello: Fixed in latest commit

Assets/Scripts/P2PNetcodeSample/Networking/EOSTransport.cs
JessTello: Fixed in lattest commit

matt-clarke (approved):

billfr-ApexSystems (approved):

PR #1256, opened 2026-03-11
Update EOSHighFrequencyP2P.cs

JessTello:
- Documents why the first character is stripped before parsing
- Reduces ambiguity for future maintainers
- Improves clarity around HighFrequency P2P coordinate message format

### DIFF
--- a/Assets/Scripts/EOSHighFrequencyP2P.cs
+++ b/Assets/Scripts/EOSHighFrequencyP2P.cs
@@ -49,30 +49,6 @@ namespace PlayEveryWare.EpicOnlineServices.Samples
         public bool sendActive = false;
         private List<float> dataDump;
         private const string CoordinateMessagePrefix = "m";
-
-        public static string SerializeCoordinatePacket(float x, float y)
-        {
-            // Coordinate packets are serialized as: "m<x>,<y>".
-            // The leading "m" identifies a coordinate message.
-            return string.Format(CultureInfo.InvariantCulture, "{0}{1},{2}", CoordinateMessagePrefix, x, y);
-        }
-
-        public static bool TryDeserializeCoordinatePacket(string packet, out float x, out float y)
-        {
-            x = 0f;
-            y = 0f;
-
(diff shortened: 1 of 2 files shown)


Assets/Scripts/EOSHighFrequencyP2P.cs
```diff
@@ -47,6 +47,10 @@ public class EOSHighFrequencyPeer2PeerManager : IEOSSubManager
 
         public bool sendActive = false;
         private List<float> dataDump;
+        // Coordinate packets are serialized as: "m<x>,<y>".
+        // The leading "m" identifies a coordinate message.
+        // The payload parser intentionally removes this prefix via Substring(1)
+        // before splitting the remaining "<x>,<y>" values.
         public const string CoordinateMessagePrefix = "m";
```
matt-clarke: I realise this was just adding a comment but a minor refactor would be good here. This class owning the serialization and deserialization makes it much simpler to use.

```
// This is now private
private const string CoordinateMessagePrefix = "m";

// New function for UIHighFrequencyPeer2PeerMenu.cs to call
public static string SerializeCoordinatePacket(int x, int y)
{
    // Coordinate packets are serialized as: "m<x>,<y>".
    // The leading "m" identifies a coordinate message.
    return $"{CoordinateMessagePrefix}{x},{y}";
}

public static bool TryDeserializeCoordinatePacket(string packet, out int x, out int y)
{
    // Remove the CoordinateMessagePrefix via Substring(1) before splitting the remaining "<x>,<y>" values
    
    // Move the code at line 320 here and make it use the new function
}
```

matt-clarke (changes requested):

Assets/Scripts/EOSHighFrequencyP2P.cs
JessTello: Done in latest commit

matt-clarke (approved):

billfr-ApexSystems (approved):

kisecrow9334 (approved):

PR #1260, opened 2026-03-19
Eosu 404 auth token reauth p2p disconnects v2

JessTello:
- Add a single flight guard s_isConnectAuthExpirationLoginInProgress to the Connect re-login flow to avoid duplicate automatic relogins.
- Gate the AuthExpiration callback in ConfigureConnectExpirationCallback so subsequent notifications return early while a reconnect is in progress.
- Clear the in progress guard when Connect.Login completes and when CreateUser completes so legitimate future retries are allowed.

### DIFF
--- a/com.playeveryware.eos/Runtime/Core/EOSManager.cs
+++ b/com.playeveryware.eos/Runtime/Core/EOSManager.cs
@@ -199,6 +199,7 @@ namespace PlayEveryWare.EpicOnlineServices
             static private NotifyEventHandle s_notifyLoginStatusChangedCallbackHandle;
             static private NotifyEventHandle s_notifyConnectLoginStatusChangedCallbackHandle;
             static private NotifyEventHandle s_notifyConnectAuthExpirationCallbackHandle;
+            static private bool s_isConnectAuthExpirationLoginInProgress;
 
             // Setting it twice will cause an exception
             static bool hasSetLoggingCallback;
@@ -1000,6 +1001,8 @@ namespace PlayEveryWare.EpicOnlineServices
                         SetLocalProductUserId(createUserCallbackInfo.LocalUserId);
                     }
 
+                    s_isConnectAuthExpirationLoginInProgress = false;
+
                     if (onCreateUserCallback != null)
                     {
                         onCreateUserCallback(createUserCallbackInfo);
@@ -1177,6 +1180,8 @@ namespace PlayEveryWare.EpicOnlineServices
                 connectInterface.Login(ref connectLoginOptions, null,
                     (ref Epic.OnlineServices.Connect.LoginCallbackInfo connectLoginData) =>
                     {
+                        s_isConnectAuthExpirationLoginInProgress = false;
+
                         if (connectLoginData.ResultCode != Result.Success)
                         {
                             Log($"Connect login was not successful. ResultCode: {connectLoginData.ResultCode}", LogType.Error);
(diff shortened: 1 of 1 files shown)


com.playeveryware.eos/Runtime/Core/EOSManager.cs
matt-clarke, line 1178: There's a very rare edge case where a developer could call this function whilst `s_isConnectAuthExpirationLoginInProgress` is true and a login attempt is ongoing.

To protect against this I would:
- Move this code into a private function
    - Either prefixed with `_` or ending in `Internal` (perhaps there are other cases in the plugin already?)
    - Add an extra bool param, which is used to indicate whether this is the re-auth attempt
    - If it's not the re-auth attempt and `s_isConnectAuthExpirationLoginInProgress` is true then call the callbacks with the `AlreadyPending` result (and log a warning explaining what happened)
- Keep this public function and have it call that private function (with false for the new param)

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -1177,6 +1180,8 @@ public void StartConnectLoginWithOptions(Epic.OnlineServices.Connect.LoginOption
                 connectInterface.Login(ref connectLoginOptions, null,
                     (ref Epic.OnlineServices.Connect.LoginCallbackInfo connectLoginData) =>
                     {
+                        s_isConnectAuthExpirationLoginInProgress = false;
```
matt-clarke, line 1183: Can you add this check? There's a convenient extension function for it in the `Epic.OnlineServices` namespace.
I've also created a thread to discuss this function in general, as I can't see it used elsewhere.

```suggestion
                        if (!connectLoginData.ResultCode.IsOperationComplete())
                        {
                            Log($"Ignoring connect login result that will be retried. ResultCode: {connectLoginData.ResultCode}");
                            return;
                        }

                        s_isConnectAuthExpirationLoginInProgress = false;
```

com.playeveryware.eos/Runtime/Core/EOSManager.cs
matt-clarke, line 992: This is a public function, so as per the comment below there's an edge case where this could be called whilst s_isConnectAuthExpirationLoginInProgress is true. This also means that it also needs to check the bool and early exit

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -1000,6 +1001,8 @@ public void CreateConnectUserWithContinuanceToken(ContinuanceToken token,
                         SetLocalProductUserId(createUserCallbackInfo.LocalUserId);
                     }
 
+                    s_isConnectAuthExpirationLoginInProgress = false;
```
matt-clarke, line 1004: Could this ever be true here? If we're re-authing then we should never be creating a user

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -1364,6 +1369,12 @@ private void ConfigureConnectExpirationCallback(Epic.OnlineServices.Connect.Logi
                     ulong callbackHandle = EOSConnectInterface.AddNotifyAuthExpiration(
                         ref addNotifyAuthExpirationOptions, null, (ref AuthExpirationCallbackInfo callbackInfo) =>
                         {
+                            if (s_isConnectAuthExpirationLoginInProgress)
+                            {
```
matt-clarke, line 1373: Can you add logging here to say that it's being ignored. It could be useful to debug an issue

matt-clarke (changes requested):

PR #1276, opened 2026-04-21
Add CallbackHelper lifecycle handling and fix unit test execution

OswaldoApexSystems:
[EOSU-1013](https://epicgames.atlassian.net/browse/EOSU-1013)
image
image

-Introduced a new CallbackHelper class to centralize notification callback lifecycle management and prevent stale callback collisions when the EOS Platform is recreated.

-Updated EOSManager and FriendTest to correctly integrate with the new callback lifecycle behavior, ensuring callbacks are cleared at the proper time during test teardown and platform shutdown.

-These changes fix failures encountered when running the full unit test suite by ensuring isolation between tests and preventing leftover EOS notification state from leaking across test runs.


Assets/Tests/PlayMode/Services/FriendsTests.cs
```diff
@@ -188,7 +189,9 @@ public IEnumerator SendReportToSelf()
             SendPlayerBehaviorReportCompleteCallbackInfo? reportResult = null;
             reportsHandle.SendPlayerBehaviorReport(ref reportOptions, null, (ref SendPlayerBehaviorReportCompleteCallbackInfo data) => { reportResult = data; });
 
-            yield return new WaitUntil(() => reportResult != null);
+            yield return new WaitUntilDone(GlobalTestTimeout, () => reportResult != null);
+
+            //yield return new WaitUntil(() => reportResult != null);
```
matt-clarke:
```suggestion
```

com.playeveryware.eos/Runtime/EOS_SDK/Core/CallbackHelper.Lifecycle.cs
matt-clarke, line 1: This file shouldn't be in the SDK source folder, we need a clear distinction between SDK source and plugin code. This is a partial class so I see why it may have been added here, but we need a different solution.
This either needs to be moved and an .asmref used or it could be added to the SDK itself if it makes sense. Let's use the thread to discuss this.

matt-clarke, line 1: Also, I want to discuss if there's another way to fix this because this seems more like a static/SDK shutdown issue

matt-clarke (changes requested):

com.playeveryware.eos/Runtime/EOS_SDK/Core/CallbackHelper.Lifecycle.cs
OswaldoApexSystems: You’re absolutely right — there should be a clear separation between SDK source and plugin/test code.
After further investigation, we confirmed that the root cause was unloading the EOS SDK in the Unity Editor, which is not recommended by the official EOS documentation. By preventing the SDK from being unloaded in Editor (EOS_DO_NOT_UNLOAD_SDK_ON_SHUTDOWN), the issue is resolved without requiring additional SDK-level lifecycle changes.
Given this, the partial class that was added to the SDK source folder is no longer necessary and will be removed. The cleanup logic will remain explicit and scoped to the unit tests, relying on the existing disposal of notification handles that already lives in the project (e.g. disposing login and connect notification handles before releasing the platform).
This avoids introducing new plugin code into the SDK source, maintains a clean separation between SDK and plugin/test responsibilities, and removes the need for a partial class + .asmref setup in this case.
Thanks for raising this — it helped narrow the solution to the correct layer.

com.playeveryware.eos/Runtime/Core/EOSManager.cs
matt-clarke, line 46: This [docs file](https://github.com/EOS-Contrib/eos_plugin_for_unity/blob/stable/com.playeveryware.eos/Documentation~/c_sharp_defines.md) covers a couple of points which are a bit of a concern:
- `"It is not the default because if it were, then any edits to the EOSConfig file would require a reboot of the Unity Editor"`
  - Can you check if this is accurate?
  - If it is then we would need to think of a solution to this, perhaps when you save the EOSConfig it prompts you to restart the Editor (after checking the new define mentioned in the comment below)
  - I suspect we also need a Unity hook to check if the file was modified outside of Unity (user could've edited it by hand, or pulled latest changes from their repo)
- `"In the past, not unloading the EOS SDK could also cause issues with state from the previous Unity run not being properly cleaned up."`
  - This is the opposite of what we're seeing right? I'm wary that there's some edge case issues that we haven't seen with this initial testing. It could definitely be a misunderstanding though
  - The [commit message](https://github.com/EOS-Contrib/eos_plugin_for_unity/commit/12cf37bf7a99473f15281369b2b70d93e4896537) where it was added also says the same, so it doesn't appear to just be a mistake in the docs (`It also
carries with it the potential for "leaked" EOS State to carry over
between play in editor runs.`)

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -36,7 +36,7 @@
 #endif
 
 // This define controls if the EOS SDK should be unloaded in the editor at shutdown to work around DLL unload errors.
-//#define EOS_DO_NOT_UNLOAD_SDK_ON_SHUTDOWN
+#define EOS_DO_NOT_UNLOAD_SDK_ON_SHUTDOWN
 
 // On macOS and Linux, there isn't a known reliable way to unload shared libraries, therefore this is the default behavior.
 #if (UNITY_EDITOR_OSX || UNITY_EDITOR_LINUX)
```
matt-clarke:
```suggestion
// As per Epic's documentation they recommend not calling PlatformInterface.Release / PlatformInterface.Shutdown in Editor
// because Unity doesn't unload managed libraries until the Editor exits. Doing so will cause unintended side effects:
// https://dev.epicgames.com/docs/epic-online-services/eos-get-started/get-started-guide/next-steps#shut-down-the-eos-sdk
//
// This is also required on macOS/Linux always, because there isn't a known reliable way to unload shared libraries
#if UNITY_EDITOR
```

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -36,7 +36,7 @@
 #endif
 
 // This define controls if the EOS SDK should be unloaded in the editor at shutdown to work around DLL unload errors.
-//#define EOS_DO_NOT_UNLOAD_SDK_ON_SHUTDOWN
+#define EOS_DO_NOT_UNLOAD_SDK_ON_SHUTDOWN
```
matt-clarke: Recommend deleting this because it's always enabled in Editor with these changes. The comment above also nicely leads into the comments below.
```suggestion
```

com.playeveryware.eos/Runtime/Core/EOSManager.cs
matt-clarke, line 47: With this change to always define `EOS_DO_NOT_UNLOAD_SDK_ON_SHUTDOWN` in Editor this means that this   [docs file](https://github.com/EOS-Contrib/eos_plugin_for_unity/blob/stable/com.playeveryware.eos/Documentation~/c_sharp_defines.md) is out of date.

The change above also removes the ability for developers to control whether it's enabled in Editor or not for Windows.

One option would be to change the above code to this, and update the docs to refer to this define instead:
```
#if UNITY_EDITOR && (!UNITY_EDITOR_WIN || !EOS_UNLOAD_SDK_ON_SHUTDOWN)
#define EOS_DO_NOT_UNLOAD_SDK_ON_SHUTDOWN
#endif
```

matt-clarke, line 1676: This appears to be the only use of `s_eosUnloadSDKOnShutdown` which isn't behind an `#if UNITY_EDITOR` check. Currently this would be false outside of the Editor if a developer adds `EOS_DO_NOT_UNLOAD_SDK_ON_SHUTDOWN` to their defines. That seems unintentional based on the plugin and Epic docs.

Added [this suggestion](https://github.com/EOS-Contrib/eos_plugin_for_unity/pull/1276/changes#r3129827722) as a way to resolve this issue

com.playeveryware.eos/Runtime/Core/EOSManager.cs
matt-clarke: See also [this comment](https://github.com/EOS-Contrib/eos_plugin_for_unity/pull/1276/changes#r3129702713) where I highlight a point about updating the `#if` check here, which would allow developers to control this. I think that's worth keeping, in case there are issues.

com.playeveryware.eos/Runtime/Core/EOSManager.cs
matt-clarke: As per [this comment](https://github.com/EOS-Contrib/eos_plugin_for_unity/pull/1276/changes#r3129800999) I think we want to make sure this is only ever false when in Editor
```suggestion
#if EOS_DO_NOT_UNLOAD_SDK_ON_SHUTDOWN && UNITY_EDITOR
```

matt-clarke: Honestly it would be nice to update these defines whilst we're here:
- `EOS_DO_NOT_UNLOAD_SDK_ON_SHUTDOWN_IN_EDITOR`
- `EOS_UNLOAD_SDK_ON_SHUTDOWN_IN_WIN_EDITOR`

matt-clarke (changes requested):

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -1640,15 +1642,29 @@ public void OnApplicationShutdown()
...
                     if (s_eosUnloadSDKOnShutdown)
                     {
                         Log("Shutting down the platform interface.");
                         ShutdownPlatformInterface();
                     }
-
+#else
+                    Log("Shutting down the platform interface.");
+                    ShutdownPlatformInterface();
+#endif
```
matt-clarke: Thanks for making that change, although this is a little cleaner
```suggestion
#if UNITY_EDITOR
                    if (s_eosUnloadSDKOnShutdown)
#endif
                    {
                        Log("Shutting down the platform interface.");
                        ShutdownPlatformInterface();
                    }
```

OswaldoApexSystems: I tried this refactor, but it causes Unity to crash.

The issue is that moving the `#if UNITY_EDITOR` above only the condition causes
`ShutdownPlatformInterface()` to execute unconditionally outside of the Editor,
without being explicitly guarded. In runtime builds this breaks the expected
SDK shutdown order and results in a native crash.

Keeping the explicit `#if UNITY_EDITOR / #else` around the entire shutdown block
ensures that:
- Editor behavior remains configurable
- Runtime builds always shut down the platform in a safe and explicit path

For this reason I’m keeping the previous structure, which is slightly more verbose
but avoids unsafe execution paths.

matt-clarke: Can you double check the changes you made to test vs my suggestion? This should be functionality identical as your code. It's compiling out the if check so the code will always run in runtime builds.

Is that not the desired functionality? I'm a bit confused because the 2nd paragraph sounds like it shouldn't always run in runtime builds, but the bullet points sounds like it should (and the code lines up with this).

OswaldoApexSystems: I double‑checked this using the suggested refactor and captured the Editor.log.
While the logic looks equivalent on paper, in practice it changes shutdown ordering.
With the refactor, EOS shutdown occurs while async operations and callbacks are still active.
The log shows EOS attempting to reuse an existing native handle (Found existing handle for EOSSDK‑Win64‑Shipping) after shutdown, which results in native crashes due to use‑after‑free / double teardown.
Keeping the explicit #if UNITY_EDITOR / #else block ensures shutdown only happens in the intended lifecycle phase and avoids teardown during Editor domain reloads and test cleanup.

[Editor.log](https://github.com/user-attachments/files/27062147/Editor.log)

matt-clarke: If it's crashing in Editor though I don't see how it could be caused by this suggestion. If you remove the `#else` section and the `#if UNITY_EDITOR` checks then the code is identical?
In both cases the `if` check is kept in Editor and the `else` will be compiled out in Editor anyway

OswaldoApexSystems: Devauthtool was not conecting, confirm that this version works

com.playeveryware.eos/Runtime/Core/EOSManager.cs
matt-clarke: Keeping this comment active because [this docs file](https://github.com/EOS-Contrib/eos_plugin_for_unity/blob/stable/com.playeveryware.eos/Documentation~/c_sharp_defines.md) needs updating with the new define

JessTello: I agree the current wording in c_sharp_defines.md is too absolute and also a bit outdated relative to the current config model. The real concern here is not simply that the config file changes, but that EOSManager now conditionally skips full SDK shutdown/unload in the Editor, which means some initialization related SDK state may persist across play sessions. Because of that, I don’t think we should claim that any config edit requires an Editor restart, a more accurate statement is that config changes which affect EOS SDK initialization may require restarting the Editor when the SDK is not fully unloaded. I also agree that the doc should stop referring to EOSConfig here and instead reflect the current ProductConfig / PlatformConfig flow. If we want better UX around this later, we can consider a targeted restart prompt or external change detection as a follow up, but I think the immediate fix is to make the documentation more precise and aligned with the current EOSManager behavior.

PR #1279, opened 2026-04-23
Eosu 404 auth token reauth p2p disconnects v3

JessTello:
Fix auth token re-auth and add P2P connection state callbacks

--Context--
The issue that motivated this PR is related to token re-authentication during the session. The logs show many EOS_InvalidAuth errors, followed by a successful UpdateUserAuthToken and then Login complete: EOS_Success.
The root cause was that the refresh flow is not calling EOS_Connect_Login with the renewed token before it expires. This suggests that the SDK is refreshing too late, or that the AddNotifyAuthExpiration callback is not being handled correctly to anticipate the expiration.
However, based on the latest comment, this seems to be a different issue. I ran tests using a VPN but could not reproduce it. In this case, the TURN/relay server assigned by EOS for the GE <> NL region could not be reached by both peers. The server reporting a NAT type of “Unknown” is a red flag, as it may be behind a symmetric NAT or have firewall restrictions that prevent TURN from working correctly.
For this second scenario, I only added extra code in EOSTransportManager. The Initialize() method was not registering AddNotifyPeerConnectionEstablished or AddNotifyPeerConnectionInterrupted. Without these callbacks, failures happen silently, with no diagnostics or recovery.

- Fix Connect re-authentication using a stale JWT when the EOS Connect token expires mid-session. The expiration callback was reusing the JWT string captured at login time, which is also expired by the time the callback fires. Now fetches a fresh JWT via CopyIdToken for EpicIdToken credentials.

- Add AddNotifyPeerConnectionEstablished and AddNotifyPeerConnectionInterrupted callbacks to EOSTransportManager and EOSPeer2PeerManager. Both managers were missing the interrupted callback entirely, and EOSPeer2PeerManager was discarding the established callback handle (preventing cleanup on logout). Connection logs now include network type (direct vs relay) to aid diagnosis of P2P connectivity issues between specific player pairs.

### DIFF
--- a/Assets/Scripts/P2PNetcodeSample/Networking/EOSTransportManager.cs
+++ b/Assets/Scripts/P2PNetcodeSample/Networking/EOSTransportManager.cs
@@ -254,6 +254,10 @@ namespace PlayEveryWare.EpicOnlineServices.Samples.Network
 
         private const byte ConnectionConfirmationChannel = byte.MaxValue;
 
+        private ulong ConnectionEstablishedNotificationsId = 0;
+        
+        private ulong ConnectionInterruptedNotificationsId = 0;
+
         [System.Diagnostics.Conditional("EOS_TRANSPORTMANAGER_DEBUG")]
         private void Log(string msg)
         {
@@ -445,10 +449,10 @@ namespace PlayEveryWare.EpicOnlineServices.Samples.Network
 
             Log($"EOSTransportManager.Shutdown: Shutting down EOSTransportManager... | EOSTransportManager={GetDebugString()}");
             CloseAllConnections();
+            UnsubscribeFromConnectionInterruptedNotifications();
+            UnsubscribeFromConnectionEstablishedNotifications();
             UnsubscribeFromConnectionClosedNotifications();
             UnsubscribeFromConnectionRequestNotifications();
-            UnsubscribeFromConnectionEstablishedNotifications();
-            UnsubscribeFromConnectionInterruptedNotifications();
             Clear();
 
 #if UNITY_EDITOR
@@ -1354,8 +1358,6 @@ namespace PlayEveryWare.EpicOnlineServices.Samples.Network
             CloseConnection(remoteUserId, socketName, true);
         }
 
-        private ulong ConnectionEstablishedNotificationsId = 0;
-
         private void SubscribeToConnectionEstablishedNotifications()
         {
             var options = new AddNotifyPeerConnectionEstablishedOptions()
@@ -1369,7 +1371,11 @@ namespace PlayEveryWare.EpicOnlineServices.Samples.Network
 
         private void UnsubscribeFromConnectionEstablishedNotifications()
         {
-            P2PHandle?.RemoveNotifyPeerConnectionEstablished(ConnectionEstablishedNotificationsId);
+            if (ConnectionEstablishedNotificationsId != 0)
+            {
+                P2PHandle?.RemoveNotifyPeerConnectionEstablished(ConnectionEstablishedNotificationsId);
+                ConnectionEstablishedNotificationsId = 0;
+            }
         }
 
         private void OnConnectionEstablishedNotification(ref OnPeerConnectionEstablishedInfo data)
@@ -1378,8 +1384,6 @@ namespace PlayEveryWare.EpicOnlineServices.Samples.Network
             Log($"EOSTransportManager.OnConnectionEstablishedNotification: Connection established with remote peer '{LoggingUtils.Redact(data.RemoteUserId)}' " +
                 $"on socket '{data.SocketId?.SocketName}' | type={data.ConnectionType} network={data.NetworkType}");
         }
-
-        private ulong ConnectionInterruptedNotificationsId = 0;
 
         private void SubscribeToConnectionInterruptedNotifications()
         {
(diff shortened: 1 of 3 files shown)


Assets/Scripts/P2PNetcodeSample/Networking/EOSTransportManager.cs
```diff
@@ -1350,6 +1354,56 @@ private void OnConnectionClosedNotification(ref OnRemoteConnectionClosedInfo dat
...
+                LocalUserId = LocalUserId,
+                SocketId = null,
+            };
+
+            ConnectionEstablishedNotificationsId = P2PHandle.AddNotifyPeerConnectionEstablished(ref options, null, OnConnectionEstablishedNotification);
+        }
+
+        private void UnsubscribeFromConnectionEstablishedNotifications()
+        {
+            P2PHandle?.RemoveNotifyPeerConnectionEstablished(ConnectionEstablishedNotificationsId);
```
matt-clarke:
```suggestion
            if (ConnectionEstablishedNotificationsId != 0)
            {
                P2PHandle?.RemoveNotifyPeerConnectionEstablished(ConnectionEstablishedNotificationsId);
                ConnectionEstablishedNotificationsId = 0;
            }
```

Assets/Scripts/P2PNetcodeSample/Networking/EOSTransportManager.cs
```diff
@@ -1350,6 +1354,56 @@ private void OnConnectionClosedNotification(ref OnRemoteConnectionClosedInfo dat
             CloseConnection(remoteUserId, socketName, true);
         }
 
+        private ulong ConnectionEstablishedNotificationsId = 0;
```
matt-clarke: Please move this to be with the other vars at the top of the file

Assets/Scripts/P2PNetcodeSample/Networking/EOSTransportManager.cs
```diff
@@ -1350,6 +1354,56 @@ private void OnConnectionClosedNotification(ref OnRemoteConnectionClosedInfo dat
...
+        }
+
+        private void OnConnectionEstablishedNotification(ref OnPeerConnectionEstablishedInfo data)
+        {
+            Debug.Assert(data.LocalUserId == LocalUserId);
+            Log($"EOSTransportManager.OnConnectionEstablishedNotification: Connection established with remote peer '{LoggingUtils.Redact(data.RemoteUserId)}' " +
+                $"on socket '{data.SocketId?.SocketName}' | type={data.ConnectionType} network={data.NetworkType}");
+        }
+
+        private ulong ConnectionInterruptedNotificationsId = 0;
```
matt-clarke: Please move this to be with the other vars at the top of the file

Assets/Scripts/P2PNetcodeSample/Networking/EOSTransportManager.cs
```diff
@@ -1350,6 +1354,56 @@ private void OnConnectionClosedNotification(ref OnRemoteConnectionClosedInfo dat
...
+                LocalUserId = LocalUserId,
+                SocketId = null,
+            };
+
+            ConnectionInterruptedNotificationsId = P2PHandle.AddNotifyPeerConnectionInterrupted(ref options, null, OnConnectionInterruptedNotification);
+        }
+
+        private void UnsubscribeFromConnectionInterruptedNotifications()
+        {
+            P2PHandle?.RemoveNotifyPeerConnectionInterrupted(ConnectionInterruptedNotificationsId);
```
matt-clarke:
```suggestion
            if (ConnectionInterruptedNotificationsId != 0)
            {
                P2PHandle?.RemoveNotifyPeerConnectionInterrupted(ConnectionInterruptedNotificationsId);
                ConnectionInterruptedNotificationsId = 0;
            }
```

Assets/Scripts/P2PNetcodeSample/Networking/EOSTransportManager.cs
```diff
@@ -445,6 +447,8 @@ public void Shutdown()
             CloseAllConnections();
             UnsubscribeFromConnectionClosedNotifications();
             UnsubscribeFromConnectionRequestNotifications();
+            UnsubscribeFromConnectionEstablishedNotifications();
+            UnsubscribeFromConnectionInterruptedNotifications();
```
matt-clarke: Minor: This code is unsubscribing in reverse order to the order they were subscribed in. It's not an issue but it's a good practice
```suggestion
            UnsubscribeFromConnectionInterruptedNotifications();
            UnsubscribeFromConnectionEstablishedNotifications();
            UnsubscribeFromConnectionClosedNotifications();
            UnsubscribeFromConnectionRequestNotifications();
```

matt-clarke (changes requested):

Assets/Scripts/P2PNetcodeSample/Networking/EOSTransportManager.cs
JessTello: Fixed

Assets/Scripts/P2PNetcodeSample/Networking/EOSTransportManager.cs
JessTello: Fixed

Assets/Scripts/P2PNetcodeSample/Networking/EOSTransportManager.cs
JessTello: Fixed

Assets/Scripts/P2PNetcodeSample/Networking/EOSTransportManager.cs
JessTello: Fixed

Assets/Scripts/P2PNetcodeSample/Networking/EOSTransportManager.cs
JessTello: Fixed

matt-clarke (approved):

billfr-ApexSystems (approved):

PR #1287, opened 2026-05-04
Support for windows ARM64

JessTello:


Assets/Plugins/Windows/ARM64/xaudio2_9redist.dll.meta
```diff
@@ -18,11 +18,11 @@ PluginImporter:
...
         Exclude Android: 1
         Exclude Editor: 1
-        Exclude Linux64: 1
-        Exclude OSXUniversal: 1
+        Exclude Linux64: 0
+        Exclude OSXUniversal: 0
         Exclude PS4: 1
         Exclude Switch: 1
-        Exclude Win: 1
+        Exclude Win: 0
```
matt-clarke: This is enabling this library for Win, Linux, Mac. Mac/Linux should just ignore it and I assume Win is 32bit, but best to fix this

com.playeveryware.eos/Runtime/Core/EOSManager_DynamicLoading.cs
```diff
@@ -58,14 +58,14 @@ public partial class EOSSingleton
...
 #if UNITY_STANDALONE_OSX
                 "GfxPluginNativeRender-macOS";
+#elif UNITY_STANDALONE_WIN && EOS_PLATFORM_WINDOWS_ARM64
+    "GfxPluginNativeRender-ARM64";
+#elif UNITY_STANDALONE_WIN && (EOS_PLATFORM_WINDOWS_64 || UNITY_64)
+    "GfxPluginNativeRender-x64";
 #elif UNITY_STANDALONE_WIN
-#if UNITY_64
-                "GfxPluginNativeRender-x64";
+    "GfxPluginNativeRender-x86";
```
matt-clarke: Fixing indent and removing 32bit check (if there's a handful of 32bit things remaining it probably makes sense to do that separately)

```suggestion
#elif UNITY_STANDALONE_WIN && EOS_PLATFORM_WINDOWS_ARM64
                "GfxPluginNativeRender-ARM64";
#elif UNITY_STANDALONE_WIN
                "GfxPluginNativeRender-x64";
```

lib/NativeCode/DynamicLibraryLoaderHelper/ManagedToUnmanagedBridge/ManagedToUnmanagedBridge.vcxproj
```diff
@@ -44,32 +60,33 @@
   <ImportGroup Label="PropertySheets" Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
     <Import Project="$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props" Condition="exists('$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props')" Label="LocalAppDataPlatform" />
   </ImportGroup>
+  <ImportGroup Label="PropertySheets" Condition="'$(Configuration)|$(Platform)'=='Debug|ARM64'">
+    <Import Project="$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props" Condition="exists('$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props')" Label="LocalAppDataPlatform" />
+  </ImportGroup>
+  <ImportGroup Label="PropertySheets" Condition="'$(Configuration)|$(Platform)'=='Release|ARM64'">
+    <Import Project="$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props" Condition="exists('$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props')" Label="LocalAppDataPlatform" />
+  </ImportGroup>
```
matt-clarke: The condition for all "PropertySheets" could just be: `"'$(Platform)'=='x64' Or '$(Platform)'=='ARM64'"`

lib/NativeCode/DynamicLibraryLoaderHelper/ManagedToUnmanagedBridge/ManagedToUnmanagedBridge.vcxproj
```diff
@@ -44,32 +60,33 @@
...
   <PropertyGroup Label="UserMacros" />
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64' Or '$(Configuration)|$(Platform)'=='Release|x64' Or '$(Configuration)|$(Platform)'=='Debug|ARM64' Or '$(Configuration)|$(Platform)'=='Release|ARM64'">
```
matt-clarke:
```suggestion
  <PropertyGroup Condition="'$(Platform)'=='x64' Or '$(Platform)'=='ARM64'">
```

lib/NativeCode/DynamicLibraryLoaderHelper/ManagedToUnmanagedBridge/ManagedToUnmanagedBridge.vcxproj
```diff
@@ -44,32 +60,33 @@
...
-  <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
+  <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64' Or '$(Configuration)|$(Platform)'=='Debug|ARM64'">
     <ClCompile>
       <PrecompiledHeader>Use</PrecompiledHeader>
       <PrecompiledHeaderFile>pch.h</PrecompiledHeaderFile>
       <WarningLevel>Level3</WarningLevel>
       <PreprocessorDefinitions>_DEBUG;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <AdditionalUsingDirectories>$(SolutionDir)Build\$(Configuration)\$(Platform)\</AdditionalUsingDirectories>
-      <LanguageStandard>stdcpp20</LanguageStandard>
+      <LanguageStandard>stdcpp17</LanguageStandard>
```
matt-clarke: Was this a deliberate change?

lib/NativeCode/DynamicLibraryLoaderHelper/ConsoleApplication/ConsoleApplication.vcxproj
matt-clarke: x64 is using v142 for release and v143 by the looks of this. Is that a quick fix to change here, or does a bug need adding to track/resolve this?

lib/NativeCode/DynamicLibraryLoaderHelper/ConsoleApplication/ConsoleApplication.vcxproj
```diff
@@ -34,6 +42,14 @@
     <WholeProgramOptimization>true</WholeProgramOptimization>
     <CharacterSet>Unicode</CharacterSet>
   </PropertyGroup>
+  <!-- ARM64 Release uses v143 toolset (v142 lacks ARM64 cross-compile support). -->
```
matt-clarke: As per comment above, if everything is changed to use v143 then this comment can probably be removed (as ARM64 isn't an exception in that case)

lib/NativeCode/DynamicLibraryLoaderHelper/ConsoleApplication/ConsoleApplication.vcxproj
```diff
@@ -45,26 +61,26 @@
   <ImportGroup Label="PropertySheets" Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
     <Import Project="$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props" Condition="exists('$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props')" Label="LocalAppDataPlatform" />
   </ImportGroup>
+  <ImportGroup Label="PropertySheets" Condition="'$(Configuration)|$(Platform)'=='Debug|ARM64'">
+    <Import Project="$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props" Condition="exists('$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props')" Label="LocalAppDataPlatform" />
+  </ImportGroup>
+  <ImportGroup Label="PropertySheets" Condition="'$(Configuration)|$(Platform)'=='Release|ARM64'">
+    <Import Project="$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props" Condition="exists('$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props')" Label="LocalAppDataPlatform" />
+  </ImportGroup>
```
matt-clarke: All the "PropertySheets" nodes have the same body?
The condition could just be: `"'$(Platform)'=='x64' Or '$(Platform)'=='ARM64'"`

(github is refusing to let me use the suggestion feature because it covers a few lines that weren't touched by the PR 😩)

lib/NativeCode/DynamicLibraryLoaderHelper/ConsoleApplication/ConsoleApplication.vcxproj
```diff
@@ -45,26 +61,26 @@
...
   <PropertyGroup Label="UserMacros" />
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
-    <OutDir>$(SolutionDir)Build\$(Configuration)\$(PlatformTarget)\</OutDir>
-    <IntDir>$(SolutionDir)Build\Temp\$(ProjectName)\$(Configuration)\$(PlatformTarget)\</IntDir>
-    <LibraryPath>$(SolutionDir)Build\$(Configuration)\$(Platform)\;$(LibraryPath)</LibraryPath>
-    <TargetName>$(ProjectName)-$(PlatformTarget)</TargetName>
-  </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64' Or '$(Configuration)|$(Platform)'=='Release|x64' Or '$(Configuration)|$(Platform)'=='Debug|ARM64' Or '$(Configuration)|$(Platform)'=='Release|ARM64'">
```
matt-clarke:
```suggestion
  <PropertyGroup Condition="'$(Platform)'=='x64' Or '$(Platform)'=='ARM64'">
```

lib/NativeCode/DynamicLibraryLoaderHelper/ConsoleApplication/ConsoleApplication.vcxproj
```diff
@@ -45,26 +61,26 @@
...
   </PropertyGroup>
-  <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
+  <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64' Or '$(Configuration)|$(Platform)'=='Debug|ARM64'">
     <ClCompile>
       <WarningLevel>Level3</WarningLevel>
       <SDLCheck>true</SDLCheck>
       <PreprocessorDefinitions>_DEBUG;_CONSOLE;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <ConformanceMode>true</ConformanceMode>
-      <LanguageStandard>stdcpp20</LanguageStandard>
+      <LanguageStandard>stdcpp17</LanguageStandard>
```
matt-clarke: Needs reverting?

Assets/Plugins/Source/Editor/Platforms/Windows/WindowsBuilder.cs
```diff
@@ -35,9 +35,158 @@ namespace PlayEveryWare.EpicOnlineServices.Editor.Build
...
+    /// of <see cref="BuildTarget.StandaloneWindows64"/>. Unity does not provide a built-in
+    /// scripting define to distinguish ARM64 from x64 on Standalone Windows, so this
+    /// project-local symbol is auto-managed by the Windows builders during preprocess.
+    ///
+    /// IMPORTANT: scripting defines set during build preprocess only affect the *next*
+    /// build's script compilation, not the current one. Switching between x64 and ARM64
+    /// for Windows builds requires running the menu item under "EOS Plugin/Advanced/Windows
+    /// ARM64..." once before the first build, or running a build twice (the first to set
+    /// the define, the second to compile against it).
+    /// </summary>
```
matt-clarke, line 52: Have you been able to test what happens on first launch of running in Editor on an ARM64 machine?
The code elsewhere relies on `EOS_PLATFORM_WINDOWS_ARM64` being defined, but I can only see that being defined when you make a build.

Assets/Plugins/Source/Editor/Platforms/Windows/WindowsBuilder.cs
```diff
@@ -35,9 +35,158 @@ namespace PlayEveryWare.EpicOnlineServices.Editor.Build
...
+            {
+                NamedBuildTarget named = NamedBuildTarget.FromBuildTargetGroup(BuildTargetGroup.Standalone);
+                string defines = PlayerSettings.GetScriptingDefineSymbols(named);
+                return !string.IsNullOrEmpty(defines)
+                    && System.Array.IndexOf(defines.Split(';'), Symbol) >= 0;
+            }
+            catch
+            {
+                return false;
+            }
```
matt-clarke, line 105: In what situations do the two try/catch blocks above throw? I'm wary of hiding an issue by not handling/logging at all

Assets/Plugins/Source/Editor/Platforms/Windows/WindowsBuilder.cs
```diff
@@ -48,8 +197,93 @@ public WindowsBuilder64() : base("Plugins/Windows/x64", BuildTarget.StandaloneWi
...
+                : "Targeting Windows x64 (Intel/AMD)");
+#endif
+            // Safety net for scripted/CI builds that bypass RegisterBuildPlayerHandler.
+            if (WindowsArm64Define.IsDefineSet())
+            {
+                ScriptingDefineUtility.RemoveDefine(BuildTarget.StandaloneWindows64, WindowsArm64Define.Symbol);
+                throw new BuildFailedException(
+                    $"Scripting define '{WindowsArm64Define.Symbol}' was active from a previous ARM64 build. " +
+                    "It has been removed. Please rebuild to compile with Windows x64 support.");
+            }
```
matt-clarke, line 249: There's some possible better solutions to this issue:
- `BuildPlayerOptions.extraScriptingDefines` allows you to set symbols which are used for the current build. This way they don't persist between builds
- You may be able to use `CompilationPipeline.RequestScriptCompilation` to force it to compile and then continue with the build instead. I say _may_ because I think the function is async and you need to use the `CompilationPipeline.compilationFinished` event. We used this in Fall Guys before

If the first option is simple to try and works then that would be ideal. Beyond that I think we need a follow up task to revisit this

Assets/Plugins/Source/Editor/Platforms/Windows/WindowsBuilder.cs
matt-clarke, line 52: Also, is there a task for adding docs to explain these menu options and how to run in Editor for ARM64?

matt-clarke (dismissed):

Assets/Plugins/Source/Editor/Platforms/Windows/WindowsBuilder.cs
```diff
@@ -35,9 +35,158 @@ namespace PlayEveryWare.EpicOnlineServices.Editor.Build
...
+            ScriptingDefineUtility.AddDefine(BuildTarget.StandaloneWindows64, Symbol);
+            UnityEngine.Debug.Log($"Scripting define '{Symbol}' enabled for Standalone Windows. Build for ARM64 architecture to produce ARM64 binaries.");
+        }
+
+        [MenuItem("EOS Plugin/Advanced/Windows ARM64/Disable scripting define")]
+        private static void DisableArm64Define()
+        {
+            ScriptingDefineUtility.RemoveDefine(BuildTarget.StandaloneWindows64, Symbol);
+            UnityEngine.Debug.Log($"Scripting define '{Symbol}' disabled for Standalone Windows. Subsequent builds will produce x64 binaries.");
+        }
```
matt-clarke, line 201: I thinks it's worth having the same comment here about needing these as a workaround that can be used before making the build to ensure it works first time (although I hope that can be resolved, as below)

Assets/Plugins/Source/Editor/Utility/BuildUtility.cs
```diff
@@ -401,13 +405,11 @@ private static bool BuildFromSolutionFile(string solutionFilePath, string binary
             }
 
             // construct the msbuild command
+            string platformFlag = string.IsNullOrEmpty(platform) ? "" : $" /p:Platform={platform}";
```
matt-clarke:
```suggestion
            string platformFlag = string.IsNullOrWhiteSpace(platform) ? "" : $" /p:Platform={platform}";
```

com.playeveryware.eos/Runtime/EOS_SDK/Generated/Windows/Common.cs
```diff
@@ -11,7 +13,7 @@
 	#endif
 #endif
 
-#if EOS_PLATFORM_WINDOWS_32 || EOS_PLATFORM_WINDOWS_64
+#if EOS_PLATFORM_WINDOWS_32 || EOS_PLATFORM_WINDOWS_64 || EOS_PLATFORM_WINDOWS_ARM64
```
matt-clarke, line 16: Will create a thread to chat about this edit and if it can be brought into the C# SDK, I see why you had to do it

matt-clarke: Can you make sure this change is included in the document that covers the edits that you have to make to the C# SDK?

matt-clarke: Actually, is this edit required? Without this it should still result in the same code because `UNITY_EDITOR_WIN` and/or `UNITY_STANDALONE_WIN` is still defined for ARM64?

lib/NativeCode/DynamicLibraryLoaderHelper/ConsoleApplication/ConsoleApplication.vcxproj
JessTello: Quick fix since the project now builds against C++20, v142 is insufficient (incomplete C++20 support). Changed Release x64 to v143 in both projects to match all other configurations. Removed the now-redundant ARM64 comment since v143 is uniform across all platforms and configurations.

Assets/Plugins/Windows/ARM64/xaudio2_9redist.dll.meta
JessTello: Exclude Win is back to 1 and Standalone: Win set to enabled: 0.

Also found that both Windows/x64/ and Windows/ARM64/ have a xaudio2_9redist.dll with the same filename, which causes a conflict when Unity includes both for a Win64 build. Fixed with defineConstraints on both meta files:
ARM64 DLL → defineConstraints: [EOS_PLATFORM_WINDOWS_ARM64]
x64 DLL → defineConstraints: [!EOS_PLATFORM_WINDOWS_ARM64]
This also handles Unity 2021.3 correctly since that version never defines EOS_PLATFORM_WINDOWS_ARM64, the ARM64 DLL is transparently ignored at all times.

com.playeveryware.eos/Runtime/Core/EOSManager_DynamicLoading.cs
JessTello: Applied. Removed the redundant EOS_PLATFORM_WINDOWS_64 || UNITY_64 branch and the x86 fallback. The path now resolves to ARM64 when EOS_PLATFORM_WINDOWS_ARM64 is defined, and defaults to x64 for all other Windows builds. 32-bit will be tracked separately if needed.

lib/NativeCode/DynamicLibraryLoaderHelper/ManagedToUnmanagedBridge/ManagedToUnmanagedBridge.vcxproj
JessTello: Applied. Consolidated all four PropertySheets ImportGroups into a single one with Condition="'$(Platform)'=='x64' Or '$(Platform)'=='ARM64'". Same simplification applied to ConsoleApplication.vcxproj.

lib/NativeCode/DynamicLibraryLoaderHelper/ManagedToUnmanagedBridge/ManagedToUnmanagedBridge.vcxproj
JessTello: Applied. Simplified the PropertyGroup condition from four Configuration|Platform combinations down to '$(Platform)'=='x64' Or '$(Platform)'=='ARM64'. Same change applied to ConsoleApplication.vcxproj.

lib/NativeCode/DynamicLibraryLoaderHelper/ManagedToUnmanagedBridge/ManagedToUnmanagedBridge.vcxproj
JessTello: It was an accidental regression when reorganizing the ItemDefinitionGroups to add ARM64 support. The project builds against C++20, so both Debug and Release configs are back to stdcpp20. Fixed in both ManagedToUnmanagedBridge.vcxproj and ConsoleApplication.vcxproj.

lib/NativeCode/DynamicLibraryLoaderHelper/ConsoleApplication/ConsoleApplication.vcxproj
JessTello: Done. all configurations now use v143, so the comment no longer applies and has been removed from both ConsoleApplication.vcxproj and ManagedToUnmanagedBridge.vcxproj.

lib/NativeCode/DynamicLibraryLoaderHelper/ConsoleApplication/ConsoleApplication.vcxproj
JessTello: Applied consolidated all four PropertySheets ImportGroups (including the pre-existing x64 ones) into a single Condition="'$(Platform)'=='x64' Or '$(Platform)'=='ARM64'". Same done for ManagedToUnmanagedBridge.vcxproj.

lib/NativeCode/DynamicLibraryLoaderHelper/ConsoleApplication/ConsoleApplication.vcxproj
JessTello: Applied, same simplification, the PropertyGroup condition

Assets/Plugins/Source/Editor/Platforms/Windows/WindowsBuilder.cs
JessTello: Unity Editor runs in x64 emulation mode, so EOS_PLATFORM_WINDOWS_ARM64 intentionally does NOT get defined there. UNITY_EDITOR_WIN resolves to EOS_PLATFORM_WINDOWS_64 in Common.cs, which is the correct path for the Editor. The x64 native plugins load via emulation and work as expected.

EOS_PLATFORM_WINDOWS_ARM64 is only relevant for Standalone ARM64 builds, not for Editor use. The ARM64 DLL meta files also already have Exclude Editor: 1 to reinforce this.

Assets/Plugins/Source/Editor/Platforms/Windows/WindowsBuilder.cs
JessTello: Good point. . Updated both catches to log a Debug.LogWarning with the exception message so any unexpected failure is visible.

Assets/Plugins/Source/Editor/Platforms/Windows/WindowsBuilder.cs
JessTello: Good suggestions. Investigated BuildPlayerOptions.extraScriptingDefines it works for the "missing define" case (ARM64 build without the define set): we can inject it into options.extraScriptingDefines and proceed without stopping the build. However it only adds symbols, so it can't help the reverse case (x64 build with a leftover ARM64 define), which still needs the define removed and a recompile.CompilationPipeline.RequestScriptCompilation + compilationFinished would handle both cases cleanly but is non-trivial to wire up correctly, especially for CI builds.
Proposal: apply extraScriptingDefines for the add-define path in this PR as a partial improvement, and track the full solution (both directions) as a follow-up task. Does that work for you?

Assets/Plugins/Source/Editor/Platforms/Windows/WindowsBuilder.cs
JessTello: Not yet will create a task to document the ARM64 menu options and the Editor workflow for ARM64 Windows.

Assets/Plugins/Source/Editor/Platforms/Windows/WindowsBuilder.cs
JessTello: Added the same workaround context to both EnableArm64Define and DisableArm64Define

Assets/Plugins/Source/Editor/Utility/BuildUtility.cs
JessTello: Applied.

com.playeveryware.eos/Runtime/EOS_SDK/Generated/Windows/Common.cs
JessTello: The edit is required. When EOS_PLATFORM_WINDOWS_ARM64 is set, the #if EOS_PLATFORM_WINDOWS_ARM64 branch at the top of the file is taken, which means the #elif UNITY_STANDALONE_WIN branch is never reached so EOS_PLATFORM_WINDOWS_64 is not defined. Without adding EOS_PLATFORM_WINDOWS_ARM64 to the guard at the bottom, the entire Common class body gets compiled out for ARM64 builds, causing missing type errors.
UNITY_STANDALONE_WIN is still defined by Unity, but the #elif chain short-circuits before it can set EOS_PLATFORM_WINDOWS_64.
Will make sure this is documented in the C# SDK edits document.

Assets/Plugins/Source/Editor/Platforms/Windows/WindowsBuilder.cs
```diff
@@ -48,7 +213,92 @@ public WindowsBuilder64() : base("Plugins/Windows/x64", BuildTarget.StandaloneWi
...
+        {
+            AddProjectFileToBinaryMapping(
+                "DynamicLibraryLoaderHelper/DynamicLibraryLoaderHelper.sln",
+                "DynamicLibraryLoaderHelper-ARM64.dll",
+                "GfxPluginNativeRender-ARM64.dll");
+        }
+
+        public override string GetPlatformString()
+        {
+            return "ARM64";
```
EpicCliffHudson: This string appears more than once in this file - if it's supposed to be consistent, it should be a constant somewhere.

Assets/Plugins/Source/Editor/Platforms/Windows/WindowsBuilder.cs
```diff
@@ -65,6 +315,11 @@ public WindowsBuilder32() : base("Plugins/Windows/x86", BuildTarget.StandaloneWi
                 "DynamicLibraryLoaderHelper-x86.dll",
                 "GfxPluginNativeRender-x86.dll");
         }
+
+        public override string GetPlatformString()
+        {
+            return "Win32";
```
EpicCliffHudson: Does this need to match somewhere else?  Should probably be a constant define somewhere if it's shared.

EpicCliffHudson: Same for x64

Assets/Plugins/Windows/Core/WindowsPlatformSpecifics.cs
```diff
@@ -128,7 +135,14 @@ public override void ConfigureSystemInitOptions(ref EOSInitializeOptions initial
         /// <param name="createOptions"></param>
         public override void ConfigureSystemPlatformCreateOptions(ref EOSCreateOptions createOptions)
         {
-            const string pluginPlatformPath ="x64";
+            // EOS_PLATFORM_WINDOWS_ARM64 is set by the Windows ARM64 build pipeline.
+            // xaudio2_9redist.dll ships in both Plugins/Windows/x64/ and Plugins/Windows/ARM64/
+            // — we must point at the architecture-matching copy at runtime.
+#if EOS_PLATFORM_WINDOWS_ARM64
+            const string pluginPlatformPath = "ARM64";
```
EpicCliffHudson, line 142: Can these be shared across modules?  They appear multiple times in this PR.

EpicCliffHudson (changes requested):

Assets/Plugins/Windows/Core/WindowsPlatformSpecifics.cs
JessTello: WindowsPlatformSpecifics is runtime code (compiled into the player), the builder classes live in the editor only assembly. There's no shared assembly that spans both contexts, so a single constant definition isn't possible without introducing a new shared assembly just for two string values which would be over engineering I think.... Let me know what do you think.

Assets/Plugins/Source/Editor/Platforms/Windows/WindowsBuilder.cs
JessTello: Applied

Assets/Plugins/Source/Editor/Platforms/Windows/WindowsBuilder.cs
JessTello: Applied

EpicCliffHudson (approved):

billfr-ApexSystems (approved):

PR #1297, opened 2026-06-11
Fix: Prevent FriendsListToggle null reference after scene changes

JessTello:
- Clear static Friends tab UI callbacks on scene unload to avoid retaining old listeners.
- Safely invoke Friends tab expand/collapse callbacks, skipping destroyed Unity targets.
- Guard optional UIFriendInteractionSource usage to prevent calls into missing or destroyed references during UI refresh.


Assets/Scripts/StandardSamples/UI/Lobbies/UIActions.cs
```diff
@@ -23,10 +23,60 @@
...
+    using UnityEngine.SceneManagement;
 
     public static class UIActions
     {
         public static Action OnCollapseFriendsTab;
         public static Action OnExpandFriendsTab;
+
+        public static void InvokeOnCollapseFriendsTab()
+        {
+            InvokeFriendsTabAction(ref OnCollapseFriendsTab);
```
matt-clarke: Just want to make the function names here a bit more generic
```suggestion
            SafeInvokeAction(ref OnCollapseFriendsTab);
```

Assets/Scripts/StandardSamples/UI/Lobbies/UIActions.cs
```diff
@@ -23,10 +23,60 @@
...
+        }
+
+        public static void InvokeOnExpandFriendsTab()
+        {
+            InvokeFriendsTabAction(ref OnExpandFriendsTab);
```
matt-clarke:
```suggestion
            SafeInvokeAction(ref OnExpandFriendsTab);
```

Assets/Scripts/StandardSamples/UI/Lobbies/UIActions.cs
```diff
@@ -23,10 +23,60 @@
...
+        }
+
+        [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.SubsystemRegistration)]
+        private static void Initialize()
+        {
+            ClearFriendsTabActions();
```
matt-clarke:
```suggestion
            ClearActions();
```

Assets/Scripts/StandardSamples/UI/Lobbies/UIActions.cs
```diff
@@ -23,10 +23,60 @@
...
+            SceneManager.sceneUnloaded -= OnSceneUnloaded;
+            SceneManager.sceneUnloaded += OnSceneUnloaded;
+        }
+
+        private static void OnSceneUnloaded(Scene scene)
+        {
+            ClearFriendsTabActions();
```
matt-clarke:
```suggestion
            ClearActions();
```

Assets/Scripts/StandardSamples/UI/Lobbies/UIActions.cs
```diff
@@ -23,10 +23,60 @@
...
+        }
+
+        private static void ClearFriendsTabActions()
```
matt-clarke:
```suggestion
        private static void ClearActions()
```

Assets/Scripts/StandardSamples/UI/Lobbies/UIActions.cs
```diff
@@ -23,10 +23,60 @@
...
+        {
+            OnCollapseFriendsTab = null;
+            OnExpandFriendsTab = null;
+        }
+
+        private static void InvokeFriendsTabAction(ref Action action)
```
matt-clarke:
```suggestion
        private static void SafeInvokeAction(ref Action action)
```

Assets/Scripts/StandardSamples/UI/Lobbies/UIActions.cs
```diff
@@ -23,10 +23,60 @@
...
+        {
+            InvokeFriendsTabAction(ref OnExpandFriendsTab);
+        }
+
+        [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.SubsystemRegistration)]
+        private static void Initialize()
+        {
+            ClearFriendsTabActions();
+            SceneManager.sceneUnloaded -= OnSceneUnloaded;
+            SceneManager.sceneUnloaded += OnSceneUnloaded;
```
matt-clarke: Are we able to remove this and fully rely on the null checking in the invoke function?

Clearing all the events when a scene is unloaded might not be a problem at the moment but there are cases where the subscription needs to survive:
- Objects marked with DontDestroyOnLoad
- Code that doesn't belong to a Unity object (e.g. static functions)

If multiple scenes are loaded at once at any point then when a scene unloads the actions are nulled too.

matt-clarke (changes requested):

ricardoamores-epic (dismissed):
LGTM

Assets/Scripts/StandardSamples/UI/Lobbies/UIActions.cs
Pong-Epic: I've removed the sceneUnloaded event listener. It will still work as you suggested, since the null checking is handled in the safe invoke function. That said, I thought it might be useful in the future for Fast Enter Play Mode, so I also tested this. `Initialize()` under `RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.SubsystemRegistration)` still gets called in Unity 6.6b with no Domain or Scene Reload.

matt-clarke (dismissed):

Assets/Scripts/StandardSamples/UI/Lobbies/UIActions.cs
```diff
@@ -23,10 +23,52 @@
...
+        private static void SafeInvokeAction(ref Action action)
+        {
+            if (action == null)
+            {
+                return;
+            }
+
+            foreach (Action handler in action.GetInvocationList())
+            {
+                if (handler.Target is UnityEngine.Object target && target == null)
```
ricardoamores-epic: Can we get a comment here on why is this needed? Why a handler could ever be null?

Pong-Epic: This is the actually null check fix, the handler.Target became null because the GameObject have been destroyed when scene change (from "Lobbies" to other), and the action may still referencing that destroyed object here so NRE before the fix

and here cannot just directly `if (handle.Target == null)` because handle.Target just plain System.object and have to become a UnityEngine.Object to check the destroyed object as `null` here.

Pong-Epic: So maybe there a bit context : 

This fix is likely null check guard before Invoke a function, and remove the reference point found it became null (got destroyed elsewhere), that mean the subscribe/unsubscribe to this Action isn't clean up completed in some flow.

matt-clarke: So there's two options here I think:
1. We remove this check and wherever we find that code isn't unsubscribing properly we fix that
2. We add a comment to the code saying why this check exists (as Pong says due to the way that Unity marks GameObjects for destruction)

I'm leaning towards option 1, it leads to better code because you're aware of when an event hasn't been unsubscribed from properly

Pong-Epic: I think that doable, I would check the unsubscribing logic where is missing as well check how bad is it, I remember I may noted when report to Apex before.

Pong-Epic:
Finially fixed the add/remove listener in correct pair in OnEnable/OnDisable

Assets/Scripts/StandardSamples/UI/Lobbies/UILobbiesMenu.cs
```diff
@@ -97,13 +97,22 @@ public class UILobbiesMenu : SampleMenuWithFriends
 
         protected override void OnEnable()
         {
-            base.OnEnable();
+            // We add the listener to the Action first in OnEnable. Then, if StartsHidden = true, it will be set to Disabled to remove the listener.
             UIActions.OnCollapseFriendsTab += EnableInterferingUIForFriendsTab;
             UIActions.OnExpandFriendsTab += DisableInterferingUIForFriendsTab;
+
+            base.OnEnable();
```
matt-clarke: Does this work? Looking at the base class it looks like it will work when it's hidden at the start, but I didn't look into when it's shown later.

```suggestion
            base.OnEnable();

            if (!Hidden)
            {
                UIActions.OnCollapseFriendsTab += EnableInterferingUIForFriendsTab;
                UIActions.OnExpandFriendsTab += DisableInterferingUIForFriendsTab;
            }
```

Pong-Epic: Checking Hidden not work when It's shown, the flag update after [SetActive(true)](https://github.com/EOS-Contrib/eos_plugin_for_unity/blob/c31b02affc1c736d850fbd3c32809729c2137b3a/Assets/Scripts/StandardSamples/UI/Common/SampleMenu.cs#L270), so I changed to the actually Unity activeSelf for that gameObject seem suitable for this case

matt-clarke (dismissed):

ricardoamores-epic (approved):

matt-clarke (approved):

PR #1298, opened 2026-06-11
Fix: Stabilize unit tests

JessTello:
Fix EOS test crashes and stabilize test setup
Fix config path resolution and copy test config to build
Fix prefabs and performance stress test for missing scripts.
Fix asmdef for correct platforms


Assets/Tests/Common/UnitTestConfig.cs
```diff
@@ -59,7 +61,7 @@ public partial class UnitTestConfig : Config
         protected UnitTestConfig() : base(
             "eos_automated_test_config.json",
             FileSystemUtility.CombinePaths(
-                FileSystemUtility.GetProjectPath(),
+                Path.GetFullPath(Path.Combine(Application.dataPath, "..")),
```
matt-clarke, line 64: This should be functionally equivalent shouldn't it? The GetProjectPath function will find the Assets folder and step up one directory too

Assets/Tests/PlayMode/Services/ClientLobbyTests.cs
```diff
@@ -42,6 +42,10 @@ public partial class ClientLobbyTests : EOSTestBase
         [SetUp]
         public void Initialize()
         {
+            Assert.Ignore(
+                "This test requires a second machine running as a server. " +
+                "Run only the 'Client' category when a server player is available.");
```
matt-clarke, line 47: If you run the 'Client' category this test will still be ignored though right?

I'm not sure of an ideal resolution right now, but this seems passable for now:
- Add a new bool to the test config ("RunLobbyTestsThatRequireHost")
- Update this code to check that and use Assert.Ignore if it's not enabled
- Update the Assert.Ignore to refer to using both the category and the option in the test config

Assets/Tests/PlayMode/Services/ClientSessionTests.cs
```diff
@@ -46,6 +46,10 @@ public partial class ClientSessionTests : EOSTestBase
         [SetUp]
         public void SessionStartup()
         {
+            Assert.Ignore(
+                "This test requires a second machine running as a server. " +
+                "Run only the 'Client' category when a server player is available.");
```
matt-clarke, line 51: If you run the 'Client' category this test will still be ignored though right?

I'm not sure of an ideal resolution right now, but this seems passable for now:
- Add a new bool to the test config ("RunLobbyTestsThatRequireHost")
- Update this code to check that and use Assert.Ignore if it's not enabled
- Update the Assert.Ignore to refer to using both the category and the option in the test config

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -35,12 +35,14 @@
 #define EOS_CAN_SHUTDOWN
 #endif
 
-// This define controls if the EOS SDK should be unloaded in the editor at shutdown to work around DLL unload errors.
-//#define EOS_DO_NOT_UNLOAD_SDK_ON_SHUTDOWN
```
matt-clarke, line 38: I've copied the pending questions from [the original PR](https://github.com/EOS-Contrib/eos_plugin_for_unity/pull/1276) here. @OswaldoApexSystems can you respond to these?

Assets/Tests/Editor/TestPlayerModeSetup.cs
```diff
@@ -66,6 +72,12 @@ public BuildPlayerOptions ModifyOptions(BuildPlayerOptions playerOptions)
                 playerOptions.locationPathName = testBuildLocation;
             }
 
+            // Resolve the output directory (locationPathName may be a .exe path or a folder).
+            string resolvedPath = Path.GetFullPath(playerOptions.locationPathName);
+            s_TestBuildDirectory = Path.HasExtension(resolvedPath)
+                ? Path.GetDirectoryName(resolvedPath)
+                : resolvedPath;
```
matt-clarke, line 79: `FileSystemUtility` seems designed to wrap `Path` and I assume is recommended for consistency.
There's quite a bit of this in this file

Assets/Tests/PlayMode/ConnectTests.cs
```diff
@@ -23,82 +23,50 @@
...
-        /// This test creates a new EOSManager object, and uses it to auth-login.
-        /// Then it does a Connect login.
-        /// After a successful Connect login, it tries to perform another Connect login.
-        /// The result should be as expected.
+        /// Performs auth + connect login while already logged in to verify that
+        /// the result is successful (idempotent login behavior).
         /// </summary>
-        /// <returns></returns>
         [UnityTest]
+        [Category(TestCategories.SoloCategory)]
```
matt-clarke, line 41: Not suggesting we change this here, but I'm thinking we should get rid of this unless you know why it exists. It isn't going to be attached to every single test that isn't a "Client" one, so it'll result in people not running all the tests they expect. Perhaps there are also some tests that are neither "Solo" or "Client" though, but it would likely make sense to just tag those separately

Assets/Tests/PlayMode/ConnectTests.cs
```diff
@@ -23,82 +23,50 @@
...
         public IEnumerator ConnectLogin_WhileAlreadyLoggedIn_ReturnsExpectedResult()
```
matt-clarke, line 42: Can you update the function name to include "DevAuthTool"?

Assets/Tests/PlayMode/AuthenticationTests.cs
```diff
@@ -23,63 +23,35 @@
...
-        /// This test creates a new EOSManager object, and uses it to auth-login.
-        /// Then after a successful login, it tries to auth login again.
-        /// The result should be as expected.
+        /// Attempts a second auth login while already logged in and verifies
+        /// that the result is successful (idempotent login behavior).
         /// </summary>
-        /// <returns></returns>
         [UnityTest]
+        [Category(TestCategories.SoloCategory)]
         public IEnumerator AuthLogin_WhileAlreadyLoggedIn_ReturnsExpectedResult()
```
matt-clarke, line 43: Can you update the function name to include "DevAuthTool"?

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -35,12 +35,14 @@
...
-#if (UNITY_EDITOR_OSX || UNITY_EDITOR_LINUX)
-#define EOS_DO_NOT_UNLOAD_SDK_ON_SHUTDOWN
+// As per Epic's documentation they recommend not calling PlatformInterface.Release / PlatformInterface.Shutdown in editor
+// because Unity doesn't unload managed libraries until the Editor exits. Doing so will cause unintended side effects:
+// https://dev.epicgames.com/docs/epic-online-services/eos-get-started/get-started-guide/next-steps#shut-down-the-eos-sdk
+// This is also required on macOS/Linux always, because there isn't a known reliable way to unload shared libraries
+#if UNITY_EDITOR && (!UNITY_EDITOR_WIN || !EOS_UNLOAD_SDK_ON_SHUTDOWN_IN_WIN_EDITOR)
+// This define controls if the EOS SDK should be unloaded in the editor at shutdown to work around DLL unload errors.
+#define EOS_DO_NOT_UNLOAD_SDK_ON_SHUTDOWN_IN_EDITOR
 #endif
```
matt-clarke, line 46: This [docs file](https://github.com/EOS-Contrib/eos_plugin_for_unity/blob/stable/com.playeveryware.eos/Documentation~/c_sharp_defines.md) covers a couple of points which are a bit of a concern:
- `"It is not the default because if it were, then any edits to the EOSConfig file would require a reboot of the Unity Editor"`
  - Can you check if this is accurate?
  - If it is then we would need to think of a solution to this, perhaps when you save the EOSConfig it prompts you to restart the Editor (after checking the new define mentioned in the comment below)
  - I suspect we also need a Unity hook to check if the file was modified outside of Unity (user could've edited it by hand, or pulled latest changes from their repo)
- `"In the past, not unloading the EOS SDK could also cause issues with state from the previous Unity run not being properly cleaned up."`
  - This is the opposite of what we're seeing right? I'm wary that there's some edge case issues that we haven't seen with this initial testing. It could definitely be a misunderstanding though
  - The [commit message](https://github.com/EOS-Contrib/eos_plugin_for_unity/commit/12cf37bf7a99473f15281369b2b70d93e4896537) where it was added also says the same, so it doesn't appear to just be a mistake in the docs (`It also
carries with it the potential for "leaked" EOS State to carry over
between play in editor runs.`)

com.playeveryware.eos/Runtime/Core/EOSManager.cs
```diff
@@ -35,12 +35,14 @@
...
```
matt-clarke, line 47: With this change to always define `EOS_DO_NOT_UNLOAD_SDK_ON_SHUTDOWN` in Editor this means that this   [docs file](https://github.com/EOS-Contrib/eos_plugin_for_unity/blob/stable/com.playeveryware.eos/Documentation~/c_sharp_defines.md) is out of date.

The change above also removes the ability for developers to control whether it's enabled in Editor or not for Windows.

One option would be to change the above code to this, and update the docs to refer to this define instead:
```
#if UNITY_EDITOR && (!UNITY_EDITOR_WIN || !EOS_UNLOAD_SDK_ON_SHUTDOWN)
#define EOS_DO_NOT_UNLOAD_SDK_ON_SHUTDOWN
#endif
```

matt-clarke (changes requested):

Assets/Tests/Common/EOSTestBase.cs
```diff
@@ -74,54 +74,53 @@ public WaitUntilDone(float timeout, Func<bool> predicate)
...
-            var eosManager = eosObject.AddComponent<EOSManager>();
-            EOSManager.Instance.Init(eosManager);
+            // Reinitialize if this is a fresh session (first run or after play mode was stopped in the editor).
+            if (!s_EOSSessionActive || s_EOSObject == null)
+            {
+                s_EOSObject = new GameObject("EOSManager");
+                UnityEngine.Object.DontDestroyOnLoad(s_EOSObject);
+                var eosManager = s_EOSObject.AddComponent<EOSManager>();
+                EOSManager.Instance.Init(eosManager);
+                s_EOSSessionActive = true;
```
ricardoamores-epic, line 97: Seem like `s_EOSSessionActive` is never set to false during the test suite and also `s_EOSObject` is never null after being initialised. They are also private, so no class can change them.
What do we need the `s_EOSSessionActive` variable for?

Assets/Tests/Common/EOSTestBase.cs
```diff
@@ -144,26 +143,25 @@ public IEnumerator SetupDevAuthLogin()
...
+        /// classes. Resources are released when the test process exits or when the
+        /// next play mode session detects a stale s_EOSObject.
         /// </summary>
         [OneTimeTearDown]
-        public void ShutdownEOS()
-        {
-            EOSManager.Instance?.OnShutdown();
-            UnityEngine.Object.Destroy(eosObject);
-        }
+        public void ShutdownEOS() { }
```
matt-clarke, line 165: Should we still be calling `EOSManager.Instance?.OnShutdown()` once all tests finish?

Assets/Tests/Common/EOSTestBase.cs
```diff
@@ -74,54 +74,53 @@ public WaitUntilDone(float timeout, Func<bool> predicate)
...
-            eosObject = new GameObject();
-            var eosManager = eosObject.AddComponent<EOSManager>();
-            EOSManager.Instance.Init(eosManager);
+            // Reinitialize if this is a fresh session (first run or after play mode was stopped in the editor).
+            if (!s_EOSSessionActive || s_EOSObject == null)
+            {
+                s_EOSObject = new GameObject("EOSManager");
+                UnityEngine.Object.DontDestroyOnLoad(s_EOSObject);
+                var eosManager = s_EOSObject.AddComponent<EOSManager>();
+                EOSManager.Instance.Init(eosManager);
```
matt-clarke, line 96: This calls `EOSManager.Instance.Init` and a comment in the restricted PR says not to because it's already called automatically. Awake does call Init, so it seems that the restricted version is correct

com.playeveryware.eos/Runtime/Core/EOSManager.cs
matt-clarke: Jess' reply from the original PR ([link](https://github.com/EOS-Contrib/eos_plugin_for_unity/pull/1276#discussion_r3405963043)):
>I agree the current wording in c_sharp_defines.md is too absolute and also a bit outdated relative to the current config model. The real concern here is not simply that the config file changes, but that EOSManager now conditionally skips full SDK shutdown/unload in the Editor, which means some initialization related SDK state may persist across play sessions. Because of that, I don’t think we should claim that any config edit requires an Editor restart, a more accurate statement is that config changes which affect EOS SDK initialization may require restarting the Editor when the SDK is not fully unloaded. I also agree that the doc should stop referring to EOSConfig here and instead reflect the current ProductConfig / PlatformConfig flow. If we want better UX around this later, we can consider a targeted restart prompt or external change detection as a follow up, but I think the immediate fix is to make the documentation more precise and aligned with the current EOSManager behavior.

PR #1308, opened 2026-08-28
Guard EOS callbacks with ResultCode.IsOperationComplete()

Pong-Epic:
Update our Sample app code that will check ResultCode.IsOperationComplete to confirm it is final result before check the ResultCode.

Ref: https://dev.epicgames.com/docs/api-ref/functions/eos-e-result-is-operation-complete?lang=en-US


Assets/Scripts/StandardSamples/UI/Login/UILoginMenu.cs
Pong-Epic, line 1348: I haven't add guard here and here look like AuthMFARequired and AuthPinGrantCode actually is not completed result but nothing happen here

matt-clarke: As per that response from James in the Slack thread we should be able to remove this AuthMFARequired check

matt-clarke: We must still check IsOperationComplete because the request might be retried
```suggestion
            else if (!callbackInfo.ResultCode.IsOperationComplete())
            {
                return;
            }
            else
```

matt-clarke (changes requested):

ricardoamores-epic (approved):

matt-clarke (approved):

